Issue #10, opened 2019-03-28
Build command that generates Docker image


davidthor:
# Overview
It's tricky figuring out how to navigate the Architect codebase to evaluate what dependencies and entrypoint are needed in order to start an architect server for a service. Instead of requiring that users create a custom Dockerfile, we should include a base image and build command that will generate a valid, default Docker image on behalf of users. 

This implementation will be a lean proxy to the docker build command that will aim to include as many flags as relevant to users.

```sh-session
$ architect build [OPTIONS] PATH | URL | -
```

The architect build command builds Docker images from a “context”. A build’s context is the set of files located in the specified `PATH` or `URL`. The build process can refer to any of the files in the context. 

The `URL` parameter can refer to three kinds of resources: Git repositories, pre-packaged tarball contexts and plain text files.

### Options
- `--tag` , `-t` - Name and optionally a tag in the ‘name:tag’ format
- `--recursive`, `-r` - Indicates that the build command should generate images for architect dependencies as well

davidthor:
Added as of v0.1.9. Closing.

PR #12, opened 2019-04-03
Docker builds

davidthor:
This PR includes the code necessary to satisfy #10 and build docker images to be published.


Dockerfile-Base
```diff
@@ -0,0 +1,23 @@
+FROM ubuntu:16.04
```
ryan-cahill: Any reason we're not using 18.04?

Dockerfile-Base
```diff
@@ -0,0 +1,23 @@
...
+RUN apt-get update && \
+  apt-get -y upgrade && \
+  apt-get install -y build-essential autoconf libtool pkg-config curl git sudo unzip python3-pip && \
+  pip3 install --upgrade pip click grpcio && \
+  curl -sL https://deb.nodesource.com/setup_10.x -o nodesource_setup.sh && \
+  bash ./nodesource_setup.sh && \
+  apt-get install -y nodejs
+
+# Map python3 to python
+RUN sudo rm /usr/bin/python && \
```
ryan-cahill: Why does sudo need to be installed and run here? When you're running a container you should already be root by default

src/commands/build.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  async run() {
+    const root_service_path = process.cwd();
+    await this.buildImage(root_service_path);
+    this.exit();
+  }
+
+  async buildImage(service_path: string) {
+    const {flags} = this.parse(Build);
+    const service_config = ServiceConfig.loadFromPath(service_path);
```
ryan-cahill: Should there be a check after this to make sure that a service config was actually found?

davidthor: `loadFromPath` already throws an error and triggers an exit when the path doesn't exist.

davidthor:
Updated based on feedback and added tests for the new build command. @ryan-cahill the overall tests will fail for unrelated, already documented reasons, but can you verify the integrity of the tests?

Dockerfile-Base
```diff
@@ -0,0 +1,23 @@
+FROM ubuntu:18.04
+LABEL maintainer = "david.thor@architect.io"
+
+# Install Node.js & Python
+RUN apt-get update && \
+  apt-get -y upgrade && \
+  apt-get install -y build-essential autoconf libtool pkg-config curl git sudo unzip python3-pip && \
```
ryan-cahill: Shouldn't need to install sudo here

Dockerfile-Base
```diff
@@ -0,0 +1,23 @@
...
+  pip3 install --upgrade pip click grpcio && \
```
ryan-cahill: This pip install and the node install should probably be in different run commands than the apt-get installs so that if one command fails, the others don't need to be re-run for no reason

davidthor: That creates more layers which increases the image size. Recommendations are to perform commands the way shown.

davidthor:
All tests pass locally os of merging. Closing.

PR #14, opened 2019-04-10
Add info/error logging for spawned services

tjhiggins:


src/commands/start.ts
```diff
@@ -134,13 +136,28 @@ export default class Start extends Command {
...
+            } else {
+              this.log(_info(`[${service_config.name}]`), _error(data));
+            }
+          }
         });
 
         cmd.on('close', () => {
+          if (hadError) {
+            this.error(_error(`Error executing architect-${service_config.language}-launcher`));
+            this.error(_error(`Failed on: ${cmd_path} ${cmd_args.join(' ')}`));
```
davidthor: The `this.error()` method in the oclif framework actually includes an exit clause. This prevents the second error from from ever executing and thus makes it so we can't tell which service was the culprit.

PR #15, opened 2019-04-24
[#165547388] Fix issues with architect install and start

tjhiggins:


src/commands/start.ts
```diff
@@ -150,6 +151,12 @@ export default class Start extends Command {
 
           resolve();
         });
+
+        cmd.on('error', (e) => {
+          this.log(_error(`Error: spawning architect-${service_config.language}-launcher`));
+          this.log(_error(e.toString()));
+          this.exit(1);
```
davidthor: We should bubble up this error in the promise rejection otherwise we have hanging promises.

davidthor:
@tjhiggins looks pretty good and tests still pass on mac. Just be sure to file a follow-up ticket for the windows placeholder fix for the user IDs. If it works on windows as-is, then let's just remove that line from the docker run command on windows.

tjhiggins:
https://www.pivotaltracker.com/story/show/165564921

src/commands/start.ts
tjhiggins: So instead of calling exit:
`reject(error)`?

davidthor: Yup. We can catch errors at the top and log them.

PR #16, opened 2019-04-30
Added `architect login` cmd

tjhiggins:
Handles the auth0 password grant
Stores access_token in the native credentials store
Created custom base Command
Added ArchitectClient for communication with hub/registry

### DIFF
--- a/src/commands/login.ts
+++ b/src/commands/login.ts
@@ -1,0 +1,80 @@
+import { flags } from '@oclif/command';
+import { AuthenticationClient } from 'auth0';
+import chalk from 'chalk';
+import { spawnSync } from 'child_process';
+import * as inquirer from 'inquirer';
+import * as keytar from 'keytar';
+
+import Command from '../base';
+
+const _error = chalk.red;
+const _success = chalk.green;
+
+export default class Login extends Command {
+  static description = 'Log in to a Architect registry';
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    username: flags.string({
+      char: 'u',
+      description: 'Username'
+    }),
+    password: flags.string({
+      char: 'p',
+      description: 'Password'
+    })
+  };
+
+  async run() {
+    this.log('Login with your Architect ID to push and pull images from Architect Hub.');
+    const answers: any = await this.promptOptions();
+    try {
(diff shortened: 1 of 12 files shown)


src/commands/login.ts
```diff
@@ -0,0 +1,69 @@
...
+    try {
+      await this.login(answers.username, answers.password);
+    } catch (err) {
+      this.error(_error(err.message));
+    }
+  }
+
+  async promptOptions() {
+    const {flags} = this.parse(Login);
+    return inquirer.prompt([{
```
davidthor: Let's skip the prompt if the flags are passed in during command execution. That would help logins on headless flows like CI pipelines.

davidthor:
Looks very straitforward. We can merge once we store the token as docker creds too

src/common/protoc-executor.ts
```diff
@@ -43,17 +43,26 @@ namespace ProtocExecutor {
...
-      '--user', process.platform === 'win32' ? '1000:1000' : '$(id -u):$(id -g)',  // TODO figure out correct user for windows
-      'architectio/protoc-all',
-      '-f', `${mounted_proto_path}`,
-      '-i', mount_dirname,
-      '-l', target_language,
-      '-o', MANAGED_PATHS.DEPENDENCY_STUBS_DIRECTORY
-    ].join(' '), {stdio: 'ignore'});
+
+    // execSync caused the output logs to hang
+    await new Promise(resolve => {
```
davidthor: Why this change if we’re just going to await the result anyway?

src/commands/services/generate.ts
```diff
@@ -0,0 +1,28 @@
+import { flags } from '@oclif/command';
```
davidthor: Looks like there are two new files here that are unrelated to login. Let’s pull those into a separate PR.

src/common/protoc-executor.ts
tjhiggins: execSync was causing the spinners to hang for whatever reason even if it was in an async func. For whatever reason await exec causes it to behave as expected.

src/commands/services/generate.ts
tjhiggins: More than 2. I can pull them out.

PR #17, opened 2019-05-08
Add architect.json label to docker images [#165855804]

tjhiggins:


src/commands/build.ts
```diff
@@ -73,7 +73,21 @@ export default class Build extends Command {
     const dockerfile_path = path.join(__dirname, '../../Dockerfile');
     await Install.run(['--prefix', service_path]);
     const tag_name = flags.tag || `architect-${service_config.name}`;
-    execSync(`docker build --build-arg SERVICE_LANGUAGE=${service_config.language} -t ${tag_name} -f ${dockerfile_path} ${service_path}`);
```
davidthor: What was the reasoning for changing this from exec to spawn?

tjhiggins: Not a good one besides like the array notation. I can change back and do a join to achieve a similar effect.

PR #18, opened 2019-05-17
Add architect push cmd

tjhiggins:
Added Listr for user feedback on cmds.


src/commands/install.ts
```diff
@@ -26,40 +26,51 @@ export default class Install extends Command {
...
 
-  async run() {
-    try {
-      const { flags } = this.parse(Install);
-      let process_path = process.cwd();
-      if (flags.prefix) {
-        process_path = path.isAbsolute(flags.prefix) ?
-          flags.prefix :
-          path.join(process_path, flags.prefix);
+  static async getTasks(root_service_path: string, recursive?: boolean): Promise<Listr.ListrTask[]> {
```
davidthor: This method should be in a common file instead of an individual command since it's called by multiple commands.

src/commands/push.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  async pushImage(service_config: ServiceConfig, tag?: string) {
+    const tag_name = tag || `architect-${service_config.name}`;
+
+    const user = await this.architect.getUser();
+    const repository_name = url.resolve(`${this.app_config.default_registry_host}/`, `${user.username}/${tag_name}`);
+    execSync(`docker tag ${tag_name} ${repository_name}`);
+    // execSync caused the output logs to hang
+    await new Promise(resolve => {
```
davidthor: Seeing this kind of logic in multiple places makes me think we should create a simple method to wrap this up nicely instead of needing to constantly create promises directly.

davidthor:
Feel free to merge this in, but make a note of the two changes requested for a follow-up

src/commands/install.ts
tjhiggins: Good call. Was going to define on our base Command and model after the Command.run func.

src/commands/push.ts
tjhiggins: Agreed. Finally found a good thread to explain.
https://github.com/sindresorhus/ora/issues/86#issuecomment-412359830

Looks like execa is a pretty common lib that returns a promise. I'll use that instead of spinning my own helper func.

PR #22, opened 2019-06-14
Publish flow

tjhiggins:


src/commands/push.ts
```diff
@@ -58,30 +50,36 @@ export default class Push extends Command {
...
-    const dependencies = await ServiceConfig.getDependencies(root_service_path, flags.recursive);
+    if (flags.recursive) {
+      await Build.run([root_service_path, '-r']);
+    } else {
+      await Build.run([root_service_path]);
+    }
+
+    const root_service = ServiceDependency.create(this.app_config, root_service_path);
+    const dependencies = flags.recursive ? root_service.local_dependencies : [root_service];
+    const user = await this.architect.user;
```
davidthor: I know there are fun ways to make variables behave like methods, but it feels really strange and confusing to have an await in front of a variable like this. If the `user` variable is loaded asynchronously, can we use a standard method naming like `getUser()` or something to help visually imply that a request is possibly going to happen?

tjhiggins: Sure

PR #23, opened 2019-06-17
Autocomplete services/environments for plan/apply cmd

tjhiggins:


src/commands/services/plan.ts
```diff
@@ -8,21 +9,96 @@ export default class Plan extends Command {
...
+      this.warn('Canceled apply');
     }
   }
+
+  async promptOptions() {
+    const { args, flags } = this.parse(Plan);
+
+    inquirer.registerPrompt('autocomplete', require('inquirer-autocomplete-prompt'));
+
+    const answers = await inquirer.prompt([{
```
tjhiggins: These questions should probably belong to a inquirer utils since they will be repeated in a lot fo these cmds.

davidthor:
Code all looks great, but let's chat about usability before we merge in. Going to put together my perspective on optimal user flow for commands and we can chat about it soon.

PR #31, opened 2019-07-18
Added environment envs

tjhiggins:
Allow setting on envs on deploy
Added service config validation


src/common/service-env.ts
```diff
@@ -0,0 +1,11 @@
+export default class ServiceEnv {
+  key?: string;
```
davidthor: Do we need a key vs. just using the name? Since we have the description for human-readable info, why not just consolidate the name/key fields?

tjhiggins: I'll remove

src/commands/deploy.ts
```diff
@@ -13,11 +29,216 @@ export default class Deploy extends Command {
 
   static flags = {
     help: flags.help({ char: 'h' }),
-    environment: flags.string({ char: 'e' }),
-    plan_id: flags.string({ char: 'p' })
+    environment: flags.string({ exclusive: ['local'] }),
+    plan_id: flags.string({ exclusive: ['local'] }),
+    local: flags.boolean({ char: 'l', exclusive: ['environment, plan_id'] }),
```
davidthor: Nice - I like the easy way to enforce exclusivity here

PR #32, opened 2019-07-18
Store username/password so login doesn't expire every 24 hours

tjhiggins:


src/base.ts
```diff
@@ -121,7 +121,14 @@ class ArchitectClient {
       clientId: this.app_config.oauth_client_id
     });
 
-    const access_token = JSON.parse(credentials[0].password).access_token;
+    const {access_token} = await auth0.passwordGrant({
```
tjhiggins: Requesting a token for each cli cmd is prob a little heavy? Should I cache the token in the os cred store too?

PR #40, opened 2019-08-09
Added debug flag _local to push local deps

tjhiggins:


src/commands/deploy.ts
```diff
@@ -36,10 +37,11 @@ export default class Deploy extends Command {
   }
 
   validate_parameters(root_service: ServiceDependency, config_json: any) {
-    const res = { ...config_json };
+    const res = JSON.parse(JSON.stringify(config_json));
```
davidthor: why does this have to be done here?

tjhiggins: Clones the input so it doesn't get modified

PR #46, opened 2019-08-13
Support architect login in headless env

tjhiggins:


src/common/credentials.ts
```diff
@@ -0,0 +1,55 @@
...
+try {
+  keytar = require('keytar');
+} catch {
+  keytar = null;
+}
+const config_path = path.join(os.homedir(), MANAGED_PATHS.HIDDEN, 'config.json');
+
+const getConfig = async () => {
+  await fs.ensureFile(config_path);
+  const config = await fs.readJSON(config_path).catch(() => ({}));
```
davidthor: try/catch block for `await` instead of appending `catch()`?

tjhiggins: meh if it was a more interesting catch I would, but not worth adding the extra lines imo

PR #63, opened 2019-10-07
Add polling for deploy/destroy commands

tjhiggins:


src/commands/deploy.ts
```diff
@@ -319,12 +322,31 @@ export default class Deploy extends Command {
...
+        }
+        if (deployment.status === match_status) {
+          clearInterval(poll);
+          resolve(deployment);
+        }
+        poll_count += 1;
+      }, 3000);
+    });
+  }
+
```
tjhiggins: Currently have most of this deploy logic copied in both cmds. I'll eventually consolidate into a helper class. Mostly just wanted to do the multi-line commend.

PR #69, opened 2019-10-24
WIP: Removed semvar checks on tag names

davidthor:


src/common/service-config.ts
```diff
@@ -121,12 +141,7 @@ export default class ServiceConfig {
   }
 
   setVersion(version: string) {
-    const validator = new SemvarValidator();
-    if (validator.test(version)) {
-      this.version = version;
-    } else {
-      throw new InvalidConfigFileError(`Invalid version "${version}" in architect.json.`);
-    }
+    this.version = version;
```
tjhiggins: I assume we will create another ticket to remove all refs of version after cleaning up the api?

PR #73, opened 2019-11-18
adding install and push commands including protobuf generation for services

ryan-cahill:


package.json
```diff
@@ -43,6 +44,7 @@
     "eslint": "^6.5.1",
     "eslint-plugin-oclif": "^0.1.0",
     "globby": "^10.0.1",
+    "listr": "^0.14.3",
```
davidthor: This refactor aimed to remove use of Listr in favor of verbose logging. Pretty much just pass everything through to stdout instead of forcing things into loading tasks.

src/common/managed-paths.ts
```diff
@@ -0,0 +1,8 @@
+enum MANAGED_PATHS {
+  ARCHITECT_JSON = 'architect.json',
+  DEPENDENCY_STUBS_DIRECTORY = 'architect_services',
+  HIDDEN = '.architect',
+  DEPLOYMENT_CONFIGS = 'deployments',
+}
+
+export default MANAGED_PATHS;
```
davidthor: See `src/paths.ts` for new home for this file.

src/common/service-config.ts
```diff
@@ -0,0 +1,262 @@
+import fs from 'fs';
```
davidthor: See `src/common/service-config/index.ts` for new home for this file.

src/common/service-dependency.ts
```diff
@@ -0,0 +1,212 @@
+import execa = require('execa');
```
davidthor: See `src/commands/deploy.ts` for details on how to use new dependency-manager resource for graphing. If you need details like `api_definitions` or `language`, please add them to the node data structures.

src/common/validation-utils.ts
```diff
@@ -0,0 +1,15 @@
+export interface StringValidator {
```
davidthor: see `src/common/utils/validation.ts`

src/common/protoc-executor.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    if (config && config.api && config.api.definitions) {
+      const service_definitions = config.api.definitions;
+      const current_service_language = config.language;
+
+      if (current_service_language === SUPPORTED_LANGUAGES.NODE) {
+        try {
+          execSync('which grpc_tools_node_protoc');
+        } catch (err) {
+          execSync('npm install -g grpc-tools', { stdio: 'ignore' });
```
davidthor: We're trying to be more verbose with logs, so let's let console logs from commands like this flow through.

src/commands/install.ts
```diff
@@ -16,6 +24,10 @@ export default class Install extends Command {
 
   static flags = {
     ...Command.flags,
+    prefix: flags.string({
+      char: 'p',
+      description: 'Path prefix indicating where the install command should execute from',
+    }),
```
davidthor: We should use the `services` flag like we do in deploys for this iteration. That allows us to specify multiple install paths gracefully if we want to code-gen for multiple folders.

src/commands/install.ts
```diff
@@ -30,43 +42,92 @@ export default class Install extends Command {
...
   };
 
-  private async installSingle(service_id: string) {
-    const service_dir = process.cwd();
-    const parts = service_id.split(':');
-    const service_name = parts[0];
-    let service_tag = parts[1];
-    const config = ServiceConfig.loadFromPath(service_dir);
+  async run() {
+    await this.tasks();
```
davidthor: Seems redundant now that we're not using Listr. Let's just make `tasks()` the new `run()` method

src/commands/install.ts
```diff
@@ -30,43 +42,92 @@ export default class Install extends Command {
...
+  }
 
-    if (config.name === service_name) {
-      throw new Error('Services cannot depend on themselves');
-    }
+  async tasks() {
+    const { args, flags } = this.parse(Install);
+    const root_service_path = flags.prefix ? flags.prefix : process.cwd();
+    const dependency_graph = await generateGraphFromPaths([process.cwd()], new EnvironmentConfigV1(), this.app.api);
```
davidthor: `process.cwd()` should be `root_service_path`, but also see my prior comment about allowing multiple paths to be passed in.

src/commands/install.ts
```diff
@@ -30,43 +42,92 @@ export default class Install extends Command {
...
 
-    const { data: service } = await this.app.api.get(`/services/${service_name}`);
-    if (!service) {
-      throw new Error(`Service not found in the registry`);
-    }
+    if (args.service_name) {
+      // eslint-disable-next-line prefer-const
+      let [service_name, service_version] = args.service_name.split(':');
+      if (!service_version) {
+        throw new Error(`Please specify a version for ${service_name}`);
```
davidthor: Since we're starting to make use of "latest" tags (see build command) we might want to set "latest" as the default here instead of error-ing.

src/commands/install.ts
```diff
@@ -30,43 +42,92 @@ export default class Install extends Command {
...
+      }
+      if (root_service.name === service_name) {
+        throw new Error('Cannot install a service inside its own config');
+      }
+
+      const new_dependencies: { [s: string]: string } = {};
+      if (new_dependencies[service_name]) {
+        throw new Error(`A version of ${service_name} is already installed.`);
+      }
```
davidthor: This conditional will never be hit since you initialize the object in the line prior

src/commands/install.ts
```diff
@@ -30,43 +42,92 @@ export default class Install extends Command {
...
+  }
 
-    if (config.name === service_name) {
-      throw new Error('Services cannot depend on themselves');
-    }
+  async tasks() {
+    const { args, flags } = this.parse(Install);
+    const root_service_path = flags.prefix ? flags.prefix : process.cwd();
+    const dependency_graph = await generateGraphFromPaths([process.cwd()], new EnvironmentConfigV1(), this.app.api);
+    const root_service = Array.from(dependency_graph.nodes.values())[0];
```
davidthor: This won't yield services in the order expected since the graph nodes are a map not an array. This index reference can't be expected to be the "root_service" unless you set `false` as the last param to `generateGraphFromPaths()` which tells it not to recurse through the list.

src/commands/install.ts
```diff
@@ -30,43 +42,92 @@ export default class Install extends Command {
...
+      await ProtocExecutor.execute((root_service as LocalServiceNode), undefined, { api_definitions_contents, service_name });
+      ServiceConfig.saveToPath(root_service_path, config);
+      cli.action.stop(chalk.green(`${args.service_name} installed`));
+    } else {
+      await this.installServices((root_service as LocalServiceNode), dependency_graph, flags.recursive);
     }
+  }
+
+  async get_remote_definitions(remote_service_version: string) {
+    const repository_name = url.resolve(`${this.app.config.registry_host}/`, `${remote_service_version}`);
```
davidthor: Replace with the following which will work better if we ever change repo naming conventions:

```ts
const { data: service } = await this.app.api.get(`/services/${service_name}`);
const repository_url = service.url.replace(/(^\w+:|^)\/\//, ''); // strips the protocol from the URL
const repository_name = `${repository_url}:${tag_name}`;
```

src/commands/install.ts
```diff
@@ -30,43 +42,92 @@ export default class Install extends Command {
...
+      for (const dependency of directDependencies) {
+        if (!(dependency instanceof DatastoreNode)) {
+          cli.action.start(chalk.blue(`Installing ${dependency.name} as dependency of ${target_dependency!.name}`), undefined, { stdout: true });
+          await ProtocExecutor.execute(target_dependency!, (dependency as LocalServiceNode));
+          cli.action.stop(chalk.green(`${dependency.name} installed`));
+          if (recursive && !_seen.includes(dependency.name)) {
+            service_dependencies.push(dependency as LocalServiceNode);
+          }
+        }
+      }
```
davidthor: This would be much cleaner as a fixed loop across all graph nodes:

```js
async installServices(graph: DependencyGraph) {
  graph.nodes.forEach((node, key) => {
    const dependencies = graph.getNodeDependencies(node);
    for (const dependency of dependencies) {
      if (!(dependency instanceof DatastoreNode)) {
        cli.action.start(chalk.blue(`Installing ${dependency.name} as dependency of ${target_dependency!.name}`), undefined, { stdout: true });
        await ProtocExecutor.execute(node!, (dependency as LocalServiceNode));
        cli.action.stop(chalk.green(`${dependency.name} installed`));
      }
    }
  });
}
```

src/commands/push.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    const local_service_nodes = dependencyArray.filter(node => node instanceof LocalServiceNode);
+    if (local_service_nodes.length > 1) { //current service is a LocalServiceDependency
+      throw new Error('Cannot push image with local dependencies');
+    }
+
+    const tasks: Promise<void>[] = [];
+    for (const dependency of dependencies_to_push) {
+      if (!(dependency instanceof DatastoreNode)) {
+        tasks.push(
```
davidthor: We experience a lot of bandwidth issues with this kind of parallelization, so let's do pushes sequentially. There's a lot of open convos on docker issues talking about how/why they don't allow `docker push ...` to accept multiple images, so we should just do one at a time.

src/common/dependency-manager/node/index.ts
```diff
@@ -0,0 +1,59 @@
+import ServiceSubscriptions from '../../service-config/subscriptions';
```
davidthor: This file is not legacy, isn't it?

src/common/service-config/index.ts
```diff
@@ -31,6 +24,9 @@ export default class ServiceConfig {
   language?: string;
   subscriptions?: ServiceSubscriptions;
   debug?: string;
+  host?: string;
+  version?: string;
```
davidthor: Trying to get away from hardcoded versions inside service configs, so let's remove this

src/commands/install.ts
ryan-cahill: Do you still want this changed after our conversation on the way to lunch?

src/commands/install.ts
ryan-cahill: That seems like it has potential to get really messy, plus shouldn't someone be able to use the `-r` flag if they want to do this?

davidthor: That only works for a single service and it's deps. Doesn't work for peer services that don't relate to each other. 

It likely won't be common to use it that way, but it creates a standard convention that matches the other commands.

src/commands/install.ts
davidthor: Yes please

PR #78, opened 2019-12-02
Local datastore fix

ryan-cahill:


src/common/dependency-manager/local-manager.ts
```diff
@@ -75,7 +75,7 @@ export default class LocalDependencyManager extends DependencyManager {
     });
 
     if (config.getDebugOptions()) {
-      node.command = config.getDebugOptions()?.command;
+      node.command = config.getDebugOptions() ?.command;
```
davidthor: There shouldn't be a space here.

src/common/docker-compose/index.ts
```diff
@@ -13,24 +17,27 @@ export const generate = (dependency_manager: DependencyManager): DockerComposeTe
...
-            datastores: {},
-            subscriptions: {},
-          },
-        }),
-        ARCHITECT_CURRENT_SERVICE: node.name,
-        ...node.parameters,
-      },
-    };
+
+    if (!isLocalDatastore(node)) {
```
davidthor: This would make more sense to generalize the node to be able to assert something like `if (!node.customHost()) { ... }`. That way you can make the solution work when regular services have host overrides too.

src/dependency-manager/src/manager.ts
```diff
@@ -130,14 +130,22 @@ export default abstract class DependencyManager {
    */
   protected async loadDatastores(parent_node: DependencyNode) {
     for (const [ds_name, ds_config] of Object.entries(parent_node.service_config.getDatastores())) {
+      const image_or_host: { [key: string]: string } = {};
+      const environment_service_config = this.environment.getServices()[`${parent_node.name}:${parent_node.tag}`];
+      if (environment_service_config ?.datastores[ds_name] ?.host) {
```
davidthor: again, shouldn't be a space where the "?" is.

src/dependency-manager/src/manager.ts
```diff
@@ -130,14 +130,22 @@ export default abstract class DependencyManager {
...
+        ...image_or_host,
         key: ds_name,
         service_config: parent_node.service_config,
-        image: ds_config.docker.image,
         tag: parent_node.tag,
         ports: {
           target: ds_config.docker.target_port,
-          expose: await this.getServicePort(),
+          expose: image_or_host['image'] ? await this.getServicePort() : ds_config.docker.target_port,
         },
```
davidthor: After reading through this, I think the idea of a host override demands a new node type like `ExternalNode` for us to parse and handle as if it is not a deployable unit. Once we know that a dependency already exists and has a host address, we don't really need any other information.

src/common/dependency-manager/local-manager.ts
ryan-cahill: :grimacing: my IDE disagrees probably because it's a very new feature

ryan-cahill: :grimacing: my IDE disagrees probably because it's a very new feature

src/dependency-manager/src/manager.ts
```diff
@@ -130,21 +131,38 @@ export default abstract class DependencyManager {
...
+      };
+      const environment_service_config = this.environment.getServices()[`${parent_node.name}:${parent_node.tag}`];
+      let dep_node;
+
+      if (environment_service_config?.datastores[ds_name]?.host) {
+        dep_node = new ExternalNode({
+          ...dep_node_config,
+          host: environment_service_config.datastores[ds_name].host!,
+          ports: {
+            target: ds_config.docker.target_port,
```
davidthor: Two comments right here:
1. The docker target_port is a perfect "default", but make sure to check the env config to see if the user specified a different port the external service is running on. That can be overridden just like the host can (a host is required in order to override the target_port though.
2. Specify the same port as both the target and expose port. Docker compose uses the target port to connect services together, but most systems will use the expose port to broker interprocess communication (e.g. Kubernetes services connect to peers using the ClusterIP port, not the process' port). While doing this, make sure to set the expose_port to required on the interface again.

src/common/docker-compose/index.ts
```diff
@@ -57,7 +61,9 @@ export const generate = (dependency_manager: DependencyManager): DockerComposeTe
         compose.services[node.normalized_ref].volumes = [`${src_path}:/usr/src/app/src`];
       }
     } else {
-      compose.services[node.normalized_ref].image = node.image;
+      if (!(node instanceof ExternalNode)) {
+        compose.services[node.normalized_ref].image = node.image;
+      }
     }
```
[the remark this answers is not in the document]
ryan-cahill: :man_facepalming:

PR #82, opened 2019-12-03
Env param injection

ryan-cahill:
GitLab ticket: https://gitlab.com/architect-io/hub-api/issues/51

### DIFF
--- a/src/common/docker-compose/index.ts
+++ b/src/common/docker-compose/index.ts
@@ -1,8 +1,27 @@
 import fs from 'fs-extra';
 import path from 'path';
-import DependencyManager, { DatastoreNode, ServiceNode } from '../../dependency-manager/src';
+import DependencyManager, { DatastoreNode, DependencyNode, EnvironmentConfig, ServiceNode } from '../../dependency-manager/src';
 import { LocalServiceNode } from '../dependency-manager/local-service-node';
 import DockerComposeTemplate from './template';
+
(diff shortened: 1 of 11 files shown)


src/common/docker-compose/index.ts
```diff
@@ -1,10 +1,31 @@
...
 import path from 'path';
-import DependencyManager, { DatastoreNode, ServiceNode } from '../../dependency-manager/src';
+import DependencyManager, { DatastoreNode, DependencyNode, ServiceNode } from '../../dependency-manager/src';
 import { ExternalNode } from '../../dependency-manager/src/graph/node/external';
 import { LocalServiceNode } from '../dependency-manager/local-service-node';
 import DockerComposeTemplate from './template';
 
+const inject_params = (environment: { [key: string]: string | number }, data_node: DependencyNode, ) => {
+  const param_prefix = `$ARC_${data_node.name.replace(/[^\w\s]/gi, '_').toUpperCase()}_${data_node.tag.replace(/[^\w\s]/gi, '_').toUpperCase()}`;
+  const host_param_placeholder = `${param_prefix}_HOST`;
```
tjhiggins: Does add a backslash to the $ not work as opposed to adding a module?
`\${param_prefix}_HOST`;

ryan-cahill: That would escape the entire variable that we're trying to insert into the string, wouldn't it?

tjhiggins: What is the output of escapeRegExp?

ryan-cahill: Fixed

PR #86, opened 2019-12-10
Accounts refactor

ryan-cahill:


src/commands/environments/destroy.ts
```diff
@@ -22,47 +23,72 @@ export default class EnvironmentDestroy extends Command {
...
-      {
-        type: 'autocomplete',
-        name: 'environment',
-        message: 'Select environment:',
-        source: async (_: any, input: string) => {
-          const params = { q: input };
-          const { data: environments } = await this.app.api.get('/environments', { params });
-          return environments.map((environment: any) => environment.name);
-        },
-        when: !args.environment,
```
davidthor: Why did we get rid of the autocomplete here?

src/commands/environments/destroy.ts
```diff
@@ -22,47 +23,72 @@ export default class EnvironmentDestroy extends Command {
...
+        try {
+          prompted_account = (await this.app.api.get(`/accounts/${value}`)).data;
+          if (prompted_account) {
             return true;
           }
-          return `Name must match: ${chalk.blue(environment)}`;
-        },
-        when: !flags.auto_approve,
+        } catch (err) {
+          return `You do not have access to the account: ${chalk.blue(value)}`;
```
davidthor: We shouldn't hint or elude to the existence of accounts if a user doesn't have access. If they don't have access, they should see messages indicating that the account doesn't exist.

src/commands/environments/destroy.ts
```diff
@@ -22,47 +23,72 @@ export default class EnvironmentDestroy extends Command {
...
+      },
+    }]);
+
+    let account;
+    if (!prompted_account) {
+      const selected_account_name = answers.account_name || args.account_name;
+      try {
+        account = (await this.app.api.get(`/accounts/${selected_account_name}`)).data;
+      } catch (err) {
+        throw new Error(`You do not have access to the account ${selected_account_name}`);
```
davidthor: Same as prior comment regarding existence of accounts in error messages.

src/commands/environments/update.ts
```diff
@@ -31,35 +31,59 @@ export default class EnvironmentUpdate extends Command {
...
   static args = [{
-    name: 'name',
+    name: 'environment',
     description: 'Name of the environment to update',
+    required: true,
     parse: (value: string) => value.toLowerCase(),
+  }, {
+    name: 'account_name',
+    description: 'Account that the environment belongs to',
+    required: false
```
davidthor: This feels like a strange syntax. Since we already force services to be "namespaced" via their own names, perhaps this should just be a single param in the format `{account_name}/{env_name}` to match how we reference services.

src/commands/environments/destroy.ts
ryan-cahill: That's a required param, so it would never be used anyway

PR #87, opened 2019-12-11
Enable serializing the DependencyGraph using class-transformer

tjhiggins:


src/dependency-manager/src/graph/index.ts
```diff
@@ -1,41 +1,74 @@
...
+      edges.push({
+        ...edge,
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        from: this.nodes_map.get(edge.from)!,
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        to: this.nodes_map.get(edge.to)!,
+      });
+    }
+    return edges;
+  }
```
davidthor: This method seems clunky and appears to only be used in one place. Thoughts on replacing it with a more targeted method like `getNodeByRef()`?

PR #88, opened 2019-12-12
Dependency manager cleanup

tjhiggins:


src/common/dependency-manager/local-manager.ts
```diff
@@ -53,24 +53,23 @@ export default class LocalDependencyManager extends DependencyManager {
...
   }
 
-  async loadLocalService(service_path: string): Promise<DependencyNode> {
+  async loadLocalService(service_path: string): Promise<ServiceNode> {
     const config = ServiceConfigBuilder.buildFromPath(service_path);
 
     if (this.graph.nodes_map.has(`${config.getName()}:latest`)) {
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      return this.graph.nodes_map.get(`${config.getName()}:latest`)!;
+      return this.graph.nodes_map.get(`${config.getName()}:latest`)! as ServiceNode;
```
dan-barrett: How will we know that this is a `ServiceNode`? Are we relying on the caller to know that `service_path` happens to refer to a ServiceNode and not some other node type?

Could we do a check for `__type` and throw if something is weird?

src/common/dependency-manager/local-service-node.ts
```diff
@@ -1,25 +1,16 @@
...
   service_path: string;
   command?: string;
-  api: {
-    type: string;
-    definitions?: string[];
-  };
 }
 
-export class LocalServiceNode extends DependencyNode implements LocalServiceNodeOptions {
+export class LocalServiceNode extends ServiceNode implements LocalServiceNodeOptions {
```
dan-barrett: This is going to leave LocalServiceNode with image and tag... but those won't be used, right?

src/common/dependency-manager/local-manager.ts
tjhiggins: For an environment a service:tag is unique so I'm relying on the caller.

src/common/dependency-manager/local-service-node.ts
tjhiggins: The way its implemented at the moment LocalServiceNode needs a tag (latest) for lookups/ref. Image won't be used, but I made it required on the base ServiceNode and this is localized to only the cli so I'm just going to live with it.

PR #90, opened 2019-12-13
Support `valueFrom` references to load parameters from downstream deps

davidthor:


davidthor:
I see - looks like the parameters attempt to load before the dependency nodes have been added. Tried looking to see if that can be fixed, but it'd be non-trivial. I'll close this in favor of yours

PR #94, opened 2019-12-17
93; move datastore call to parent_node in recursive loop

dan-barrett:
This change will be made in conjunction with removing `graph.addNode(top_level)` from the `server.dependency-manager` on the API. It should not affect any changes on the client dependency manager because there this method is overriden.


src/dependency-manager/src/manager.ts
```diff
@@ -177,11 +177,12 @@ export default abstract class DependencyManager {
    */
   async loadDependencies(parent_node: ServiceNode) {
     const dependency_promises = [];
+    this.graph.addNode(parent_node);
+    await this.loadDatastores(parent_node);
+
     for (const [dep_name, dep_id] of Object.entries(parent_node.service_config.getDependencies())) {
       const dep_node = await this.loadService(dep_name, dep_id);
-      this.graph.addNode(dep_node);
       this.graph.addEdge(parent_node, dep_node);
```
tjhiggins: Pull master into this branch. Made changes last night to edges in the graph.

dan-barrett:
Done. FYI this is branched off of custom-auth-domain (and the merge request
is back to custom-auth-domain) so I could continue to use that branch. I
think that will be fine and we'll get the latest master in
custom-auth-domain once this PR is approved.

dan-barrett:
Deprecating this branch and PR in favor of https://github.com/architect-team/architect-cli/pull/96

PR #102, opened 2019-12-23
Cli error messaging and auth check

ryan-cahill:


src/hooks/prerun.ts
```diff
@@ -0,0 +1,28 @@
+import chalk from 'chalk';
+import fs from 'fs-extra';
+import path from 'path';
+import { CREDENTIAL_PREFIX } from '../app-config/auth';
+import AppConfig from '../app-config/config';
+import CredentialManager from '../app-config/credentials';
+import ARCHITECTPATHS from '../paths';
+
+export default async (options: any) => {
+  if (options.Command.id !== 'login') {
```
tjhiggins: Will this prompt you to login for every cmd except login?

ryan-cahill: Yeah, although I should probably add a local deploy to the list of exemptions. I wanted to add something like this because so many of our commands just do really weird things when you aren't logged in, and there's no indication of what to do or why something gives strange results

tjhiggins: @davidthor thoughts? maybe each command should dictate whether login is required.

davidthor: I had the same comment, but forgot to file my review haha. Why don't we just catch the 401 properly and handle it accordingly? That way we don't need any logic in the CLI telling us what's allowed and what isn't.

tjhiggins: I generally like that, but then we also need it for docker and I'm not sure we can determine that without scraping stderr.

davidthor: Perhaps we do both then. Have the handler trigger by default on a 401, but also skip the command entirely if the command has an `auth_required` flag.

ryan-cahill: My thought here is that it's the most configurable if we just add a `@login_required` decorator on functions inside of commands where we need auth. We can't really add it to commands themselves in the case of the deploy command, for example

ryan-cahill: My thought here is that it's the most configurable if we just add a @login_required decorator on functions inside of commands where we need auth. We can't really add it to commands themselves in the case of the deploy command, for example

tjhiggins: Good point on deploy. I'd still love to keep it as a simple flag. Maybe we could do a flag
`auth_required = true`  for most commands, but then special cases get handled with a getter:
```ts
get auth_required() {
  return !flags.local
}
```

ryan-cahill: I added a function rather than a getter because if I was going to use a getter, a setter would have also been required, and we don't want one

tjhiggins: Why is a setter required? eslint? Plenty of places in the api where we don't have a setter.

ryan-cahill: Seemed that it had to do with something under the hood for the actual `Command` class. There was also the issue though that we need to parse the deploy command flags, which couldn't be done in a setter

src/base-command.ts
```diff
@@ -18,6 +28,24 @@ export default abstract class extends Command {
   async init() {
     if (!this.app) {
       this.app = await AppService.create(this.config.configDir);
+
+      if (this.auth_required() && !process.env.ARC_DISABLE_AUTH_CHECK) {
+        const config_dir = this.config.configDir;
```
tjhiggins: I think this can be simplified to:

```ts
if (this.auth_required && !await this.app.auth.getToken()) {
  this.error(chalk.red(`Please log in using 'architect login'`));
}
```

You'll need to add a `getToken` method to the auth.ts
```ts
async getToken() {
  return await this.credentials.get(`${CREDENTIAL_PREFIX}/token`);
}
```

src/base-command.ts
```diff
@@ -18,6 +28,24 @@ export default abstract class extends Command {
   async init() {
     if (!this.app) {
       this.app = await AppService.create(this.config.configDir);
+
+      if (this.auth_required() && !process.env.ARC_DISABLE_AUTH_CHECK) {
```
tjhiggins: Remove the ARC_DISABLE_AUTH_CHECK env. If there are failing tests we can mock the CredentialManager.
```ts
const credential_spy = sinon.fake.returns('token');
sinon.replace(CredentialManager.prototype, 'get', credential_spy);
```

src/dependency-manager/src/graph/index.ts
```diff
@@ -24,6 +24,13 @@ export default abstract class DependencyGraph {
     return node;
   }
 
+  removeNodeByRef(ref: string) {
```
tjhiggins: May as well add a removeEdgeByRef while we are in here.

src/commands/login.ts
```diff
@@ -4,6 +4,10 @@ import Command from '../base-command';
 import inquirer = require('inquirer');
 
 export default class Login extends Command {
+  auth_required() {
+    return false;
```
tjhiggins: The init cmd should also be exempt. Are we missing any other commands?

ryan-cahill: Should it though? We need to prefix the service name with an account name

src/app-config/credentials.ts
```diff
@@ -1,12 +1,14 @@
...
+import path from 'path';
 import AppConfig from './config';
 
 const CREDENTIALS_FILENAME = 'creds.json';
 
 interface Credential {
   account: string;
   password: string;
+  id: string;
+  type: string;
```
tjhiggins: where are id/type used?

src/app-config/auth.ts
```diff
@@ -4,7 +4,7 @@ import LoginRequiredError from '../common/errors/login-required';
 import AppConfig from './config';
 import CredentialManager from './credentials';
 
-const CREDENTIAL_PREFIX = 'architect.io';
+export const CREDENTIAL_PREFIX = 'architect.io';
```
tjhiggins: remove export

src/commands/login.ts
tjhiggins: I think it handles that gracefully atm. Pretty sure David wanted it for non-auth.

src/dependency-manager/src/graph/index.ts
```diff
@@ -24,6 +24,18 @@ export default abstract class DependencyGraph {
...
+    this.nodes = this.nodes.filter(node => node.ref !== ref);
+    this.__nodes_map = undefined;
+    this.edges = this.edges.filter(edge => edge.from !== ref && edge.to !== ref);
+    this.__edges_map = undefined;
+  }
+
+  removeEdgeByRef(from_ref: string, to_ref: string) {
```
tjhiggins: Only pass in a ref: string and change to check `edge.ref`

test/dependency-manager/manager.test.ts
```diff
@@ -56,4 +56,26 @@ describe('manager', function () {
...
+  });
+
+  it('remove graph edge', async () => {
+    const calculator_env_config_path = path.join(__dirname, '../mocks/calculator-environment.json');
+    const manager = await LocalDependencyManager.createFromPath(axios.create(), calculator_env_config_path);
+    const serialized_graph = serialize(manager.graph);
+    const graph = deserialize(LocalDependencyGraph, serialized_graph);
+
+    expect(graph.edges).lengthOf(3);
+    graph.removeEdgeByRef('architect/division-service-grpc:latest', 'architect/subtraction-service-rest:latest');
```
tjhiggins: From above change: `graph.removeEdgeByRef(graph.edges[0].ref)`

src/app-config/credentials.ts
ryan-cahill: They're not used, but they were required here to avoid a Typescript error, or using `Promise<any>` as a return type instead

tjhiggins: Just needed an export on the Credential interface. 

This was using the node internal Credential class
https://github.com/architect-team/architect-cli/pull/102/commits/70784ed74a59d1bc1d040be1e39291c1024df41f#diff-6ad7c92ac31697eb55fb1be46d6d36c2R54

PR #111, opened 2020-01-10
clean up flags for onboarding environment create

dan-barrett:


src/commands/environments/create.ts
```diff
@@ -51,7 +51,7 @@ export default class EnvironmentCreate extends Command {
   static flags = {
     ...Command.flags,
     namespace: flags.string({ char: 'n' }),
-    type: flags.string({ char: 't', options: ['kubernetes'], default: 'kubernetes' }),
+    type: flags.string({ char: 't', options: ['KUBERNETES'], default: 'KUBERNETES' }),
```
tjhiggins: Can we make this case insensitive?

dan-barrett: Yes

dan-barrett:
Making additional changes after conversation with David about platform naming. Hold off on merging.

PR #113, opened 2020-01-12
streamline environment create for early onboarding

dan-barrett:
Cleans up the environment:create flow per David's requested cases.

I'd like to harden this with more tests next time we're in here because I get the sense we're a bit under-covered here. Maybe we take that opportunity when we add ECS as a platform type.

### DIFF
--- a/src/commands/environments/create.ts
+++ b/src/commands/environments/create.ts
@@ -51,7 +51,7 @@ export default class EnvironmentCreate extends Command {
   static flags = {
     ...Command.flags,
     namespace: flags.string({ char: 'n' }),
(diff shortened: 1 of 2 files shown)


src/commands/environments/create.ts
```diff
@@ -84,12 +85,16 @@ export default class EnvironmentCreate extends Command {
       const { data: environment } = await this.app.api.post(`/accounts/${account_id}/environments`, data);
       return environment;
     } catch (err) {
+      //TODO:89:we shouldn't have to do this on the client side
```
tjhiggins: Agreed these should all be responses from the api and have only one error here or a global handler.

PR #114, opened 2020-01-15
adding liveness probes to api spec

ryan-cahill:


src/dependency-manager/src/service-config/base.ts
```diff
@@ -35,6 +35,14 @@ export interface ServiceEventSubscriptions {
 export interface ServiceApiSpec {
   type: string;
   definitions?: string[];
+  liveness_probe?: ServiceLivenessProbe;
+}
+
+export interface ServiceLivenessProbe {
+  success_threshold: number;
```
tjhiggins: Any reason to not add `interval` as well?

src/dependency-manager/src/service-config/v1.ts
```diff
@@ -45,7 +54,16 @@ export class ServiceConfigV1 extends ServiceConfig {
   debug?: string;
   parameters: { [s: string]: ServiceParameterV1 } = {};
   datastores: { [s: string]: ServiceDatastoreV1 } = {};
-  api: ApiSpecV1 = { type: 'rest' };
+  api: ApiSpecV1 = {
+    type: 'rest',
+    liveness_probe: {
+      path: '/health',
```
tjhiggins: I don't think we want defaults/force them to configure correctly. Make liveness_probe optional.

tjhiggins: Also make every key in the interfaces optional `?` except `path`. Figure we can have the defaults server-side if they don't fill everything out.

PR #118, opened 2020-01-24
removing platforms

ryan-cahill:


src/commands/platforms/destroy.ts
```diff
@@ -0,0 +1,74 @@
...
+      },
+      when: !flags.auto_approve,
+    }]);
+
+    answers = { ...args, ...flags, ...answers };
+    const { data: account_platform } = await this.app.api.get(`/accounts/${account.id}/platforms/${platform_name}`);
+    const { data: account_environments } = await this.app.api.get(`accounts/${account.id}/environments`);
+
+    if (account_environments.rows.some((env: any) => env.platform.id === account_platform.id)) {
+      throw new Error('Cannot delete a platform that still has environments');
```
tjhiggins: Errors related to business logic should be triggered in the api. If you throw an httpexception in the api do we show the error in the cli?

src/commands/platforms/destroy.ts
```diff
@@ -0,0 +1,74 @@
...
+    const [account_name, platform_name] = args.namespaced_platform.split('/');
+    if (!account_name || !platform_name) {
+      throw new Error(`Please specify a namespaced platform in the form <account_name>/<platform_name>`);
+    }
+
+    let account;
+    try {
+      account = (await this.app.api.get(`/accounts/${account_name}`)).data;
+    } catch (err) {
+      throw new Error(`The account ${account_name} does not exist.`);
```
tjhiggins: Don't re-throw api errors. Lets expect that the api will return a human readable error msg. If it doesn't we should add an issue. Main is issue is not every error will be account doesn't exist.

src/commands/platforms/destroy.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    cli.action.start(chalk.blue('Destroying platform'));
+    try {
+      await this.app.api.delete(`/platforms/${account_platform.id}`);
+      cli.action.stop(chalk.green('Platform destroyed'));
+    } catch (err) {
+      if (err.response?.data?.statusCode === 403) {
+        throw new Error(`You do not have permission to delete this platform.`);
+      }
+      throw new Error(err);
```
tjhiggins: `throw err` works

src/commands/platforms/destroy.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    if (account_environments.rows.some((env: any) => env.platform.id === account_platform.id)) {
+      throw new Error('Cannot delete a platform that still has environments');
+    }
+
+    cli.action.start(chalk.blue('Destroying platform'));
+    try {
+      await this.app.api.delete(`/platforms/${account_platform.id}`);
+      cli.action.stop(chalk.green('Platform destroyed'));
+    } catch (err) {
```
tjhiggins: Same problem as the above. We should make our api error msging robust instead of having to transform at each usage.

PR #119, opened 2020-01-24
moving removeservice and getdependents to manager from graph

ryan-cahill:


src/dependency-manager/src/graph/index.ts
```diff
@@ -86,4 +86,38 @@ export default abstract class DependencyGraph {
 
     return Array.from(nodes.values());
   }
+
+  removeService(service_ref: string, cleanup: boolean) {
```
tjhiggins: Lets make this more generic. 
rename `removeService` -> `removeNodeAndDangling` 
rename `service_ref` -> `node_ref`

PR #120, opened 2020-02-06
Added host to service config to support external

tjhiggins:


src/dependency-manager/src/service-config/v1.ts
```diff
@@ -57,6 +57,8 @@ export class ServiceConfigV1 extends ServiceConfig {
   name = '';
   description?: string;
   keywords?: string[];
+  image?: string;
+  ports = [8080];
```
davidthor: Per an offline convo, we should allow ports to have names so they can be referenced by upstream callers directly.

PR #134, opened 2020-03-16
decouple docker from service push

dan-barrett:


src/commands/register.ts
```diff
@@ -0,0 +1,131 @@
...
+      }
+    }
+  }
+
+  private async post_service_to_api(dto: CreateServiceVersionInput, account_id: string): Promise<any> {
+    try {
+      const { data: service_digest } = await this.app.api.post(`/accounts/${account_id}/services`, dto);
+      return service_digest;
+    } catch (err) {
+      //TODO:89:we shouldn't have to do this on the client side
```
tjhiggins: Agreed. Remove it and move to api.

src/common/utils/docker.ts
```diff
@@ -54,6 +54,11 @@ export const pushImage = async (image_ref: string) => {
   await docker(['push', image_ref]);
 };
 
+export const getDigest = async (image_ref: string): Promise<string> => {
+  const digest = await docker([`inspect`, `--format='{{index .RepoDigests 0}}'`, image_ref], { stdout: false });
+  return digest.stdout.split('@')[1].replace('\'', '');
```
tjhiggins: Do you want to replace every instance or only the first one?

src/dependency-manager/src/service-config/v1.ts
```diff
@@ -97,6 +98,10 @@ export class ServiceConfigV1 extends ServiceConfig {
     return this.image || '';
   }
 
+  getArtifact(): string {
+    return this.artifact || '';
```
tjhiggins: shouldn't we return string|undefined?

src/common/utils/docker.ts
dan-barrett: There is only one instance. Since it's `stdout` it's getting `'repository@sha256:abcd'` and I just want to get rid of the trailing apostrophe after the split.

dan-barrett:
Good to merge @tjhiggins

PR #135, opened 2020-03-19
helpful error if someone formats a valuefrom incorrectly

ryan-cahill:


src/dependency-manager/src/manager.ts
```diff
@@ -102,6 +102,10 @@ export default abstract class DependencyManager {
               }
               env_params_to_expand[`${param_target_datastore_name}.${node.normalized_ref}.${param_name}`.toUpperCase().replace(/[.-]/g, '_')] =
                 param_value.default.valueFrom.value.replace(/\$/g, `$${node.normalized_ref}.${param_target_datastore_name}_`.toUpperCase()).replace(/[.-]/g, '_');
+            } else {
+              if (typeof param_value.default === 'object' && 'valueFrom' in param_value.default) {
```
tjhiggins: Do we need this check? It looks redundant given its in an if block with `if (param_value.default instanceof Object && param_value.default?.valueFrom) {`

ryan-cahill: Yeah you're right

PR #139, opened 2020-03-23
add parameters field to environment config

dan-barrett:


src/dependency-manager/src/manager.ts
```diff
@@ -213,15 +220,16 @@ export default abstract class DependencyManager {
...
-        if (service_param.required && !env_params.has(key)) {
-          throw new MissingRequiredParamError(key, service_param.description, service_ref);
-        }
 
         let val = env_params.get(key) || service_param.default || '';
         if (typeof val === 'number') {
           val = val.toString();
         }
 
+        if (service_param.required && !val) {
```
tjhiggins: Curious for your thoughts, but I think we should allow for empty parameters. That's why we had changed to `has(key)`. They would still have to explicitly define the param as the empty string.

dan-barrett: Ah, haha, I see... I thought I had found a bug and was fixing it. 😛 

Don't actually need to change this if this is intentional. I'll move it back.

PR #140, opened 2020-03-24
Switch to github actions

davidthor:


davidthor:
@tjhiggins this will do four things: 
1) swap us from circleci to github actions so we can remain on-platform,
2) fix the issue with running the CLI headless directly on ubuntu with keytar, and
3) allow us to cut releases directly from the github interface which will automatically publish releases and build binaries to be auto-attached to the releases.

README.md
```diff
@@ -136,13 +136,13 @@ OPTIONS
...
-  -o, --compose_file=compose_file  [default: C:\Users\TJ\AppData\Local\Temp\architect-deployment-1585610210066.json]
-                                   Path where the compose file should be written to
+  -o, --compose_file=compose_file  [default:
+                                   /tmp/architect-deployment-1585063705042.json] Path where the compose file should be written to
 
   --auto_approve
 ```
 
-_See code: [src\commands\deploy.ts](https://github.com/architect-team/architect-cli/blob/v0.3.19/src\commands\deploy.ts)_
+_See code: [src/commands/deploy.ts](https://github.com/architect-team/architect-cli/blob/v0.3.15/src/commands/deploy.ts)_
```
tjhiggins, line 145: This looks like you resolved incorrectly. The current version is 0.3.19.

davidthor:
Fixed the version ref - you can also ignore the circleCI error since this would move us entirely to github actions. We'd turn off the CircleCI integration once this is merged.

PR #145, opened 2020-03-31
29 multiple interfaces

ryan-cahill:
Testing branch for a live look: https://github.com/architect-team/react-demo/tree/multiple-interfaces


src/dependency-manager/src/service-config/v1.ts
```diff
@@ -93,6 +99,10 @@ export class ServiceConfigV1 extends ServiceConfig {
     return this.api;
   }
 
+  getInterfaces(): { [name: string]: ServiceInterfaceSpec } {
+    return this.interfaces || {};
```
tjhiggins:
```ts
const default_port = this.port || 8080;
return this.interfaces || { default_port: { port: default_port } };
```
Thoughts on always having an interface? Then we don't need any conditional logic elsewhere.

src/dependency-manager/src/manager.ts
```diff
@@ -98,10 +112,15 @@ export default abstract class DependencyManager {
...
         for (const [param_name, param_value] of Object.entries(node.parameters)) { // load param references
           if (param_value instanceof Object && param_value.valueFrom) {
-            const param_target_service_name = (param_value as ValueFromParameter).valueFrom.dependency;
+            const value_from_param = param_value as ValueFromParameter;
+            const param_target_service_name = value_from_param.valueFrom.dependency;
             const param_target_datastore_name = (param_value as DatastoreValueFromParameter).valueFrom.datastore;
+
             if (param_target_service_name) {
-              const param_target_service = this.graph.getNodeByRef(param_target_service_name);
+              if (value_from_param.valueFrom.interface && !all_interface_names.includes(value_from_param.valueFrom.interface)) {
```
tjhiggins: This should only check for the interfaces on the param_target_service. 

`if (value_from_param.valueFrom.interface && value_from_param.valueFrom.interface in param_target_service.interfaces)`

src/dependency-manager/src/manager.ts
```diff
@@ -74,15 +75,28 @@ export default abstract class DependencyManager {
...
       env_params_to_expand[`${node.normalized_ref.toUpperCase()}_HOST`.replace(/[.-]/g, '_')] = external_host ? external_host : internal_host;
       env_params_to_expand[`${node.normalized_ref.toUpperCase()}_EXTERNAL_PORT`.replace(/[.-]/g, '_')] = '80';
-      env_params_to_expand[`${node.normalized_ref.toUpperCase()}_INTERNAL_PORT`.replace(/[.-]/g, '_')] = node.ports.target.toString();
-      env_params_to_expand[`${node.normalized_ref.toUpperCase()}_PORT`.replace(/[.-]/g, '_')] = external_host ? '80' : node.ports.target.toString();
+      env_params_to_expand[`${node.normalized_ref.toUpperCase()}_INTERNAL_PORT`.replace(/[.-]/g, '_')] = node.ports[0].target.toString();
+      env_params_to_expand[`${node.normalized_ref.toUpperCase()}_PORT`.replace(/[.-]/g, '_')] = external_host ? '80' : node.ports[0].target.toString();
+
+      if (node instanceof ServiceNode && node.interfaces) {
+        for (const [interface_name, interface_details] of Object.entries(node.interfaces)) {
+          env_params_to_expand[`${node.normalized_ref.toUpperCase()}_${interface_name.toUpperCase()}_PORT`.replace(/[.-]/g, '_')] = interface_details.port.toString();
```
tjhiggins: Can you add some tests around this?

src/common/dependency-manager/local-manager.ts
```diff
@@ -90,15 +90,28 @@ export default class LocalDependencyManager extends DependencyManager {
 
     const env_service = this.environment.getServiceDetails(`${config.getName()}:latest`);
     const configPort = config.getPort();
+
+    let ports;
+    if (Object.keys(config.getInterfaces()).length > 0) {
```
tjhiggins: Add support for interface overrides. Has to be all or nothing. 
```
{
	"services": {
		"architect/test-service": {
			"host": "...",
			"port": 1234
		},
			
		"architect/test-service-multi-interface": {
			"interfaces": {
				"main-api": {
					"host": "...",
					"port": 1234
				},
				"admin-api": {
					"host": "...",
					"port": 2345
				}
			}
		}
	}
```

src/dependency-manager/src/manager.ts
```diff
@@ -117,8 +120,8 @@ export default abstract class DependencyManager {
             const param_target_datastore_name = (param_value as DatastoreValueFromParameter).valueFrom.datastore;
 
             if (param_target_service_name) {
-              if (value_from_param.valueFrom.interface && !all_interface_names.includes(value_from_param.valueFrom.interface)) {
-                throw new Error(`Interface ${value_from_param.valueFrom.interface} is not defined on any service.`);
+              if (value_from_param.valueFrom.interface && !interface_map[param_target_service_name].includes(value_from_param.valueFrom.interface)) {
```
tjhiggins, line 123: why do you need the map? you get the node in question the line below.

`if (value_from_param.valueFrom.interface && value_from_param.valueFrom.interface in param_target_service.interfaces)`

src/common/dependency-manager/local-manager.ts
```diff
@@ -90,28 +90,17 @@ export default class LocalDependencyManager extends DependencyManager {
...
-
     const node = new LocalServiceNode({
       service_path: service_path.endsWith('.json') ? path.dirname(service_path) : service_path,
       service_config: config,
       image: config.getImage(),
       tag: 'latest',
-      ports,
+      ports: await Promise.all(Object.values(config.getInterfaces()).map(async value => {
+        return {
+          target: env_service?.port ? env_service.port : (configPort ? configPort : value.port),
```
tjhiggins, line 100: Remove env_service?.port that should only be used for external nodes.

src/common/dependency-manager/local-manager.ts
```diff
@@ -178,4 +178,39 @@ export default class LocalDependencyManager extends DependencyManager {
...
+   */
+  async loadExternalService(env_service_config: EnvironmentService, service_ref: string) {
+    if ('interfaces' in env_service_config) {
+      for (const [name, interface_details] of Object.entries(env_service_config.interfaces || {})) {
+        if (!interface_details.host || !interface_details.port) {
+          throw new Error(`As an interface specified in the environment config, interface ${name} requires that both a host and port be declared.`);
+        }
+      }
+
+      if (env_service_config.debug?.path) {
```
tjhiggins, line 193: Hmm not sure what to do about this - prob not worth overriding to only validate local. Would prefer this method to be the same for local/remote.

I wonder if we need to have a method on manager to load a config (local or remote). Also was hoping to have the graph validate itself so maybe we need to store service_config on ExternalNode. Thoughts?

src/dependency-manager/src/manager.ts
```diff
@@ -76,16 +77,32 @@ export default abstract class DependencyManager {
...
+    const gateway_port = gateway_node ? gateway_node.ports[0].expose.toString() : '80';
     for (const node of this.graph.nodes) {
       const external_host = subdomain_map[node.ref] ? `${subdomain_map[node.ref]}.localhost` : '';
       const internal_host = node instanceof ExternalNode ? node.host : node.normalized_ref;
+
       env_params_to_expand[`${node.normalized_ref.toUpperCase()}_EXTERNAL_HOST`.replace(/[.-]/g, '_')] = external_host;
-      env_params_to_expand[`${node.normalized_ref.toUpperCase()}_INTERNAL_HOST`.replace(/[.-]/g, '_')] = internal_host;
-      env_params_to_expand[`${node.normalized_ref.toUpperCase()}_HOST`.replace(/[.-]/g, '_')] = external_host ? external_host : internal_host;
+      if (internal_host) {
+        env_params_to_expand[`${node.normalized_ref.toUpperCase()}_INTERNAL_HOST`.replace(/[.-]/g, '_')] = internal_host;
```
tjhiggins, line 87: Can you add a helper method that takes a node and a string to transform these values? The # of replaces is giving me hives. 
```
`${node.normalized_ref.toUpperCase()}_EXTERNAL_HOST`.replace(/[.-]/g, '_')
```
something like:
`scopeEnv(node, 'EXTERNAL_HOST')`

src/common/dependency-manager/local-manager.ts
```diff
@@ -40,7 +40,7 @@ export default class LocalDependencyManager extends DependencyManager {
     const dependency_resolvers = [];
     for (const [ref, env_svc_cfg] of Object.entries(dependency_manager.environment.getServices())) {
 
-      if (env_svc_cfg.host && env_svc_cfg.port) {
+      if ((env_svc_cfg.host && env_svc_cfg.port) || env_svc_cfg.interfaces) {
```
tjhiggins, line 43: This is exactly what we talked about, but it does look wrong. 

Maybe all nodes should have interfaces instead of a list of ports? Hides behind a bush.

src/common/dependency-manager/local-manager.ts
ryan-cahill: I went back on this one a few times. Really the only reason that this function is overridden here is because the external node has no idea what it is and needs to find out one way or another. I like the idea of having some kind of whatIsMyServiceConfig function. By doing so, we would also be able to store a service config on an external node like you mentioned. It could also potentially be useful in other places? I'll test it out tomorrow and see what it would look like

src/dependency-manager/src/manager.ts
ryan-cahill: Absolutely

src/common/dependency-manager/local-manager.ts
ryan-cahill: Ports are so widespread throughout our stack currently that it might be a whole new ticket (tickets?) in itself to make that conversion. I really like the idea though, as maybe ports were too simple and inflexible. It's also pretty weird that for this ticket I (read: David) ran into a bug where the port of the service config was conflicting with the interfaces. I would love for that never to be able to happen again

tjhiggins: Have interfaces be a getter that returns a default interface if none are defined - mirror service node.

tjhiggins: Ignore the other comment about making ports interfaces

src/common/dependency-manager/local-manager.ts
tjhiggins: lets add a method to the dependency manager
`loadServiceConfig(node_ref:string)` (loads from api by default)

then override in local dependency manager
```
loadServiceConfig(node_ref:string) {
  if (this.environment.services[node_ref]?.debug?.path) {
    // load from local path
  } else {
    return super.loadServiceConfig(node_ref);
  }
}
```

tjhiggins: then we can use this method in the base dependency_manager loadExternalService

tjhiggins: use this method in the loadService and loadLocalService methods too

ryan-cahill: One thing about the base dependency manager is that it doesn't have an instance of an api. Should the method be abstract in the base manager? Then it will either load from the path or the api in the local manager, or the database in the server manager?

tjhiggins: Okay keep it as an abstract method and define on the local

src/dependency-manager/src/environment-config/base.ts
```diff
@@ -48,6 +26,9 @@ export abstract class EnvironmentConfig {
   getServiceDetails(key: string): EnvironmentService | undefined {
     const services = this.getServices();
     const ref = Object.keys(services).find(svc_key => key.startsWith(svc_key));
-    return ref ? services[ref] : undefined;
+    if (!ref) {
+      throw new Error(`Service ${key} not found in environment config`);
```
tjhiggins, line 30: This method can return undefined so we shouldn't throw an error. This will def break loading of deps that aren't defined in the env json.

src/dependency-manager/src/service-config/v1.ts
```diff
@@ -69,6 +74,7 @@ export class ServiceConfigV1 extends ServiceConfig {
   api: ApiSpecV1 = {
     type: 'rest',
   };
+  interfaces?: { [s: string]: InterfaceSpecV1 } = {};
```
tjhiggins, line 77: Is this optional?

src/dependency-manager/src/manager.ts
```diff
@@ -294,7 +325,7 @@ export default abstract class DependencyManager {
 
     for (const [dep_name, dep_id] of Object.entries(parent_node.service_config.getDependencies())) {
       const env_service = this.environment.getServiceDetails(`${dep_name}:${dep_id}`);
-      if (env_service?.host && env_service.port) {
+      if (env_service?.getHost() && env_service.getPort()) {
```
tjhiggins, line 328: This should probably be the same check for interfaces?

src/dependency-manager/src/graph/node/service.ts
```diff
@@ -48,6 +48,10 @@ export class ServiceNode extends DependencyNode implements ServiceNodeOptions {
     return this.service_config.getApiSpec();
   }
 
+  get interfaces() {
+    return this.service_config.getInterfaces() || {};
```
tjhiggins, line 52: getInterfaces shouldn't be optional for a service so no need to safeguard.

src/common/dependency-manager/local-manager.ts
```diff
@@ -118,13 +120,15 @@ export default class LocalDependencyManager extends DependencyManager {
     const dependency_resolvers = [];
     for (const [dep_name, dep_id] of Object.entries(parent_node.service_config.getDependencies())) {
 
-      const env_services = this.environment.getServiceDetails(`${dep_name}:${dep_id}`);
-      if (env_services?.host && env_services?.port) {
-        await this.loadExternalService(env_services, `${dep_name}:${dep_id}`);
+      const env_service = this.environment.getServiceDetails(`${dep_name}:${dep_id}`);
+      if (env_service!.getInterfaces()) {
```
tjhiggins, line 124: this looks like a real lint issue. should be `if (env_service?.getInterfaces()) {`

take a look at the github actions lint warnings

src/common/dependency-manager/local-manager.ts
```diff
@@ -119,72 +82,69 @@ export default class LocalDependencyManager extends DependencyManager {
...
-          dep_node = await this.loadLocalService(this.linked_services[dep_name]);
-        } else {
-          dep_node = await this.loadService(dep_name, dep_id);
-        }
-
-        this.graph.addNode(dep_node);
-        const edge = new ServiceEdge(parent_node.ref, dep_node.ref);
-        this.graph.addEdge(edge);
+  async loadService(service_name: string, service_tag: string, recursive = true): Promise<ServiceNode | ExternalNode> {
+    const ref = `${service_name}:${service_tag}`;
```
dan-barrett, line 86: Is there a reason this method shouldn't just take the `ref` as an argument?

I feel like we do a lot of string manipulation of the ref `${name}:${tag}`, ref.split(':'), etc. Obviously this is unavoidable in some places but ideally we're using the 'ref' as the key in as many places as possible and we only split on the edges of the system where necessary.

This is more of a general comment, maybe it's necessary in this scenario...

src/common/dependency-manager/local-manager.ts
```diff
@@ -119,72 +82,69 @@ export default class LocalDependencyManager extends DependencyManager {
...
+    const existing_node = this.graph.nodes_map.get(ref);
+    if (existing_node) {
+      return existing_node as ServiceNode | ExternalNode;
```
dan-barrett, line 89: Why does this need a cast?

What scenarios might this not be a ServiceNode or an ExternalNode and can we protect against that?

src/common/dependency-manager/local-manager.ts
```diff
@@ -119,72 +82,69 @@ export default class LocalDependencyManager extends DependencyManager {
...
-        expose: await this.getServicePort(),
-      },
-      parameters: await this.getParamValues(
-        `${config.getName()}:${service_digest.tag}`,
-        config.getParameters(),
-      ),
-    });
-    this.graph.addNode(node);
-    return node;
+    this.graph.addNode(service_node);
```
dan-barrett, line 119: I know you didn't change this, but this was the reason I don't like the object-method pattern that we were talking about on Friday. If I call `loadService()` I expect to get a service back. I hardly expect that the method is _also_ going to modify some other state.

I think moving this would require some substantial rework (and probably is made more difficult by the recursive nature of this method), so I think it's okay to leave it... I just want to point out that this is a pattern I generally try to avoid since it leads to behavior that may not be expected by the caller.

src/common/dependency-manager/local-manager.ts
```diff
@@ -119,72 +82,69 @@ export default class LocalDependencyManager extends DependencyManager {
...
+
+  /**
+  * @override
+  */
+  async loadServiceConfig(node_ref: string) {
+    const debug_path = this._environment.getServices()[node_ref]?.getDebugOptions()?.path;
+    if (debug_path) {
+      return ServiceConfigBuilder.buildFromPath(path.join(path.dirname(this.config_path), debug_path));
+    } else {
+      const [account_name, service_name, service_tag] = node_ref.split(/\/|:/);
```
dan-barrett, line 145: Same as above comment. I'd love to centralize this. I know it's not a lot of logic but it's highly specific and a regex that is not easy to read. Same with the `.replace(...)` methods for normalizing the ref, etc. We do string manipulation of the `service`, the `service:tag`, and the `account/service:tag` in many places.

Perhaps a utility class somewhere in the dep-manager with a handful of static methods. We have a lot of these across the codebase (CLI/API/UI). ie `ServiceNames.split(ref: string);`

src/common/docker-compose/index.ts
```diff
@@ -21,110 +21,95 @@ export const generate = (dependency_manager: LocalDependencyManager, build_prod
...
+        ports: [`${await dependency_manager.gateway_port}:${node.ports[0]}`],
         volumes: ['/var/run/docker.sock:/tmp/docker.sock:ro'],
         depends_on: [],
       };
     }
 
     if (node instanceof ServiceNode || node instanceof DatastoreNode) {
+      const ports = [];
+      for (const port of node.ports) {
+        ports.push(`${await dependency_manager.getServicePort()}:${port}`);
```
dan-barrett, line 33: `Promise.all()`?

src/common/docker-compose/index.ts
```diff
@@ -142,7 +127,11 @@ export const generate = (dependency_manager: LocalDependencyManager, build_prod
       service_to.restart = 'always';
       compose.services[node_to.normalized_ref].depends_on.push(node_from.normalized_ref);
     } else if (edge instanceof ServiceEdge) {
-      compose.services[node_from.normalized_ref].depends_on.push(node_to.normalized_ref);
+      if (!seen_edges.has(`${edge.to}__${edge.from}`)) { // Detect circular refs and pick first one
```
dan-barrett, line 130: This would be another decent candidate for the static string manipulation methods mentioned above.

src/common/utils/kubernetes-platform.utils.ts
```diff
@@ -17,7 +17,7 @@ export class KubernetesPlatformUtils {
   ): Promise<CreatePlatformInput> {
 
     let kubeconfig: any;
-    const kubeconfig_path = untildify(flags.kubeconfig!);
+    const kubeconfig_path = untildify(flags.kubeconfig);
```
dan-barrett, line 20: Hurray! 🥳

src/common/dependency-manager/local-manager.ts
tjhiggins: It will never not be a ServiceNode/ExternalNode unless another node has the same ref which would be bad, but nodes_map is a map of DependencyNode so it doesn't know better.

src/common/dependency-manager/local-manager.ts
```diff
@@ -36,80 +35,44 @@ export default class LocalDependencyManager extends DependencyManager {
...
-        if (env_svc_cfg.ingress) {
-          const gateway = new GatewayNode({
-            ports: { target: 80, expose: await dependency_manager.getServicePort(80) },
-            parameters: {},
-          });
+  static async createFromPath(api: AxiosInstance, env_config_path: string, linked_services: LinkedServicesMap = {}, debug = false): Promise<LocalDependencyManager> {
+    const dependency_manager = new LocalDependencyManager(api, env_config_path, linked_services, debug);
+    for (const ref of Object.keys(dependency_manager._environment.getServices())) {
+      const [name, tag] = ref.split(':');
+      const svc_node = await dependency_manager.loadService(name, tag);
```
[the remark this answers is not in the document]
tjhiggins: I'm hesitant to use Promise.all on an unbounded list (ex make 100 api calls at the same time). There is probably another story here to more intelligently load service configs. Either with a batched promise.all or grabbing in bulk.

src/common/dependency-manager/local-manager.ts
tjhiggins: Prob right - I'll look into it

src/common/dependency-manager/local-manager.ts
tjhiggins: Fair enough - I'll see how heavy a lift it would be.

src/common/docker-compose/index.ts
tjhiggins: This one prob makes sense

src/dependency-manager/src/service-config/v1.ts
```diff
@@ -51,38 +74,63 @@ class ApiSpecV1 {
...
-  image?: string;
-  port?: string;
-  command?: string | string[];
-  entrypoint?: string | string[];
-  dependencies: { [s: string]: string } = {};
-  language?: string;
-  debug?: string;
-  parameters: { [s: string]: ServiceParameterV1 } = {};
-  datastores: { [s: string]: ServiceDatastoreV1 } = {};
+  protected name?: string;
```
tjhiggins, line 105: TODO: Remove protected - they won't be expose unless we explicitly type as ServiceConfigV1 and we need access to edit replicas, parameters, etc. in the UI.

src/common/dependency-manager/local-manager.ts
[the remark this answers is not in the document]
tjhiggins: this looks decent: https://www.npmjs.com/package/p-limit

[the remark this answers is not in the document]
tjhiggins: although to make the graph deterministic I'd have to remove the nested addNode calls

PR #146, opened 2020-03-31
adding entrypoint option

ryan-cahill:


src/common/docker-compose/index.ts
```diff
@@ -71,6 +75,10 @@ export const generate = (dependency_manager: DependencyManager, build_prod = fal
           compose.services[node.normalized_ref].command = debug_options.command;
         }
 
+        if (node.service_config.getEntrypoint()) {
+          compose.services[node.normalized_ref].entrypoint = node.service_config.getEntrypoint();
```
tjhiggins: This is redundant you can remove. LocalServiceNode extends ServiceNode.

src/common/docker-compose/index.ts
```diff
@@ -71,6 +75,10 @@ export const generate = (dependency_manager: DependencyManager, build_prod = fal
           compose.services[node.normalized_ref].command = debug_options.command;
```
tjhiggins: Can you add entrypoint to the getDebugOptions. Would be nice to be able to override in the env.json.

src/dependency-manager/src/service-config/v1.ts
```diff
@@ -172,7 +177,13 @@ export class ServiceConfigV1 extends ServiceConfig {
   }
 
   getDebugOptions(): ServiceDebugOptions | undefined {
-    return this.debug ? { command: this.debug } : undefined;
+    if (this.debug || this.entrypoint) {
```
tjhiggins: Oh wait I had this wrong. For a sec I thought debug options was from the env json. You can probably undo this change. Could you add support for overriding the entrypoint in the env json? We currently have a debug block where you can override command.

PR #149, opened 2020-04-07
volumes for service and env configs

ryan-cahill:


src/common/docker-compose/index.ts
```diff
@@ -75,28 +76,51 @@ export const generate = (dependency_manager: DependencyManager, build_prod = fal
...
+        let volumes: string[] = [];
         const src_path = path.join(node.service_path, 'src');
         if (fs.pathExistsSync(src_path)) {
-          compose.services[node.normalized_ref].volumes = [`${src_path}:/usr/src/app/src`];
+          volumes = [`${src_path}:/usr/src/app/src`]; // Mount the src directory
         }
 
         const env_service = dependency_manager.environment.getServices()[node.ref];
+        const service_volumes = node.service_config.getVolumes();
+        const env_volumes = env_service?.debug?.volumes ? env_service.debug.volumes : {};
```
davidthor, line 87: We should always be extracting things from environment and service configs from getters/setters so we can swap out how the login works more easily.

test/dependency-manager/volumes.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      },
+      "volumes": {
+        "service_config_volume": {
+          "mountPath": "/tmp/volume"
+        },
+        "volume_from_param": {
+          "mountPath": "$VOLUME_PATH"
+        },
+        "split_volume": {
+          "mountPath": "/tmp:/vol/service_config"
```
davidthor, line 32: What's the case for putting a split into a mount path? Splits are the joining of a host path to a container path, and the service config won't have any knowledge of the host. Splits should be created as a result of joining a service and environment config.

test/dependency-manager/volumes.test.ts
```diff
@@ -0,0 +1,89 @@
...
+            "volumes": {
+              "service_config_override": "/tmp/service_config_overridden",
+              "env_defined_volume": "/tmp/env_defined_volume"
+            }
+          }
+        }
+      }
+    };
+
+    mock_fs({
```
davidthor, line 54: Didn't know this was a thing! Love it.

test/dependency-manager/volumes.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  afterEach(function () {
+    // Restore stubs
+    sinon.restore();
+    // Restore fs
+    mock_fs.restore();
+  });
+
+  it('service config volume', async () => {
+    expect(compose.services['architect.backend.latest'].volumes).to.include.members(['/tmp/volume']);
```
davidthor, line 71: These tests should be more specific and test that we're successfully joining the host path from the environment config with the container path from the service config.

src/common/docker-compose/index.ts
ryan-cahill: Currently the service config is just represented by a map, so that's why it's done this way here. In the PR for multiple interfaces though, that's updated, so I can wait until that's merged to add what you're talking about here

test/dependency-manager/volumes.test.ts
ryan-cahill: Some of the code that Jon added looked like it expected something like this to be a possibility. I think I'm misunderstanding how you want the env config used overall though and how that will map to compose values, so let's have a short conversation about that tomorrow

test/dependency-manager/volumes.test.ts
ryan-cahill: I think I see what you're saying now actually. I misunderstood how you wanted the env config values to map to service config values

src/common/docker-compose/index.ts
davidthor: All you'd have to do is add a getter to the [base](https://github.com/architect-team/architect-cli/blob/30-volumes/src/dependency-manager/src/environment-config/base.ts#L42) to get volumes. Why not just do that here?

test/dependency-manager/volumes.test.ts
davidthor: The volumes will be composed by combining the host location from the environment config with the container location from the service config. Sounds like you've figured it out though.

src/common/docker-compose/index.ts
ryan-cahill: I didn't want to do that because rather than adding it to the base, we'll soon be able to put that on the `EnvironmentService` where it probably belongs. I've updated the PR though

PR #151, opened 2020-04-08
Docker context

davidthor:
Added support for overriding more docker info both locally while debugging and for production flows:

### Locally

```json
{
  "name": "architect/test-service",
  "debug": {
    "docker": {
      "command": "...",
      "dockerfile": "...",
      "context": "....",
      "entrypoint": "..."
    }
  }
}
```

### Remote builds

_Note: There will have to be a fast-follow to ensure that commands are used for deploys from the platform. Dockerfile and context overrides are build-time features and were worked into the `architect build` command in this PR._

```json
{
  "name": "architect/test-service",
  "docker": {
    "command": "...",
    "dockerfile": "...",
    "context": "....",
    "entrypoint": "..."
  }
}
```

### DIFF
--- a/src/dependency-manager/src/service-config/base.ts
+++ b/src/dependency-manager/src/service-config/base.ts
@@ -51,8 +51,15 @@ export interface ServiceLivenessProbe {
   interval?: string;
 }
 
+export interface ServiceDockerSpec {
+  dockerfile?: string;
+  context?: string;
+  command?: string | string[];
+  entrypoint?: string | string[];
+}
+
 export interface ServiceDebugOptions {
-  command: string | string[];
+  docker?: ServiceDockerSpec;
 }
 
 export abstract class ServiceConfig {
@@ -60,6 +67,7 @@ export abstract class ServiceConfig {
   abstract getName(): string;
   abstract getLanguage(): string;
   abstract getImage(): string;
+  abstract getDockerOptions(): ServiceDockerSpec;
   abstract getCommand(): string | string[];
   abstract getEntrypoint(): string | string[];
   abstract getDependencies(): { [s: string]: string };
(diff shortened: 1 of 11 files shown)


test/mocks/mock-service/src/index.js
```diff
@@ -0,0 +1 @@
+console.log('SUCCESS');
```
tjhiggins, line 1: remove?

src/dependency-manager/src/service-config/base.ts
```diff
@@ -65,6 +72,7 @@ export abstract class ServiceConfig {
   abstract getName(): string;
   abstract getLanguage(): string;
   abstract getImage(): string;
+  abstract getDockerOptions(): ServiceDockerSpec;
```
tjhiggins, line 75: This feels redundant since we already have command/entrypoint at the top level. See two lines below.

Think I'd prefer:
```
  "command": "node index.js"
  "debug": {
    "dockerfile": ".docker/Dockerfile",
    "context": "./src/",
    "command": "echo \"DEBUG\""
    ...
  }
```

Dockerfile/context are weird since they only apply to building the image and won't apply to k8s/ecs. Also weird because we also have the platforms block for defining platform specific config.

davidthor: I thought about doing that to flatten the relationship, but then thought it might be weird with volumes or other non-docker references in the debug block. For example, we might want to define volumes in the debug block like we discussed earlier:

```json
{
  "debug": {
    "docker": {
      "//": "..."
    },
    "volumes": {
      "watch-directory": {
        "description": "Directory containing source code to be mounted locally so files can be watched and updated",
        "mountPath": "./src"
      }
    }
  }
}
```

Do you prefer the above with or without the docker sub-block?

PR #162, opened 2020-04-14
Added support for yaml config files

davidthor:


src/dependency-manager/src/service-config/builder.ts
```diff
@@ -13,31 +15,68 @@ class MissingConfigFileError extends Error {
...
+  static buildFromPath(input: string): ServiceConfig {
+    const try_files = ServiceConfigBuilder.getConfigPaths(input);
 
-  static buildFromPath(service_path: string): ServiceConfig {
-    let config_path = service_path;
-    if (!service_path.endsWith('.json')) {
-      config_path = path.join(service_path, ServiceConfigBuilder.CONFIG_FILENAME);
+    // Make sure the file exists
+    let file_contents;
+    for (const file of try_files) {
```
tjhiggins, line 32: You also need to update the local dependency manager. Supports explicit architect.json and now yaml ref.
```
const node = new LocalServiceNode({
      service_path: service_path.endsWith('.json') ? path.dirname(service_path) : service_path,
      service_config: service_config,
      node_config: this.getNodeConfig(service_config, 'latest'),
      image: service_config.getImage(),
      tag: 'latest',
    });
```

src/dependency-manager/src/service-config/builder.ts
```diff
@@ -13,31 +15,68 @@ class MissingConfigFileError extends Error {
...
+    return [
+      input,
+      path.join(input, 'architect.json'),
+      path.join(input, 'architect.yml'),
+      path.join(input, 'architect.yaml'),
+    ];
+  }
+
+  static buildFromPath(input: string): ServiceConfig {
+    const try_files = ServiceConfigBuilder.getConfigPaths(input);
```
tjhiggins, line 28: Looping over the files seems a little weird. You could check if the input is a file.
```
var fs = require('fs');
var stats = fs.statSync("c:\\dog.jpg");
console.log('is file ? ' + stats.isFile());
```

davidthor: The first item in the returned array is exactly the input. If it's a file that exists, we never reach the second array item.

tjhiggins: Why return a list though? Why not return the path given an input?

davidthor: At some point I need to check if the folder contains an architect.json, architect.yml, or architect.yaml unless I'm given a file path explicitly. The iterator handles all cases and prioritizes them as-is.

PR #164, opened 2020-04-20
requiring interfaces

ryan-cahill:


test/dependency-manager/interfaces.test.ts
```diff
@@ -41,19 +45,29 @@ describe('interfaces', function () {
...
               "dependency": "architect/checkout",
+              "interface": "main",
               "value": "$HOST:$PORT"
             }
           }
+        },
+        "HOST": {
+          "default": {
+            "valueFrom": {
+              "self": true,
```
tjhiggins, line 64: If you don't specify a dependency and you specify an interface - I think it should be inferred that you're self referencing. 
@davidthor thoughts?

davidthor: I concur. Self references are default

PR #167, opened 2020-04-21
Using architect nginx proxy

ryan-cahill:


src/common/docker-compose/index.ts
```diff
@@ -31,7 +31,7 @@ export const generate = async (dependency_manager: LocalDependencyManager): Prom
   for (const node of dependency_manager.graph.nodes) {
     if (node instanceof GatewayNode) {
       compose.services[node.normalized_ref] = {
-        image: 'jwilder/nginx-proxy',
+        image: 'registry.architect.io/architect-nginx/proxy:latest',
```
tjhiggins, line 34: Just thought of this - can I pull this image without being logged in? We should still be able to do a local deploy without having an architect account.

ryan-cahill: Works on my machine

ryan-cahill: Might be nice to make the validation endpoint open though. That currently won't work but I can't think of a reason to protect it

tjhiggins: Yeah I mentioned that to Dan. @dan-barrett can you make public or think of a reason not to?

src/common/docker-compose/index.ts
```diff
@@ -31,7 +31,7 @@ export const generate = async (dependency_manager: LocalDependencyManager): Prom
...
         restart: 'always',
         ports: [`${await dependency_manager.gateway_port}:${node.ports[0]}`],
```
tjhiggins, line 36: Can you add the env var for DISABLE_ACCESS_LOGS?

ryan-cahill: I added it to the Dockerfile when I built the image from source so that we didn't need to keep it here. Did you want it here instead? Or should we keep a copy of the Dockerfile somewhere? it's a one line change

tjhiggins: Yeah I don't want to fork their docker image. Lets build their master's dockerfile and add the env here.

PR #171, opened 2020-04-23
46 social login to cli

dan-barrett:


src/app-config/auth.ts
```diff
@@ -32,6 +37,7 @@ export default class AuthClient {
       this.auth_results = JSON.parse(token.password) as AuthResults;
       const expires_at = this.auth_results.issued_at + this.auth_results.expires_in;
       // Refresh the token if its expired to force a docker login
+      // TODO:46:can we refresh a token without the username/password?  We're going to have to if we want a silent refresh...
```
dan-barrett, line 40: @tjhiggins, today we do a silent token refresh with the username/password. I'm digging into ways this might be possible only using the existing token.

src/app-config/auth0_shim.ts
```diff
@@ -0,0 +1,318 @@
...
+    const params = {
+      ...authorizeOptions,
+      client_id: auth0_client_id,
+      scope: this.getUniqueScopes(
+        'openid profile email',
+        authorizeOptions.scope ? authorizeOptions.scope : ''
+      ),
+      response_type: 'code',
+      response_mode: 'query',
+      audience: 'architect-hub-api', //TODO:46:
```
dan-barrett, line 62: @tjhiggins do you foresee us needing api access from the CLI? If not I think I can remove this....

src/app-config/auth.ts
dan-barrett: I've removed this, ignore

package.json
```diff
@@ -83,6 +94,9 @@
     "plugins": [
       "@oclif/plugin-help"
     ],
+    "hooks": {
+      "init": "./lib/hooks/init/tty"
```
tjhiggins, line 98: 👍

src/app-config/auth.ts
tjhiggins: I assume from our conv on Friday you're investigating using offline-access? To recap I'm okay switching everything to use refresh tokens.

src/app-config/auth0_shim.ts
```diff
@@ -0,0 +1,315 @@
+import atob from "atob";
```
tjhiggins, line 1: I don't hate this, but did you try using the regular auth0 lib as opposed to the auth0 spa?
https://github.com/architect-team/architect-cli/pull/171/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R19

https://auth0.github.io/auth0.js/global.html#buildAuthorizeUrl

tjhiggins: Also worth looking into 
https://auth0.github.io/auth0.js/global.html#renewAuth

dan-barrett: So after some digging both of these share the same problematic property: they're in the `auth0-js` library which is built for use in the browser. Neither this nor `auth0-spa-js` (not totally clear on why they have both) will work in our CLI.

dan-barrett: I did try calling a few different ways on the `auth0` library, the "universal" one that we have in the CLI already and none of these methods exist.

tjhiggins: That's weird - found this issue 
https://github.com/auth0/node-auth0/issues/173#issuecomment-381767911

tjhiggins: Looks like there is another npm lib called auth0-js which is confusing. Can you give that a try - looks like its meant for the browser, but maybe the Authentication class would work?
https://github.com/auth0/auth0.js#auth0authentication

dan-barrett: Ah, didn't see these come in. I gotta configure github notifications better.

Yes, I came across that same issue too. That indicated to me that they didn't support it this, right? The auth0 guy basically just says, "build your own".

And yeah, `auth0-js` was the library I mentioned above. I'm not sure we should be importing a non-universal library in the CLI even if one specific flow works. That opens us up to runtime issues.

dan-barrett:
@tjhiggins 

Added and tested `refresh_token` flow. This is good to merge barring any outstanding issues you have with the auth0 `buildAuthorizeUrl()` flow.

PR #173, opened 2020-04-23
adding dummy user for public images and using architect proxy image

ryan-cahill:


src/app-config/auth.ts
```diff
@@ -35,6 +35,14 @@ export default class AuthClient {
       if (expires_at < (new Date().getTime() / 1000)) {
         await this.refreshToken().catch(() => undefined);
       }
+    } else {
+      await docker([
```
tjhiggins, line 39: Glad this worked. Ideally we wouldn't call this command every init?

ryan-cahill: Where else would we put it to guarantee it was called if it needed to be when the user is logged out? We also potentially refresh the token every init if the user is logged in, so it seemed appropriate that this went here as well

tjhiggins: I think it should go here, but I wonder if we could setToken or check the docker config auths.

PR #177, opened 2020-04-27
Validate on build transform groups

davidthor:


src/common/dependency-manager/local-graph.ts
```diff
@@ -9,32 +10,42 @@ import { ExternalNode } from '../../dependency-manager/src/graph/node/external';
...
-      subTypes: [
-        { value: ServiceEdge, name: 'service' },
-        { value: NotificationEdge, name: 'notification' },
-        { value: IngressEdge, name: 'ingress' },
-      ],
-    },
-    keepDiscriminatorProperty: true,
-  })
+  @TransformEdges('allow-shorthand')
+  @TransformNodes('transform-shorthand')
```
tjhiggins, line 49: edges

package.json
```diff
@@ -14,8 +14,10 @@
     "auth0": "2.20.0",
     "axios": "0.19.0",
     "class-transformer": "0.2.3",
+    "class-validator": "^0.12.2",
     "cli-table3": "0.5.1",
     "cli-ux": "5.3.2",
+    "deepmerge": "^4.2.2",
```
tjhiggins, line 20: remove? I don't see any usage anymore

src/common/dependency-manager/local-graph.ts
```diff
@@ -9,32 +10,42 @@ import { ExternalNode } from '../../dependency-manager/src/graph/node/external';
...
-        { value: DatastoreNode, name: 'datastore' },
-        { value: ExternalNode, name: 'external' },
-        { value: ServiceNode, name: 'service' },
-        { value: GatewayNode, name: 'gateway' },
-      ],
-    },
-    keepDiscriminatorProperty: true,
-  })
+  @TransformNodes('allow-shorthand')
+  @TransformNodes('transform-shorthand')
```
tjhiggins, line 45: It was a good idea, but I don't love this solution for the base case. Would prefer to make granular validation a followup issue, but isn't super pressing imo.

test/dependency-manager/v1-spec/environment.test.ts
```diff
@@ -0,0 +1,113 @@
+import { expect } from 'chai';
```
tjhiggins, line 1: Can we create an explicit validation.test.ts file?
- Test for a variety of expected validation failures
- Test for typos at various places in the env/service config

PR #178, opened 2020-04-27
Nested dependencies

tjhiggins:


src/dependency-manager/src/service-config/v1.ts
```diff
@@ -18,6 +18,24 @@ function transformParameters(input: any) {
   return output;
 }
 
+function transformDependencies(input: any) {
+  const output: any = {};
+  for (const [key, value] of Object.entries(input)) {
+    let config;
+    if (value instanceof Object) {
+      config = { ...value, name: key };
```
tjhiggins, line 26: `config = {private: true, ...value, name: key };`

src/commands/install.ts
```diff
@@ -200,7 +201,9 @@ export default class Install extends Command {
           cli.action.start(chalk.grey(`-- Installing ${service_name} as dependency of ${config.getName()}`), undefined, { stdout: true });
           config.addDependency(service_name, service_tag);
           ServiceConfigBuilder.saveToPath(node.service_path, config);
-          const dep_node = await dependency_manager.loadService(`${service_name}:${service_tag}`);
+
+          // Hack to remove loadService from dependency manager
+          const dep_node = await dependency_manager.loadServiceFromConfig(plainToClass(ServiceConfigV1, { extends: `${service_name}:${service_tag}` }));
```
davidthor, line 206: `plainToClass(...)` should probably be replaced with `ServiceConfigBuilder.buildFromJson(...)` so that we don't hardcode v1 spec builds in several places.

src/dependency-manager/src/service-config/v1.ts
```diff
@@ -177,12 +220,13 @@ export class ServiceConfigV1 extends ServiceConfig {
...
 
-  getDependencies(): { [s: string]: string } {
-    return this.dependencies || {};
+  getDependencies() {
+    return this.dependencies;
   }
 
   addDependency(name: string, tag: string) {
-    this.dependencies[name] = tag;
+    this.dependencies[name] = new ServiceConfigV1();
```
tjhiggins, line 228: This isn't really used, but still should add:

`this.dependencies[name].name = name;`

test/dependency-manager/dependencies.test.ts
```diff
@@ -0,0 +1,504 @@
...
+      expect(graph.edges).length(1);
+
+      expect((graph.nodes[0] as ServiceNode).node_config.getParameters()).keys(['WORKED']);
+      expect((graph.nodes[0] as ServiceNode).node_config.getParameters().WORKED.default).eq(1);
+      expect((graph.nodes[1] as ServiceNode).node_config.getParameters()).keys(['WORKED']);
+      expect((graph.nodes[1] as ServiceNode).node_config.getParameters().WORKED.default).eq(2);
+    });
+
+    /*
+    it('chained refs', async () => {
```
tjhiggins, line 470: @davidthor is thing something that you want to support?

src/dependency-manager/src/environment-config/v1.ts
```diff
@@ -19,10 +19,35 @@ interface DnsConfigSpec {
   searches?: string | string[];
 }
 
+function transformServices(input: any) {
```
tjhiggins, line 22: This method should probably be shared/the same for services/dependencies

test/dependency-manager/external.test.ts
```diff
@@ -0,0 +1,282 @@
...
+      '/stack/src/frontend/architect.json': JSON.stringify(frontend_config),
+      '/stack/arc.env.json': JSON.stringify(env_config),
+    });
+
+    const manager = await LocalDependencyManager.createFromPath(axios.create(), '/stack/arc.env.json', undefined, true);
+    const graph = manager.graph;
+    expect(graph.nodes).length(1);
+    expect(graph.nodes[0]).instanceOf(ExternalNode);
+    expect(graph.nodes[0].interfaces.app.host).eq('app.localhost');
+    expect(graph.nodes[0].interfaces.app.port).eq(80);
```
davidthor, line 105: Just for clarity, what's the final port mapping result of this? I see we "override" the port from the env config, but is the intent to change the runtime port of the process on the container, or to specify the "external" port that we map to? Seems like both ports are relevant in different context's.

test/dependency-manager/external.test.ts
```diff
@@ -0,0 +1,282 @@
...
+      '/stack/arc.env.json': JSON.stringify(env_config),
+    });
+
+    const manager = await LocalDependencyManager.createFromPath(axios.create(), '/stack/arc.env.json', undefined, true);
+    const graph = manager.graph;
+    expect(graph.nodes).length(3);
+    expect(graph.nodes[0]).instanceOf(ServiceNode);
+    expect(graph.nodes[1]).instanceOf(ServiceNode);
+    expect(graph.nodes[2]).instanceOf(ExternalNode);
+    expect(graph.nodes[2].interfaces._default.host).eq('db.localhost');
```
davidthor, line 278: Aren't we in the process of deprecating the "_default" interface? If so, we might want to change the tests.

tjhiggins: This node is an external datastore which to fit the mold has _default as its interface. We should add a ticket to kill datastores if we are happy with this once its merged and integrated into api/app.

src/commands/install.ts
tjhiggins: This code is specifically for build grpc stubs - I'll change to buildFromJson, but should we be removing this cruft entirely?

test/dependency-manager/external.test.ts
tjhiggins: Yeah port func as both atm. If its a ServiceNode then we use it as the runtime port for the container. If its an external node than its the port for the external host. Maybe not ideal since most of the time it'll need to be 443. Works great for managed service overrides because often the port is the same in both cases.

src/dependency-manager/src/manager.ts
```diff
@@ -340,7 +342,16 @@ export default abstract class DependencyManager {
...
   }
 
-  async abstract loadServiceNode(service_ref: string): Promise<ServiceNode>;
+  /**
+   * Queries the API to create a node and config object for a service based on
+   * its name and tag
+   */
+  async loadService(service_ref: string, recursive = true): Promise<ServiceNode | ExternalNode> {
+    // TODO terrible
+    return this.loadServiceFromConfig(plainToClass(ServiceConfigV1, { ref: service_ref }), recursive);
```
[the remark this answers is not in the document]
tjhiggins: I killed this func in my last commit

src/commands/install.ts
davidthor: Yes, we should. No longer used for anything since we deprecated the SDKs.

test/dependency-manager/external.test.ts
davidthor: I believe that means we should do one of two things in a follow up:

1. Ensure that ports are only "overridden" when the host is too via config validation, or
2. Make the exposed port a first-class field that can be overridden separately from the process port

Any thoughts on which you like better?

tjhiggins: Prob #1

test/dependency-manager/dependencies.test.ts
davidthor: I would expect that to work, yes.

tjhiggins: jerk

davidthor: haha, aside from it being more work for you, is there a reason not to? Seems like something I would expect to work if I were a user.

tjhiggins: yeah I'd expect it to work - just requires a while loop of api calls until we stop getting refs

PR #179, opened 2020-04-29
Basic service config validation

davidthor:


src/dependency-manager/src/service-config/v1.ts
```diff
@@ -1,142 +1,366 @@
...
+  }
+
+  const output: Dictionary<ParameterDefinitionSpecV1> = {};
   for (const [key, value] of Object.entries(input)) {
-    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-    // @ts-ignore
-    if (value instanceof Object && !value.valueFrom) {
-      output[key] = value;
+    if (typeof value === 'object') {
+      if (value.value_from || value.valueFrom) {
```
tjhiggins, line 19: Can you update docs?

src/dependency-manager/src/service-config/v1.ts
```diff
@@ -1,142 +1,366 @@
...
+const transformVolumes = (input?: Dictionary<string | Dictionary<any>>): Dictionary<ServiceVolumeV1> | undefined => {
+  if (!input) {
+    return undefined;
+  }
+
+  const output: Dictionary<ServiceVolumeV1> = {};
   for (const [key, value] of Object.entries(input)) {
-    output[key] = value instanceof Object ? value : { host_path: value };
+    output[key] = value instanceof Object
+      ? plainToClass(ServiceVolumeV1, value)
```
tjhiggins, line 44: Can you ignore or resolve the various lint warnings?

src/dependency-manager/src/service-config/v1.ts
```diff
@@ -1,142 +1,366 @@
...
+import { Dictionary } from '../utils/dictionary';
 import { Dict } from '../utils/transform';
+import { validateDictionary, validateNested } from '../utils/validation';
+import { ParameterDefinitionSpecV1 } from '../v1-spec/parameters';
 import { ServiceApiSpec, ServiceConfig, ServiceDatastore, ServiceDebugOptions, ServiceEventNotifications, ServiceEventSubscriptions, ServiceInterfaceSpec, ServiceParameter, VolumeSpec } from './base';
 
-function transformParameters(input: any) {
-  const output: any = {};
+const transformParameters = (input?: Dictionary<any>): Dictionary<ParameterDefinitionSpecV1> | undefined => {
+  if (!input) {
```
tjhiggins, line 12: Maybe change these checks for undefined explictly? Don't want to throw away empty string/0

src/dependency-manager/src/service-config/v1.ts
```diff
@@ -1,142 +1,366 @@
...
+
+  @IsOptional()
+  @IsString()
   image?: string;
+
   @Transform(value => (transformParameters(value)))
-  parameters: { [key: string]: ServiceParameterV1 } = {};
-}
+  @IsOptional()
+  parameters?: Dictionary<ParameterDefinitionSpecV1>;
```
tjhiggins, line 109: I'm a fan of the Dictionary type

test/dependency-manager/validation.test.ts
```diff
@@ -0,0 +1,233 @@
+import { expect } from 'chai';
+import { flattenValidationErrors } from '../../src/common/utils/errors';
+import { ServiceConfigBuilder } from '../../src/dependency-manager/src';
+
+describe('validation (v1 spec)', () => {
```
tjhiggins, line 5: Can we make sure to throw an error on unknown values? `forbidUnknownValues`
https://github.com/typestack/class-validator#passing-options

Something as stupid as `parameter` vs `parameters`

Then add test cases to cover a typo at various nestings?

src/dependency-manager/src/service-config/v1.ts
tjhiggins: Maybe even warrants a test case

davidthor: Do we do something different for `parameters: ""` than for not having a parameters field at all?

src/dependency-manager/src/service-config/v1.ts
davidthor: The other way was just ugly and we use it everywhere. Thought this would encourage us to use a type more often.

src/dependency-manager/src/service-config/v1.ts
tjhiggins: ha realized that it was the dict as input so nvm

test/dependency-manager/validation.test.ts
davidthor: Good idea. I'll make that the default of the `.validate()` method

tjhiggins: Cool - might cause issues with your valueFrom hack, but you could always define both with a comment
value_from
valueFrom

src/dependency-manager/src/utils/dictionary.ts
```diff
@@ -0,0 +1,3 @@
+export interface Dictionary<T> {
+  [key: string]: T;
```
tjhiggins, line 2: Should this be `[key: string]: T | undefined;`
Not sure if this is worth the hassle, but is technically correct for all random key lookups

PR #181, opened 2020-05-04
Avoiding adding extraneous params from datastore mapping

ryan-cahill:


src/dependency-manager/src/manager.ts
```diff
@@ -231,6 +231,14 @@ export default abstract class DependencyManager {
         });
 
         for (const [param_name, param_value] of service_params) {
+          let skip_param = false;
```
tjhiggins, line 234: Can we have `scopeEnv` make prefixes unique? If we added a special char (#) to the end of the prefix like:
`${node.normalized_ref}#${key}`
Then have a method called `unscopeEnv` that strips it off?

Then we wouldn't have to worry about the whether services/datastores have the same startswith.

ryan-cahill: If we were going to do that, we'd have to add this prefixing logic to both sides of the mapping, and add more switching logic in various places depending on if we're looping over a datastore or a service. I like how the logic is at least consolidated here. If we were going to go with a method of prefixing things, it might be nice just to refactor a lot more of the function

tjhiggins: What if I had two services named the following:
architect/cloud
architect/cloud-api

ryan-cahill: Definitely weird. Also maybe weirder: a datastore named primary and an interface named primary :scream: 
That was where my suggestion for refactoring more of the function came from. I think we need to prefix more than just things related to datastores. Thoughts?

tjhiggins:
Closing in favor of https://github.com/architect-team/architect-cli/pull/184

PR #182, opened 2020-05-04
Migrated react-demo to the examples folder of the CLI

davidthor:


examples/react-app/frontend/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "frontend",
+  "name": "@architect-examples/react-app",
```
tjhiggins, line 2: I don't think this is a valid name for a docker image. Remove @.

davidthor: That's not the architect.json, it's the package.json. NPM uses `@` for namespacing (see our CLI package.json)

PR #187, opened 2020-05-06
External multiple interface support

ryan-cahill:
I'll update relevant configs right before we're ready to merge


src/common/dependency-manager/local-manager.ts
```diff
@@ -143,8 +146,8 @@ export default class LocalDependencyManager extends DependencyManager {
 
   toExternalHost(node: DependencyNode) {
```
ryan-cahill, line 147: @tjhiggins would you want this to take an interface key as a param? Or leave as is because until we make decisions about magic params, we really only support one interface per service?

ryan-cahill: Leaning towards option one

src/common/dependency-manager/local-manager.ts
```diff
@@ -36,11 +35,15 @@ export default class LocalDependencyManager extends DependencyManager {
...
-        if (env_ingress) {
+        const interfaces = svc_node.node_config.getInterfaces();
+        const external_interfaces = Object.values(interfaces).filter(i => i.subdomain);
+        const interface_count = Object.keys(interfaces).length;
+        if (interface_count === 1 && external_interfaces.length && external_interfaces[0].subdomain) { // max one interface per container if external exists https://github.com/nginx-proxy/nginx-proxy#multiple-ports
           const gateway = new GatewayNode();
           dependency_manager.graph.addNode(gateway);
-          dependency_manager.graph.addEdge(new IngressEdge(gateway.ref, svc_node.ref, env_ingress.subdomain));
+          dependency_manager.graph.addEdge(new IngressEdge(gateway.ref, svc_node.ref, external_interfaces[0].subdomain));
+        } else if (interface_count > 1 && external_interfaces.length) {
```
tjhiggins, line 45: `interface_count > 1 && external_interfaces.length > 1`

src/dependency-manager/src/manager.ts
```diff
@@ -12,6 +12,8 @@ import GatewayNode from './graph/node/gateway';
 import { DatastoreParameter, DependencyParameter, ServiceConfig, ValueFromParameter, VaultParameter } from './service-config/base';
 import VaultManager from './vault-manager';
 
+export declare type LinkedServicesMap = { [serviceName: string]: string };
```
tjhiggins, line 15: This only is used in the local - move back there

tjhiggins: this is also kind of a silly type...  you could change to Dictionary<string>

src/common/dependency-manager/local-manager.ts
```diff
@@ -5,12 +5,13 @@ import path from 'path';
...
 import IngressEdge from '../../dependency-manager/src/graph/edge/ingress';
 import GatewayNode from '../../dependency-manager/src/graph/node/gateway';
+import { Dictionary } from '../../dependency-manager/src/utils/dictionary';
 import { readIfFile } from '../utils/file';
 import PortUtil from '../utils/port';
 import LocalDependencyGraph from './local-graph';
 import { LocalServiceNode } from './local-service-node';
 
-declare type LinkedServicesMap = { [serviceName: string]: string };
+export declare type LinkedServicesMap = Dictionary<string>;
```
tjhiggins, line 14: ha I meant the type is unnecessary, but its fine.

ryan-cahill: I thought so too but the wording made it sound like you wanted to keep it. Removed in the liveness probe PR

PR #192, opened 2020-05-07
removing kaytar warning, fixing logged out auth error message, fixing crpyto warnings on node <=11

ryan-cahill:


src/base-command.ts
```diff
@@ -22,8 +22,8 @@ export default abstract class extends Command {
   async init() {
     if (!this.app) {
       this.app = await AppService.create(this.config.configDir);
-
-      if (this.auth_required() && !await this.app.auth.getToken()) {
+      const token = await this.app.auth.getToken();
+      if (this.auth_required() && (!token || (token.account === 'unknown' && token.password === 'unknown'))) {
```
tjhiggins, line 26: We probably don't want to show this error in base command since I should be able to deploy locally without logging in.

ryan-cahill: That's handled here: https://github.com/architect-team/architect-cli/blob/master/src/commands/deploy.ts#L30

PR #193, opened 2020-05-07
add protocol and url to loadParameters

dan-barrett:
Closes: https://gitlab.com/architect-io/architect-cli/-/issues/61


src/dependency-manager/src/manager.ts
```diff
@@ -130,13 +130,31 @@ export default abstract class DependencyManager {
           internal_port = this.toInternalPort(node, interface_name);
         }
 
+        const port = external_host ? external_port : internal_port;
+        const host = external_host ? external_host : internal_host;
+
+        const internal_protocol = 'http';
+        const internal_url = internal_protocol + '://' + internal_host + ':' + internal_port;
+
+        const external_protocol = external_host ?
```
tjhiggins, line 139: Can you add an abstract toExternalProtocol method on the manager?
The  host will be localhost when deploying to docker kubernetes and it should be https.

src/dependency-manager/src/manager.ts
```diff
@@ -123,20 +124,39 @@ export default abstract class DependencyManager {
           internal_host = interface_details.host;
           external_port = interface_details.port.toString();
           internal_port = interface_details.port.toString();
+          // external_protocol = interface_details.protocol;
+          // internal_protocol = interface_details.protofol;
+          external_protocol = 'https';
```
dan-barrett, line 129: @tjhiggins , @davidthor - do we want to assume `https` for `EXTERNAL_PROTOCOL` of ExternalNodes?

src/dependency-manager/src/manager.ts
dan-barrett: Done. Your comment about not needing changes on the api threw me off of this pattern yesterday.

tjhiggins: My mistake - just thought of this case

src/dependency-manager/src/manager.ts
tjhiggins: Fine for now

tjhiggins: internal_protocol should also be https. then remove comments above

dan-barrett: Done

PR #194, opened 2020-05-08
liveness probes as parts of interfaces, and test setting/overriding from config

ryan-cahill:
I'll update relevant configs right before we're ready to merge


src/dependency-manager/src/service-config/v1.ts
```diff
@@ -209,8 +199,19 @@ class InterfaceSpecV1 extends BaseSpec {
   @IsString({ always: true })
   host?: string;
 
+  @Type(() => Number)
```
tjhiggins, line 202: I think you can remove. Now that we have validation this shouldn't be an issue except in our own tests.

ryan-cahill: Are you sure? Both the tests and local deployments had problems if I didn't include this

tjhiggins: Yeah that's because the tests themselves defined as strings which would fail validation if we ran it. Try removing now that you updated the tests.

ryan-cahill: Yeah you're right. Updated

PR #200, opened 2020-05-13
WIP: change react-demo to use template syntax

dan-barrett:
@davidthor -

Just want to use this format to talk through open product questions I have regarding the template syntax. ([issue 76](https://gitlab.com/architect-io/architect-cli/-/issues/76))

Here I've changed over the react-demo to the new syntax following your ory example. I want to use this to test my changes since the ory example also contains other not-yet-supported syntax (namely `ref`).

First thing I'd like is for you to just verify that this is what you had in mind and that you'd expect this to be equivalent to the old version of the react-demo.


examples/react-app/backend/architect.yml
```diff
@@ -14,26 +14,11 @@ dependencies:
...
-      value: $POSTGRES_USER
-  POSTGRES_PASSWORD:
-    value_from:
-      dependency: db
-      value: $POSTGRES_PASSWORD
-  POSTGRES_DB:
-    value_from:
-      dependency: db
-      value: $POSTGRES_DB
+  POSTGRES_HOST: ${ dependencies['db'].interfaces.postgres.internal.host }
```
dan-barrett, line 17: My first note is that every "parameter" here that I've changed (in both frontend and backend) is what we would consider to be a "trait" after [issue43](https://gitlab.com/architect-io/product/-/issues/43). Is that correct?

dan-barrett:
@tjhiggins 

I'm continuing to add tests around this, but this is ready for review. I have deploys working.

Note that this does _not_ include the external->internal change we talked about this morning. I did it this AM but decided to roll it back to reduce the number of moving parts after other tests started failing. I will submit that as a second PR after this one... should be small.

src/dependency-manager/src/graph/node/index.ts
```diff
@@ -19,6 +19,10 @@ export abstract class DependencyNode implements DependencyNodeOptions {
       .replace(/\//g, '.');
   }
 
+  get namespace_ref() {
```
tjhiggins, line 22: Once we have everything working lets readdress normalized_ref to work across docker, k8s, ecs, and parameter prefix.

src/dependency-manager/src/utils/interpolation/parameter-interpolator.ts
```diff
@@ -0,0 +1,198 @@
...
+    if (passes >= MAX_DEPTH) {
+      throw new Error('Stack Overflow Error'); //TODO:76: better message
+    }
+
+    return interpolated_parameters;
+  }
+
+  public static build_friendly_name_map(graph: DependencyGraph): { [key: string]: { [key: string]: string } } {
+    const friendly_name_map: { [key: string]: { [key: string]: string } } = {};
+    for (const node of graph.getServiceNodes()) {
```
tjhiggins, line 48: Need to account for ExternalNodes as well

test/dependency-manager/parameter-interpolation.test.ts
```diff
@@ -0,0 +1,126 @@
+import { expect } from '@oclif/test';
+import axios from 'axios';
+import mock_fs from 'mock-fs';
+import moxios from 'moxios';
+import sinon from 'sinon';
+import Build from '../../src/commands/build';
+import LocalDependencyManager from '../../src/common/dependency-manager/local-manager';
+import { ServiceNode } from '../../src/dependency-manager/src';
+
+describe('parameter-interpolation', function () {
```
tjhiggins, line 10: Add test case with recursive loop

dan-barrett:
Closing this PR in favor of one based on child branch: https://github.com/architect-team/architect-cli/pull/215

PR #204, opened 2020-05-13
Update liveness probes

ryan-cahill:


src/dependency-manager/src/service-config/v1.ts
```diff
@@ -356,6 +356,12 @@ export class ServiceConfigV1 extends ServiceConfig {
   @IsOptional({ always: true })
   interfaces?: Dictionary<InterfaceSpecV1>;
 
+  @Type(() => LivenessProbeV1)
+  @IsOptional({ always: true })
+  @IsInstance(LivenessProbeV1, { always: true })
+  @Exclusive(['path', 'command'], { always: true, message: 'Path and command are exclusive.' })
```
tjhiggins, line 362: I like it. Can I suggest we define on each field like oclif?
```ts
{
  @Exclusive(['command'], { always: true })
  path?: string;
  @Exclusive(['command'], { always: true })
  port?: number;
  @Exclusive(['path', 'port'], { always: true })
  command?: string;
}
```

tjhiggins: Probably worth looking at how ValidateIf gets access to the target obj for checking other fields than the one you're currently validating.

src/dependency-manager/src/service-config/v1.ts
```diff
@@ -167,13 +167,23 @@ class LivenessProbeV1 extends BaseSpec {
...
   interval?: string;
+
+  @ValidateIf(obj => !obj.path, { always: true })
+  @IsDefined({ always: true, message: 'Command should be defined if path and port are not defined.' })
+  @IsString({ always: true })
+  command?: string;
+
+  @IsOptional({ always: true })
+  @IsNumber(undefined, { always: true })
+  port?: number;
```
tjhiggins, line 186: Talked to David and we decided to kill the api block in the service config. Mind removing that as part of this ticket?

tjhiggins: If we remove api then we also need to remove a default liveness_probe
1. it doesn't apply if there are multiple interfaces
2. we don't know if an interface is restful

PR #206, opened 2020-05-14
Command arrays

ryan-cahill:


src/dependency-manager/src/service-config/v1.ts
```diff
@@ -176,10 +176,11 @@ class LivenessProbeV1 extends BaseSpec {
   @IsString({ always: true })
   interval?: string;
 
+  @Transform(value => value instanceof Array ? value : [value])
   @ValidateIf(obj => !obj.path || ((obj.path || obj.port) && obj.command), { always: true })
   @Exclusive(['path', 'port'], { always: true, message: 'Command and path with port are exclusive' })
-  @IsString({ always: true })
-  command?: string;
+  @IsArray({ always: true })
```
tjhiggins, line 182: Can you use each:true instead of IsArray?
`@IsString({ always: true, each: true })`

PR #210, opened 2020-05-18
Ingress edge updates for multiple external interfaces

ryan-cahill:


src/dependency-manager/src/graph/edge/ingress.ts
```diff
@@ -1,15 +1,12 @@
-import { Transform } from 'class-transformer';
 import DependencyEdge from '.';
+import { ServiceInterfaceSpec } from '../../service-config/base';
 
 export default class IngressEdge extends DependencyEdge {
   __type = 'ingress';
-  @Transform(value => (value ? value.toLowerCase() : value))
-  subdomain!: string;
+  to_interfaces: { [s: string]: ServiceInterfaceSpec };
```
tjhiggins, line 6: I didn't think we needed to add this since we are using all interfaces on the service.
Was envisioning the following when we need the interfaces for an ingress edge.
```ts
node = graph.getRef(edge.to)
node.interfaces
```

PR #211, opened 2020-05-18
yaml support for remote deployments

ryan-cahill:


src/commands/deploy.ts
```diff
@@ -214,10 +215,16 @@ export default class Deploy extends Command {
     }
 
     const all_answers = { ...args, ...flags, ...answers, ...env_answers };
-    const configPayload = fs.readJSONSync(env_config_path) as object;
```
tjhiggins, line 217: Lets use the dependency managers EnvironmentConfig.loadFromJson method

PR #215, opened 2020-05-22
76 template interpolation on node config

dan-barrett:
@tjhiggins Woof, okay, finally got the global interpolation working and tests passing. 

Will do TODO cleanup the other comments we talked about yesterday in the AM.

Jotted down some of the things we talked about as well as other things I came across that I want to clean up here: https://gitlab.com/architect-io/architect-cli/-/issues/85

(why new PR? Did all the work this AM on a child branch since I was experimenting with two different methods. Branches diverged and the merge is more messy than it's worth so I'm going to close the other PR and we'll go with this one.)


src/dependency-manager/src/manager.ts
```diff
@@ -216,22 +194,217 @@ export default abstract class DependencyManager {
...
+    let passes = 0;
+    const MAX_DEPTH = 100;
+
+    for (const node of this.graph.nodes) {
+      if (node instanceof ServiceNode || (node instanceof ExternalNode && node.node_config instanceof ServiceConfig)) {
+        const serial_config = serialize(node.node_config);
+        const namespaced_serial_config = ParameterInterpolator.namespaceExpressions(node.namespace_ref, serial_config, friendly_name_map[node.ref]);
+        node.node_config = deserialize(ServiceConfigV1, namespaced_serial_config);
+      } else if (node instanceof ExternalNode) {
+        //TODO:76: we can't support interpolation of datastore nodes unless we make ServiceDatastore serializable
```
dan-barrett, line 255: @tjhiggins this is the one open TODO I want to discuss... the `node_config` on the ExternalNode _when it's a ServiceDatastore_ is not serializable. Therefore External Datastores will not support interpolation yet.

src/dependency-manager/src/interpolation/interpolation-context.ts
```diff
@@ -0,0 +1,6 @@
+import { ParameterValueV2 } from '../service-config/base';
+
+export interface InterpolationContext {
```
tjhiggins, line 3: Do we need this anymore? Shouldn't the context be a ServiceConfig?

src/dependency-manager/src/utils/interpolation/interpolation-context.ts
```diff
@@ -0,0 +1,38 @@
+import { ParameterValueV2 } from '../../service-config/base';
+
+export type EnvironmentParameterMap = { [key: string]: ParameterContext };
+
+export interface InterpolationContext {
```
tjhiggins, line 5: Oh here it is again. Same q. Would prefer not to duplicate the service config.

src/dependency-manager/src/utils/interpolation/parameter-interpolator.ts
```diff
@@ -0,0 +1,119 @@
...
+import DependencyGraph from '../../graph';
+import { InterpolationContext } from '../../interpolation/interpolation-context';
+import { ParameterValueV2 } from '../../service-config/base';
+import { EnvironmentInterfaceContext, EnvironmentInterpolationContext, ServiceInterfaceContext } from './interpolation-context';
+
+export class ParameterInterpolator {
+
+  public static interpolateString(param_value: string, environment_context: EnvironmentInterpolationContext): string {
+    Mustache.tags = ['${', '}']; // sets custom delimiters
+    Mustache.escape = function (text) { return text; } // turns off HTML escaping
```
tjhiggins, line 12: fix all lint errors

src/dependency-manager/src/utils/interpolation/parameter-interpolator.ts
```diff
@@ -0,0 +1,119 @@
...
+    //TODO:77: add validation logic to catch expressions that don't refer to an existing path
+    return Mustache.render(param_value, environment_context);
+  }
+
+  public static build_friendly_name_map(graph: DependencyGraph): { [key: string]: { [key: string]: string } } {
+    const friendly_name_map: { [key: string]: { [key: string]: string } } = {};
+    for (const node of graph.nodes) {
+      if (node instanceof ServiceNode) {
```
tjhiggins, line 20: Need to account for ExternalNodes. Add a test case to cover.

src/dependency-manager/src/utils/interpolation/parameter-interpolator.ts
```diff
@@ -0,0 +1,119 @@
...
+      const service_context = ParameterInterpolator.mapNodeToInterpolationContext(node, interface_context[node.ref]);
+      environment_context[node.namespace_ref] = service_context;
+    }
+
+    return environment_context;
+  }
+
+  public static mapNodeToInterpolationContext(node: ServiceNode, interface_context: ServiceInterfaceContext): InterpolationContext {
+    return {
+      parameters: Object.entries(node.node_config.getParameters())
```
tjhiggins, line 50: I think I see why you defined InterpolationContext separately. I assume converting the config to json isn't enough since we transform in the getters?

src/dependency-manager/src/manager.ts
tjhiggins: I see. Semi ignore my other comment. - why is it not serializable?

src/dependency-manager/src/utils/interpolation/parameter-interpolator.ts
```diff
@@ -0,0 +1,119 @@
+import Mustache from 'mustache';
+import { ServiceNode } from '../..';
+import DependencyGraph from '../../graph';
+import { InterpolationContext } from '../../interpolation/interpolation-context';
+import { ParameterValueV2 } from '../../service-config/base';
+import { EnvironmentInterfaceContext, EnvironmentInterpolationContext, ServiceInterfaceContext } from './interpolation-context';
+
+export class ParameterInterpolator {
```
tjhiggins, line 8: Not just parameters anymore

src/dependency-manager/src/utils/interpolation/parameter-interpolator.ts
```diff
@@ -0,0 +1,119 @@
...
+    namespaced_value = namespaced_value.replace(interfaces_search_string, `$\{ ${node_ref}.interfaces`);
+
+    const parameters_search_string = /\$\{\s*parameters/g;
+    namespaced_value = namespaced_value.replace(parameters_search_string, `$\{ ${node_ref}.parameters`);
+
+    const bracket_notation_matcher = /\$\{\s*dependencies\[(.*?)\]\./g;
+    const dot_notation_matcher = /\$\{\s*dependencies\.(.*?)\./g;
+
+    namespaced_value = namespaced_value.replace(bracket_notation_matcher, (m) => {
+      const dep = ParameterInterpolator.extract_friendly_name_from_brackets(m);
```
tjhiggins, line 90: Not loving this nested regex. Can we use a match group?

tjhiggins:
Need to figure this out or at least make it easier to define supported blocks because of future stories like this one:
https://gitlab.com/architect-io/architect-cli/-/issues/83#note_347448629

src/dependency-manager/src/interpolation/interpolation-context.ts
dan-barrett: This was a duplicate file probably from a merge. Removed it. See answer regarding ServiceConfig below.

src/dependency-manager/src/utils/interpolation/interpolation-context.ts
dan-barrett: This object is not the same shape as a ServiceConfig which is why we need it. The interfaces block is enriched for example.

If we made the interpolation context the ServiceConfig we'd either have to:
(1) remove support for the enriched interface expressions `*.interfaces.external.*`
(2) change the ServiceConfig so that we can write those expressions.

The problem with the latter is that the ServiceConfig represents the shape of the file that the user is allowed to give us. That's what we validate and that's what we version. This is different from the shape of the object that we want to enrich.

I'd be open to using the actual `node_config` as the interpolation context, but this follows the same lines of what I've been saying: the `node_config` is not a `ServiceConfig` if we're going to write to it and enrich it. It deserves its own type.

src/dependency-manager/src/manager.ts
dan-barrett: Sorry, meant to say it's not deserializable. It's because ServiceDatastore is an interface.  We could probably change to make it a class unless that has implications that I'm unaware of.

src/dependency-manager/src/utils/interpolation/parameter-interpolator.ts
dan-barrett: I'll try to generalize this along these lines and your point about adding new supported blocks. That's a good idea.

src/dependency-manager/src/utils/interpolation/parameter-interpolator.ts
dan-barrett: This is just one example that underlies the real problem mentioned above: ServiceConfig represents the shape of the architect.json that the user is allowed to give us. I'm not sure we want to overload its use case... that will inevitably lead us to add cruft to the the ServiceConfig which will only be used in one context or another and will inevitably lead to conflicting requirements or a convoluted object that tries to serve two purposes poorly instead of one well.

src/dependency-manager/src/manager.ts
tjhiggins: Yeah we could make it a class. Still no way to distinguish between ServiceConfig/ServiceDatastore when we transform. We really need to kill datastores...

tjhiggins: I think I might be okay not supporting ExternalNodes (seems like there are currently issues). I'm going to add a story to kill since its easy to check node.host.

tjhiggins: https://gitlab.com/architect-io/architect-cli/-/issues/89

src/dependency-manager/src/utils/interpolation/parameter-interpolator.ts
tjhiggins: Like I said above I'm okay not supporting external nodes as a first pass and hopefully they would be supported out of the box by:
https://gitlab.com/architect-io/architect-cli/-/issues/89

src/dependency-manager/src/utils/interpolation/interpolation-context.ts
tjhiggins: I'm leaning towards option 2 since the ServiceConfig is the transformed object not the shape of the file. However, I think that should be another story that can follow this one. I can live with InterpolationContext for some period of time.

dan-barrett:
@tjhiggins , made updates for all of the above comments that we don't have no issues for

dan-barrett:
@tjhiggins 

pushed boolean change. added a test to catch that case in the future.

PR #216, opened 2020-05-22
handling remote validation errors

ryan-cahill:


src/commands/deploy.ts
```diff
@@ -219,7 +220,16 @@ export default class Deploy extends Command {
     const config_payload = classToPlain(await EnvironmentConfigBuilder.buildFromPath(env_config_path));
 
     cli.action.start(chalk.blue('Creating deployment'));
-    const { data: deployment } = await this.app.api.post(`/environments/${all_answers.environment_id}/deploy`, { config: config_payload });
+
+    let deployment;
+    try {
+      deployment = (await this.app.api.post(`/environments/${all_answers.environment_id}/deploy`, { config: config_payload })).data;
```
tjhiggins, line 226: Can we make our api's response flatten the errors so we don't need to have logic in the cli/app?

src/commands/push.ts
```diff
@@ -74,10 +74,14 @@ export default class Push extends Command {
         });
         cli.action.stop(chalk.green(`Image verified`));
 
+        const config = classToPlain(node.service_config); // debug block data for local should not be posted
+        if ((config as any).debug) {
+          (config as any).debug.path = undefined;
```
tjhiggins, line 79: Why do we need to set path to undefined? I thought validation would prevent this.

src/commands/push.ts
```diff
@@ -87,7 +91,11 @@ export default class Push extends Command {
   }
 
   private async post_service_to_api(dto: CreateServiceVersionInput, account_id: string): Promise<any> {
-    const { data: service_digest } = await this.app.api.post(`/accounts/${account_id}/services`, dto);
-    return service_digest;
+    try {
+      const { data: service_digest } = await this.app.api.post(`/accounts/${account_id}/services`, dto);
+      return service_digest;
+    } catch (err) {
+      throw new Error(err.response.data.message);
```
tjhiggins, line 98: I'd prefer to remove all of these try/catch and let the base command catch method handle these types of errors. It might already.

tjhiggins: Lets update it there - then it'll look better for all api failures. Remove the getFormattedJSON method and replace it with something nice like:
```ts
const error_data = err.response.data;
this.error(chalk.red(`status: ${error_data.statusCode}\npath: ${error_data.path}\nmessage: ${error_data.message}`));
```

PR #218, opened 2020-05-28
add dependency-manage npm build to github workflow

dan-barrett:
added `npm run build` of dependency-manager to the `test` github workflow.

### DIFF
--- a/.github/workflows/cli-rc.yml
+++ b/.github/workflows/cli-rc.yml
@@ -60,6 +60,8 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - run: npm install
       - run: npm test
+      - run: npm run build
+        working-directory: ./src/dependency-manager
   cut_release:
     if: github.event_name == 'push' && github.ref == 'refs/heads/master'
     needs: [lint, test]
--- a/.github/workflows/cli-rc.yml
+++ b/.github/workflows/cli-rc.yml
@@ -62,7 +62,7 @@ jobs:
         working-directory: ./src/dependency-manager
       - run: npm run build
         working-directory: ./src/dependency-manager
-      - run: rm -rf node_modules
+      - run: rm -r node_modules
         working-directory: ./src/dependency-manager
       - run: npm install
       - run: npm run build
--- a/.github/workflows/cli-rc.yml
+++ b/.github/workflows/cli-rc.yml
@@ -59,7 +59,10 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm install
+      - run: npm run build
       - run: npm test
+      - run: npm install
+        working-directory: ./src/dependency-manager
       - run: npm run build
         working-directory: ./src/dependency-manager
   cut_release:
--- a/.github/workflows/cli-rc.yml
+++ b/.github/workflows/cli-rc.yml
@@ -62,6 +62,8 @@ jobs:
         working-directory: ./src/dependency-manager
       - run: npm run build
         working-directory: ./src/dependency-manager
+      - run: rm -rf node_modules
+        working-directory: ./src/dependency-manager
       - run: npm install
       - run: npm run build
       - run: npm test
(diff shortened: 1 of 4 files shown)


dan-barrett:
@tjhiggins good to go... as you can see [here](https://github.com/architect-team/architect-cli/runs/717210264#step:7:12), pipeline now fails with missing mustache package in the dependency-manager package.json.

Ready for merge.

Notes: 
* pipeline was still passing if we do `npm run build` in dep-manager _after_ the CLI `npm install` because the dependency-manager finds the `mustache` package from the CLI `node_modules` so we actually have to do the dep-manager first before the CLI
* the CLI tests _fail_ if the `node_modules` directory of the dep-manager contains packages (you can reproduce this locally and has been an issue as long as I can remember) so we have to `rm -r node_modules` from the dep-manager before carrying on.
* realized that these commands are not necessarily run in bash and have to be OS agnostic, thankfully Windows Powershell has aliased `Remove-Item` to `rm` but the `-f` flag doesn't work. Shouldn't be an issue in this case but tripped me up for a while.

.github/workflows/cli-rc.yml
```diff
@@ -59,6 +59,13 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm install
+        working-directory: ./src/dependency-manager
+      - run: npm run build
```
tjhiggins, line 63: why do we build - remove and build again?

src/dependency-manager/package.json
```diff
@@ -17,13 +17,15 @@
   "homepage": "https://github.com/architect-team/architect-cli",
   "license": "GPL-3.0",
   "dependencies": {
+    "@types/mustache": "^4.0.1",
```
tjhiggins, line 20: Can you add @types to the dev dependencies?

.github/workflows/cli-rc.yml
davidthor: If this is some kind of workaround, seems like we'd be better off updating the npm build command to do what we want instead of working around it in the CI job.

src/dependency-manager/package.json
dan-barrett: Done

.github/workflows/cli-rc.yml
dan-barrett: The `npm run build`s are two different builds, one is the dep-manager and one is the cli.

We remove `node_modules` from the dep-manager after we're done with it because the CLI tests do not pass when there are any `node_modules` in the dep-manager. This is an issue that dates back beyond last fall. If memory serves correctly I think @davidthor also pointed it out one of our daily standups semi-recently, maybe a month back.

You can reproduce the issue locally: `cd src/dependency-manager`, `npm install`, `cd ../..`, `npm run test` (should fail). then `rm -rf src/dependency-manager/node_modules`, `npm run test` (should pass).

I'm not sure it's really a workaround that belongs in the `npm run build` script.  I've opened an issue at the very least so we can track it: https://gitlab.com/architect-io/architect-cli/-/issues/93

src/dependency-manager/package.json
```diff
@@ -17,13 +17,13 @@
...
   "license": "GPL-3.0",
   "dependencies": {
-    "@types/shell-quote": "^1.6.2",
     "axios": "^0.19.0",
     "class-transformer": "^0.2.3",
     "class-validator": "^0.12.2",
     "dotenv-expand": "^5.1.0",
     "fs-extra": "^8.1.0",
     "js-yaml": "^3.13.1",
+    "mustache": "^4.0.1",
```
tjhiggins, line 26: Need to update the package-lock.json. Make sure you remove the node modules dir for dependency-manager after installing

.github/workflows/cli-rc.yml
tjhiggins: I get it now - just confusing to look at

src/dependency-manager/package.json
dan-barrett: Done, good call

PR #221, opened 2020-06-02
83 environment block

dan-barrett:
@tjhiggins - here is the environments block


src/common/docker-compose/index.ts
```diff
@@ -60,6 +60,24 @@ export const generate = async (dependency_manager: LocalDependencyManager): Prom
...
+      const ports = [];
+      for (const port of node.ports) {
+        ports.push(`${available_ports.shift()}:${port}`);
+      }
+      compose.services[node.normalized_ref] = {
+        image: node.image ? node.image : undefined,
+        ports,
+        depends_on: [],
+        environment: {
+          ...node.parameters, // TODO:86: we can remove parameters from this block once we remove valueFrom
```
dan-barrett, line 73: Just note that right now I'm setting both the `parameters` and the `getEnvironmentVariables()` to the docker-compose environment variables. This was a workaround to support the old valueFrom (which rewrites parameters).

You should be able to remove this once this is in your branch and valueFrom is not supported... I think we should only be putting our `getEnvironmentVariables()` in the environment block.

PR #229, opened 2020-06-15
Cli version header

ryan-cahill:


src/app-config/service.ts
```diff
@@ -91,6 +93,7 @@ export default class AppService {
       const { token_type, access_token } = this.auth.auth_results;
       this._api.defaults.headers = {
         Authorization: `${token_type} ${access_token}`,
+        'Cli-Version': this.version,
```
dan-barrett, line 96: Should this be `X-Cli-Version` since this is a custom header?

test/commands/build.test.ts
```diff
@@ -64,7 +64,7 @@ describe('build', function () {
     });
     const tmp_config_file = path.join(tmp_dir, ARCHITECTPATHS.CLI_CONFIG_FILENAME);
     fs.writeJSONSync(tmp_config_file, config);
-    const app_config_stub = sinon.stub().resolves(new AppService(tmp_dir));
+    const app_config_stub = sinon.stub().resolves(new AppService(tmp_dir, '0.0.1'));
```
dan-barrett, line 67: Will these need to be changed as the CLI gets updated or is `0.0.1` more of a "mock version" for testing purposes only?

My only comment would be that we consolidate these references if we're having to update them frequently, but I think this pattern is fine if they stay `0.0.1` forever.

src/base-command.ts
```diff
@@ -15,13 +15,9 @@ export default abstract class extends Command {
...
   };
 
-  private getFormattedJSON(obj: object) {
-    return JSON.stringify(obj, null, 2);
-  }
-
   async init() {
     if (!this.app) {
-      this.app = await AppService.create(this.config.configDir);
+      this.app = await AppService.create(this.config.configDir, this.config.userAgent.split(/\/|\s/g)[2]);
```
dan-barrett, line 20: Nice 👍 , presumably this will grab this: `0.4.10-rc.1`?

Worth sending the entire userAgent in another header (eg. `@architect-io/cli/0.4.10-rc.1 darwin-x64 node-v13.3.0`) for analytics purposes? scope creep...

test/commands/build.test.ts
ryan-cahill: I don't really see why we would need to change them since these are just tests, and this isn't related to any of the tests

src/base-command.ts
ryan-cahill: I had done that originally, but didn't really think the extra information was that useful. When we do, maybe we can send it to Segment or something, but that data currently has no place to go anyway

src/app-config/service.ts
ryan-cahill: I had been unsure about that at first, but sources say the `X-` prefix is basically deprecated

https://specs.openstack.org/openstack/api-wg/guidelines/headers.html

dan-barrett: Wow... would you look at that! Good call, and good to know.

src/base-command.ts
dan-barrett: Fair enough... this is me just being greedy. Agreed, probably not worth it for now.

Will be useful someday when we encounter issues that we can't reproduce and want to know environment details, but we're probably a long ways away from that right now since we can just ask the friendly what their userAgent is.

dan-barrett:
@tjhiggins @ryan-cahill approved from me

PR #236, opened 2020-07-01
add register command for components

dan-barrett:


src/commands/register.ts
```diff
@@ -2,142 +2,161 @@ import { flags } from '@oclif/command';
...
-export default class ServiceRegister extends Command {
-  static aliases = ['service:register', 'services:register', 'svcs:register', 'svc:register'];
-  static description = 'Register a new Service with Architect Cloud';
+export default class ComponentRegister extends Command {
+  static aliases = ['component:register', 'components:register', 'c:register', 'comp:register'];
+  static description = 'Register a new Component with Architect Cloud';
 
   static flags = {
     ...Command.flags,
-    services: flags.string({
```
ryan-cahill, line 26: Should this still exist if someone wants to build/push only one service of the component rather than all of them? That could probably save some people a lot of time in CI jobs and when iterating on individual services

src/commands/register.ts
```diff
@@ -2,142 +2,161 @@ import { flags } from '@oclif/command';
...
-      char: 's',
-      description: 'Path to a service to build',
+    components: flags.string({
```
ryan-cahill, line 26: Should this be singular?

src/commands/register.ts
```diff
@@ -2,142 +2,161 @@ import { flags } from '@oclif/command';
...
+      description: 'Path to a component to build',
       exclusive: ['environment'],
       multiple: true,
     }),
     environment: flags.string({
       char: 'e',
-      description: 'Path to an environment config including local services to build',
-      exclusive: ['service', 'image'],
+      description: 'Path to an environment config including local components to build',
+      exclusive: ['component', 'image'],
```
ryan-cahill, line 35: Do we need `'image'` in this array?

src/commands/register.ts
```diff
@@ -2,142 +2,161 @@ import { flags } from '@oclif/command';
...
-          tag: flags.tag,
-          digest: digest,
-          image: image_without_tag,
-          config: classToPlain(node.service_config),
-        };
-        cli.action.start(chalk.blue(`Registering service ${node.service_config.getName()}:${flags.tag} with Architect Cloud...`));
-        await this.post_service_to_api(service_dto, selected_account.id);
-        cli.action.stop(chalk.green(`Successfully registered service`));
-      }
+    const [account_name, _] = component_config.getRef().split('/');
```
ryan-cahill, line 78: Is there a reason for the `_`? Why not get index 0 of the split value?

src/commands/register.ts
```diff
@@ -2,142 +2,161 @@ import { flags } from '@oclif/command';
...
+      config: classToPlain(component_config),
+    };
+
+    cli.action.start(chalk.blue(`Registering component ${component_config.getName()}:${tag} with Architect Cloud...`));
+    await this.post_component_to_api(component_dto, selected_account.id);
+    cli.action.stop(chalk.green(`Successfully registered component`));
+  }
+
+
+  private async build_docker_image_if_necessary(config_path: string, service_config: ServiceConfig, image_tag: string) {
```
ryan-cahill, line 101: Should this be named `push_image_if_necessary` or something? I feel like the pushing here is the more important part

src/commands/register.ts
```diff
@@ -2,142 +2,161 @@ import { flags } from '@oclif/command';
...
+    const digest = await getDigest(image).catch(e => {
+      cli.action.stop(chalk.red(`Inspect failed`));
+      throw new Error(`The image is not reachable by docker: ${image}`);
+    });
+    cli.action.stop(chalk.green(`Image verified`));
+  }
+
+
+  private async post_component_to_api(dto: PutComponentVersionDto, account_id: string): Promise<any> {
+    const { data: component_version } = await this.app.api.post(`/accounts/${account_id}/components`, dto);
```
ryan-cahill, line 159: Can this just go on the line where it's used above? It also doesn't look like anything really needs to be returned

src/commands/register.ts
dan-barrett: I thought the same thing, but previously `services` was plural so I kept it the same.

Paging @davidthor for opinion:

```bash
architect register --components ./architect.yml
## OR
architect register --component architect.yml

## ...keeping in mind that it accepts multiple paths:
architect register --components ./architect.yml ../another/architect.yml
```

davidthor: I think the `--components` flag is going to be exponentially less useful with the current design. Related services are likely to be in the same component making it unlikely you'd be publishing multiple components at once.

src/commands/register.ts
dan-barrett: Nope. Fixed.

src/commands/register.ts
dan-barrett: This is an interesting comment. The backend as we currently designed it I don't think would support this... _unless_ we wrapped the `Service` in a component and persisted it _as a_ `Component`.

I'd be curious to get @tjhiggins 's thoughts.

But the short answer is: I think this flag would be nonsensical for now without some reconsideration of the API.

davidthor: I think the best approach for us long term is to treat our components as OCI compliant images. OCI registries already solve build-time optimization and partial builds via layering meaning we wouldn't need the user to do anything in particular for this. We'd just need components to be conformant (which they aren't right now) so we can store them in a registry. I'd recommend we avoid any obligations on the users side with the working assumption that we will do something like that one day.

src/commands/register.ts
dan-barrett: I don't know. I think this was just habit. I like destructuring. But I did change cause you're right that's obviously cleaner.

src/commands/register.ts
```diff
@@ -2,142 +2,155 @@ import { flags } from '@oclif/command';
...
+      component_config.getServiceByRef(service_ref)?.setImage(image);
     }
+
+    const component_dto = {
+      tag: tag,
+      config: classToPlain(component_config),
+    };
+
+    cli.action.start(chalk.blue(`Registering component ${component_config.getName()}:${tag} with Architect Cloud...`));
+    await this.app.api.post(`/accounts/${selected_account.id}/components`, component_dto);
```
tjhiggins, line 96: We can talk after standup, but this will post the transformed config - ideally we would store the unaltered config.

I'll merge for now, but we should address this.

dan-barrett: For this: do you think we need a `loadUnmodifiedConfigFromPath(config_path)` as part of the `ComponentConfigBuilder`? Because basically what I'm hearing is we just want to do a yml parse or json parse and send as is correct?

tjhiggins: you're hearing correct :/ 

I already added a `readFromPath` to the componentconfigbuilder

PR #237, opened 2020-07-02
register raw config

dan-barrett:


dan-barrett:
@tjhiggins this pushes the raw config instead of the "built" ComponentConfig. The problem here is that presumably as the raw config schema changes, there's no mechanism by which "raw config" is sensibly typed and tied to the (versioned) `ComponentConfig` . I added basic types to get it working with a `TODO` comment, but I think this requires a closer look and conversation.

One option is that the `ComponentConfig` entity actually has the raw_config attached to it such that running `ComponentConfigBuilder.buildFromX()` doesn't actually obscure the version that the user wrote. Created a ticket for this problem here: https://gitlab.com/architect-io/architect-cli/-/issues/104

PR #239, opened 2020-07-07
separate platform create and optional create environment in deploy command

ryan-cahill:


src/commands/platforms/create.ts
```diff
@@ -0,0 +1,166 @@
+import { flags } from '@oclif/command';
```
tjhiggins, line 1: Looks great. Having David product review. Can you add tests for the platform and environment cmds?

Prob worth covering both that CI flow (all values are passed via cmd) and the user flow with inquirer.
https://stackoverflow.com/questions/49862039/how-to-write-unit-tests-for-inquirer-js

ryan-cahill:
@davidthor I updated and added a test for the command `architect platform:create`. Feel free to pull and try again

davidthor:
Looking good, but running into an issue with a deployment. Not sure if it's related to this ticket or components overall, so CCing @tjhiggins too.

I tried to deploy this repo using a basic architect deploy of the config:
https://github.com/davidthor/ory-kratos-test

```sh
$ architect deploy -p davidthor/test -e davidthor/my-env environment.yml
Creating deployment... !
 ›   Error: /environments/eef37090-5d34-4d3d-9242-750d342db3c8/deploy (404)
 ›   statusCode: 404
 ›   message: Could not find component version: 
 ›   'file/undefined:/Users/davidthor/Workspaces/Architect/ory-kratos-test/server'
 ›   timestamp: 2020-07-08T22:50:52.379Z
 ›   path: /environments/eef37090-5d34-4d3d-9242-750d342db3c8/deploy
```

@tjhiggins it could simply be do to trying to use `file:` references in a deployment targeting a remote environment?

tjhiggins:
Are you deploying against dev or your local?

davidthor:
Dev and to an architect public environment.

tjhiggins:
I think that makes sense - obviously we should either infer latest tag or error saying you cannot deploy remotely with local refs. What would you prefer?

ryan-cahill:
In the case of a 404, should we tell the user that the component needs to be registered first, then register and deploy? It makes sense to me that this would fail as things stand, but we know what we have to do and can guide the user

davidthor:
> I think that makes sense - obviously we should either infer latest tag or error saying you cannot deploy remotely with local refs. What would you prefer?
> 
> edit1:
> Inferring latest tag feels a little weird since I could register concourse/ci:5.8.0 and not have a latest.
> 
> edit2:
> We could infer the most recent tag registered, but that might be a little dicey/non-deterministic

Definitely more conversations to be had regarding how versions, tags, releases work and how we do inference. I think in the short term its ok to infer latest when nothing is specified. Even if that means that a component can be registered without a latest tag.

> In the case of a 404, should we tell the user that the component needs to be registered first, then register and deploy? It makes sense to me that this would fail as things stand, but we know what we have to do and can guide the user

We definitely need better error messaging because I had no idea, but it's also totally fine if we can't build and register services directly from an environment config for now. We'll have to scope a story where we can do build operations as part of deploy flows, but that can come later.

PR #254, opened 2020-07-21
Fusionauth example

dan-barrett:
Still some open lines of work here that might prove interesting (ie integrating with ory client per @davidthor 's recommendation). Still an unknown (but seemingly non-blocking) java exception on startup and would like to do some more digging on that at some point. But for now moving onto other high priority stories in this sprint per @tjhiggins


davidthor:
@tjhiggins this was an interesting case where we definitely need to update documentation. @dan-barrett tried to create two interfaces in the environment that map to the same service (listening on multiple ports). While this _should_ work (and does in remote deploys I believe), this is a known limitation of the nginx gateway we're using for local deploys that we failed to document.  I'm creating a ticket for us to detect this and spit it out in the logs prior to a deployment. 

@dan-barrett two notes on this at a glance:

1. Per my comments above, you can simply remove the `mgmt` interface from the environment and auth.localhost will resolve as expected!
2. The `fusionauth.properties` file is causing the console errors because it has bad contents, but these errors are ignored in the long term because you successfully override the values via environment parameters. If you remove the volume this error will go away, and it appears that configuration via config file is redundant with configuration via environment parameters. You only need one or the other, not both.

examples/fusionauth/environment.external-postgres.yml
```diff
@@ -0,0 +1,25 @@
...
+        interfaces:
+          main:
+            host: host.docker.internal
+            protocol: postgresql
+            port: 5432
+
+      fusionapi:
+        volumes:
+          config:
+            host_path: ./fa_config
```
tjhiggins, line 24: This could be defined in the debug block for the service

examples/fusionauth/architect.yml
```diff
@@ -0,0 +1,81 @@
...
+      POSTGRES_PASSWORD: ${ parameters.DATABASE_PASSWORD }
+      POSTGRES_DB: fusionauth
+    interfaces:
+      main:
+        port: 5432
+        protocol: postgresql
+
+  fusionapi:
+    description: Fusion API
+    language: go
```
ryan-cahill, line 54: Wasn't this a Java service?

dan-barrett: Haha! Indeed it was. I thought it was go early on because Matt Gowie's client was in go but clearly forgot to change it when I realized it was java. Thanks

examples/fusionauth/environment.with-mgmt-interface.yml
```diff
@@ -0,0 +1,19 @@
+interfaces:
+  auth: ${ components['architect/fusionauth'].interfaces.http.url }
+  mgmt: ${ components['architect/fusionauth'].interfaces.mgmt.url }
```
tjhiggins, line 3: This should work now. If you pull in the latest cli.

dan-barrett:
@tjhiggins two interfaces works, this is good to merge

@ryan-cahill good call on killing image first

examples/fusionauth/environment.external-postgres.yml
dan-barrett: The config file (and thus mounted volume) is needed for all deploys, local and remote, so it doesn't belong in the debug block. Created ticket: https://gitlab.com/architect-io/architect-cli/-/issues/116

PR #255, opened 2020-07-21
Remove @Transforms from config specs

tjhiggins:


src/dependency-manager/src/component-config/base.ts
```diff
@@ -10,7 +10,7 @@ export interface ParameterDefinitionSpec {
 }
 
 export abstract class ComponentConfig extends BaseSpec {
-  abstract __version: string;
+  abstract __version?: string;
```
dan-barrett, line 13: Why optional now? Or should it have always been that way?

src/dependency-manager/src/component-config/base.ts
```diff
@@ -20,9 +20,12 @@ export abstract class ComponentConfig extends BaseSpec {
   abstract getKeywords(): string[];
   abstract getAuthor(): string;
   abstract getParameters(): Dictionary<ParameterDefinitionSpec>;
```
dan-barrett, line 22: Worth adding convenience `getParameter(key: string)` methods to all these that return Dictionaries?

src/dependency-manager/src/component-config/base.ts
```diff
@@ -50,7 +53,22 @@ export abstract class ComponentConfig extends BaseSpec {
     return classToClass(this);
   }
 
+  expand() {
```
dan-barrett, line 56: I like this a lot. 

My static instincts would have me write this method like this:

```
static expand(config: ComponentConfig);
```
And call it like this:

```
const expanded_config = ComponentConfig.expand(some_config);
```

This makes it clear that `config.expand()` isn't actually modifying any state on the `config`, which is what I was initially wondering. However, I know we don't generally follow that pattern so consider this a soft recommendation.

src/dependency-manager/src/component-config/base.ts
```diff
@@ -50,7 +53,22 @@ export abstract class ComponentConfig extends BaseSpec {
     return classToClass(this);
```
dan-barrett, line 53: Can we add return types to `copy()` and `expand()`?

src/dependency-manager/src/component-config/builder.ts
```diff
@@ -149,6 +149,12 @@ export class ComponentConfigBuilder {
       }
     }
 
+    if (config.debug && (typeof config.debug === 'string' || config.debug instanceof Array)) {
```
dan-barrett, line 152: Why do we need this new exceptional case suddenly?

src/dependency-manager/src/component-config/v1.ts
```diff
@@ -25,16 +25,40 @@ export type ParameterValueSpecV1 = string | number | boolean | ParameterDefiniti
...
 
 export interface ComponentContextV1 {
   dependencies: Dictionary<ComponentContextV1>;
   parameters: Dictionary<ParameterValue>;
-  interfaces: Dictionary<ServiceInterfaceSpec>;
+  interfaces: Dictionary<InterfaceSpec>;
   services: Dictionary<ServiceContextV1>;
 }
 
+export function transformServices(input?: Dictionary<object | ServiceConfigV1>): Dictionary<ServiceConfigV1> {
```
dan-barrett, line 38: So it looks like we just moved this here from `service-config/v1.ts` because it's on the the getter now which is here, correct?

I'm increasingly a fan of this pattern.

src/dependency-manager/src/component-config/base.ts
tjhiggins: We don't use and it muddles the output when you convert to json.

src/dependency-manager/src/component-config/builder.ts
tjhiggins: Removing support for some old shorthand where we let you set
debug: \<debug docker start command\>

src/dependency-manager/src/component-config/v1.ts
tjhiggins: We used to use to transform dependencies in service-config, but now it is only used in component-config.

src/dependency-manager/src/component-config/base.ts
tjhiggins: I'm not sure this is any better than getParameters()[key]

src/dependency-manager/src/component-config/base.ts
tjhiggins: Let me know if this helps and you like the use of JSDoc annotations. Thinking we should use more for critical code.
https://github.com/architect-team/architect-cli/pull/255/commits/261ba6a936fc89cf2166af45685be46c2fa9e4fd#diff-7cc7e5ad27691772a28e4dd5c45ae372R28

Capture

tjhiggins: More info here: https://code.visualstudio.com/docs/languages/typescript#_jsdoc-support

src/dependency-manager/src/component-config/base.ts
dan-barrett: Could we just kill it then?

src/dependency-manager/src/component-config/base.ts
dan-barrett: Fair enough as long as getParameters() never returns null or undefined.

src/dependency-manager/src/component-config/base.ts
dan-barrett: Yes, I do like JSDocs to clarify this. It still doesn't replace it being static (and literally barring it from modifying underlying object state) but it does help clarify intent.

dan-barrett:
These are good to go from my end, solid improvement IMO.

src/dependency-manager/src/component-config/base.ts
tjhiggins: We can, but trying to avoid another data migration. Any properties we remove would require it.

PR #257, opened 2020-07-28
87 consolidate normalizedref

dan-barrett:


src/common/utils/validation.ts
```diff
@@ -12,6 +12,8 @@ export class SemvarValidator implements StringValidator {
   }
 }
 
+// TODO:87: consolidate these with new Slug validators
```
dan-barrett, line 15: I'd like to get an rc out to make my API changes easier. Note there are still some small cleanup TODO's in this PR though.

src/dependency-manager/src/component-config/base.ts
```diff
@@ -43,12 +44,14 @@ export abstract class ComponentConfig extends ConfigSpec {
     return `${this.getRef()}-interfaces`;
   }
 
-  getComponentVersion() {
-    return this.getRef().split(':')[1];
+  //TODO:87:why are these implemented in base.ts instead of v1.ts
```
dan-barrett, line 47: I noticed a couple of these methods are implemented here in the `base.ts` instead of `v1.ts`. Any particular reason?

src/dependency-manager/src/component-config/v1.ts
```diff
@@ -151,12 +153,15 @@ export class ComponentConfigV1 extends ComponentConfig {
   @Transform((value) => !value ? {} : value)
   interfaces?: Dictionary<InterfaceSpecV1 | string>;
 
-  getName() {
-    return this.name.split(':')[0];
+  //TODO:87:can we add getComponentName(), getAccountName(), getTag()?
```
dan-barrett, line 156: Would you be opposed to adding two methods here: `getAccountName()` and `getComponentName()`?

One of my goals with this is that we're not trying to do a `split('/')` everywhere else across the codebase. But if we only expose the `getName()` method here, that's what we're going to end up doing again.

src/dependency-manager/src/utils/slugs.ts
```diff
@@ -0,0 +1,241 @@
+
+export type ArchitectSlug = string; // simply string that passes: ArchitectSlugValidator (ie "account-name")
+export type ComponentSlug = string; // "account-name/component-name"
+export type ComponentTag = string; // "tag"
+export type ComponentVersionSlug = string; // "account-name/component-name:tag"
+export type ServiceSlug = string; // "account-name/component-name/service-name"
+export type ServiceVersionSlug = string; // "account-name/component-name/service-name:tag"
+export type EnvironmentSlug = string; // "@environment-account-name/environment-name"
+export type ServiceInstanceSlug = string; // "account-name/component-name/service-name:tag@environment-account-name/environment-name"
```
dan-barrett, line 9: Note: I added support for a theoretical `ServiceInstance` slug which we don't yet use, but which I suspect we may find useful. It is globally unique (ie across environments), and is a friendly string that represents a deployed service including the environment.

dan-barrett:
@ryan-cahill - I remember, say 6 months ago, you were seeing `config:get` failing in Github Actions on `npm run test` only on Mac OS. But it was not failing on either of our local machines. We had a convo in slack about it but that's long buried and I don't remember what became of it. Any chance you remember what that was? It's happening again on this PR.

https://github.com/architect-team/architect-cli/runs/917344674#step:11:900

ryan-cahill:
Yeah it was just timing out. I increased the timeout for that particular set of tests from the default five seconds to ten, and things were fine

dan-barrett:
@ryan-cahill - was talking through the slug changes with @tjhiggins earlier and wanted to get some of your opinions.

Currently we're doing this:

`account--component--service--md5hash`

For example: `architect--cloud--db--y8pscont`

I'm hesitant to limit any given slug to just 20 characters. This is too low (even makes `rest-addition-service` an invalid name). So I loosened the naming limits. But that introduces the problem on the back side: resource names in AWS become too long. 

`architect--rest-addition-service--primary-db--y8pscont`

The result is that we need to abridge slugs as part of the normalized-ref. Initially I opted to simply lop off the latter part of each slug (after the account name), I'll call this **Option 1**:

`architect--rest-a--primar--y8pscont`

But then it occurred to me that this might obscure similar names (ie a `primary-mysql` and a `primary-elasticsearch` would become indistinguishable):

```
architect--rest-a--primar--y8pscont // primary-mysql
architect--rest-a--primar--aoaqhrdv // primary-elasticsearch
```

So then I tried abridging the word in the middle, I'll call this **Option 2**:

```
architect--res-vice--prim-ysql--y8pscont // primary-mysql
architect--rest-vice--prim-arch--aoaqhrdv // primary-elasticsearch
```

**Option 3** is that we kill the component/service names altogether and add a characters to the hash:

```
architect--y8pscontba0
architect--aoaqhrdv9cn
```

There are obviously all sorts of variants on this. The MD5 I think pretty much guarantees collision prevention, so I'm not so much worried about uniqueness as I am about readability. It's clear that we may have to throw resolvability out the window. Preferences? Other ideas?

One categorically different idea that I'm open to is that we ditch the idea of trying to be consistent across all platforms and specifically have a method to convert the normalized ref `to_aws_name()` and then we can forget about limits altogether on the normalized ref.

dan-barrett:
@tjhiggins , @ryan-cahill -

Continuing to think about this last night, and I think normalized ref is simply trying to serve too many masters. Proposal:

* Kill `normalized_ref()` entirely
* only use `ref` throughout architect
* at platform-specific edges of our system, use platform-specific transforms to play nicely under different requirements:

* the AWS terraform service can be where that logic lives for AWS: `#to_aws_ref(ref: ServiceRef)`. 
* likewise in Kubernetes, keep this somewhere in the kubernetes terraform service: `#to_kubernetes_ref(ref: ServiceRef)` 
* likewise in docker_compose: `#to_docker_compose_ref(ref: ServiceRef)`

ryan-cahill:
@dan-barrett I was thinking the same thing. We obviously want everything to work with the best UX possible as it relates to naming for resources and it would be unfortunate to have to bring down the UX of k8s and compose to conform with AWS, which won't matter to a lot of customers. As long as we can do the transform at the exact edge of the terraform generation, I like those six bullets

dan-barrett:
@tjhiggins  this is good to merge again. Changes-

* removed abbreviation logic; simply lopping off any characters at the end of the the string (before the hash)
* removed normalized_ref; calling it `url_safe_ref` throughout because i think that is more descriptive than `normalized_ref`, though we can change this back or to something else if you prefer.
* to_url_safe() takes a max length now
* combined the validators and killed all the duplicates
* removed hard references to the new format in `components.test.ts` and `interfaces.test.ts` and simply transform them, so hopefully no more changing the tests there if we need to iterate on this again.

src/dependency-manager/src/utils/slugs.ts
```diff
@@ -0,0 +1,270 @@
...
+export type GatewaySlug = 'gateway'; // the string-literal
+export type InterfacesSlug = string; // <account-name>/<component-name>:<tag>-interfaces
+
+export type SlugKind = 'component' | 'component_version' | 'service' | 'service_version' | 'environment' | 'service_instance' | 'gateway' | 'interfaces';
+
+/**
+ * Represents the unique elements of a fully parsed slug. This object will have different fields depending on the
+ * type of slug (ie was it a tagged component vs an untagged component vs a service slug)
+ */
+export interface ParsedSlug {
```
tjhiggins, line 19: Do we still need these interfaces? Seems heavy - especially if we want to be able to sanitize any string value

src/dependency-manager/src/utils/refs.ts
```diff
@@ -0,0 +1,156 @@
...
+  private static to_digest(uri: string): string {
+    return crypto.createHash('md5').update(uri)
+      .digest("base64") // base64 adds entropy in a more compact string
+      .toLowerCase() // we need to makes everything lower which unfortunately removes some entropy
+      .replace(/[\\/+=]/g, ''); // we also remove occurances of slash, plus, and equals to make url-safe
+  }
+
+  // ordered from most specific to least specific
+  private static OrderedEntityParsers: { [key in SlugKind]: EntityParser<any> } = {
+    'service_instance': {
```
tjhiggins, line 115: These should probably be class objects - would prefer description, validator, splitter were grouped as opposed to all being top level.

src/commands/deploy.ts
```diff
@@ -256,23 +256,15 @@ export default class Deploy extends Command {
   }
 
   validateEnvironmentNamespacedInput(value: string) {
-    const value_split = value.split('/');
-    if (value_split.length !== 2) {
-      return 'Environment name must be in the form my-account/environment-name';
-    }
-    if (!EnvironmentNameValidator.test(value_split[0]) || !EnvironmentNameValidator.test(value_split[1])) {
-      return `Each part of name must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character`;
+    if (!Slugs.EnvironmentSlugValidator.test(value)) {
```
tjhiggins, line 259: Looks like we need to also consolidate with the other regex for:
```
export const IMAGE_REGEX = '[a-zA-Z0-9-_]+';
export const REPOSITORY_REGEX = `${IMAGE_REGEX}\\/${IMAGE_REGEX}`;
// https://github.com/docker/distribution/blob/v2.7.1/reference/reference.go#L15
export const TAG_REGEX = '[\\w][\\w\\.-]{0,127}';
export const REPOSITORY_TAG_REGEX = `${REPOSITORY_REGEX}(?::${TAG_REGEX})?`;
```

src/dependency-manager/src/utils/slugs.ts
tjhiggins: Are these only used in the case of a splitter?

src/commands/deploy.ts
tjhiggins: You'll see that the regex ends up getting enforced when we call 
```
errors = await validateDictionary(expanded, 'services', errors, undefined, { ...options, groups: (options.groups || []).concat('component') }, new RegExp(`^${IMAGE_REGEX}$`));
```

dan-barrett:
@tjhiggins -

FYI, got the regexes consolidated finally and broke out the Slugs class into separate classes like you recommended. I'm not 100% happy with how the breakout came out so I'm going back through and tweaking it a bit to make it more consistent, but just thought I'd ping you here in case you want to review before I get too much further.

dan-barrett:
@tjhiggins this is good to merge again

src/dependency-manager/src/utils/slugs.ts
dan-barrett: Copying this from offline from slack for posterity:

yes, those are used as the return values of the splitters. I think they will be useful. That is what saves us from having to do:

`ref.split('/')[2]` when we want to get at the service_name say.
Instead, it's: `const { service_name} = Slugs.splitServiceSlug(ref);` which feels both more typesafe and readable

dan-barrett:
@tjhiggins , done with changes we talked about. killed `try_split_slug`

PR #258, opened 2020-07-28
adding ARC_ env var support and --parameter flag

ryan-cahill:


src/commands/deploy.ts
```diff
@@ -84,6 +84,12 @@ export default class Deploy extends Command {
       hidden: true,
       exclusive: ['account', 'environment', 'auto_approve', 'lock', 'force_unlock', 'refresh'],
     }),
+    parameter: flags.string({
+      description: 'Component parameters',
+      exclusive: ['local', 'compose_file'],
```
tjhiggins, line 89: I think we can remove the exclusive aspect. Should still apply to local deploys.

src/commands/deploy.ts
```diff
@@ -180,6 +193,14 @@ export default class Deploy extends Command {
       }
     }
 
+    const extra_params = this.getExtraEnvironmentVariables(flags.parameter);
+    for (const param_name of Object.keys(extra_params)) {
+      if (env_config.getParameters()[param_name] === undefined) {
+        throw new Error(`Parameter ${param_name} not found in env config`);
+      }
+    }
+    env_config.setParameters(extra_params);
```
tjhiggins, line 202: We are still posting the raw_config below. I think its safe to post env_config now - since I deferred the @Transform. 
Make sure to test.

PR #261, opened 2020-08-05
Account scoped

tjhiggins:


src/commands/deploy.ts
```diff
@@ -180,17 +175,31 @@ export default class Deploy extends Command {
...
-    // Validate env config
-    const env_config = await EnvironmentConfigBuilder.buildFromPath(env_config_path);
-    for (const [ck, cv] of Object.entries(env_config.getComponents())) {
-      if (cv.getExtends()?.startsWith('file:')) {
-        this.error(`Cannot deploy component remotely with file extends: ${ck}: ${cv.getExtends()}`);
+    let env_config: EnvironmentConfig;
+
+    let merge: boolean;
+    if (args.environment_config_or_component.includes(':')) {
+      const [name, tag] = args.environment_config_or_component.split(':');
```
ryan-cahill, line 186: Should we error check this for at least an amount of split parts or something? just the condition to get into this code feels very fragile

src/common/utils/platform.ts
```diff
@@ -0,0 +1,52 @@
...
+
+interface Platform {
+  id: string;
+  name: string;
+}
+
+export class PlatformUtils {
+  static flags = {
+    platform: flags.string({
+      description: 'Architect Platform',
```
ryan-cahill, line 14: Is this not getting a char so we don't overwrite parameters?

src/common/utils/account.ts
```diff
@@ -0,0 +1,45 @@
...
+    } else {
+      inquirer.registerPrompt('autocomplete', require('inquirer-autocomplete-prompt'));
+      let accounts: Account[] = [];
+      const answers: any = await inquirer.prompt([
+        {
+          type: 'autocomplete',
+          name: 'account',
+          message: 'Select an account',
+          filter: (x) => x, // api filters
+          source: async (answers_so_far: any, input: string) => {
```
ryan-cahill, line 32: Is this supposed to be paginated automatically? This currently just cycles through the same ten choices. I assume it's the same for the other utils

src/commands/register.ts
```diff
@@ -87,8 +86,12 @@ export default class ComponentRegister extends Command {
   private async register_component(config_path: string, tag: string) {
     const { raw_config } = await ComponentConfigBuilder.rawFromPath(config_path);
 
+    if (!raw_config.name) {
+      throw new Error('Config must have a name');
```
ryan-cahill, line 90: "Component config" maybe, just in case someone tries an env config by mistake

src/common/utils/account.ts
tjhiggins: It doesn't support pagination unfortunately - you can filter down by typing.

At some point we can try to add pagination, but it hopefully won't be a big deal early on.
I don't expect people to have a lot of accounts/platforms - maybe envs
https://github.com/mokkabonna/inquirer-autocomplete-prompt

ryan-cahill: Can we include in the message for each one of these classes something like "Select an account, type to search" or something then? When someone hits the up or down arrow, the prompt "(Use arrow keys or type to search)" disappears and I can see this being confusing for some users. The only instructions left are to use the arrow keys to move up or down

Screenshot from 2020-08-05 12-58-56
Screenshot from 2020-08-05 12-59-09

tjhiggins: Yeah its not ideal. I was trying to avoid forking/writing our own, but feel free to story/prioritize improvements

ryan-cahill: Looked into it and in order to make a change to that actual line we'd have to fork the autocomplete module to update the Paginator used [here](https://github.com/mokkabonna/inquirer-autocomplete-prompt/blob/master/index.js#L14), then probably copy [this file](https://github.com/SBoudrias/Inquirer.js/blob/132682ecfb3972b0efc9b11a370e649a4a8b49c6/packages/inquirer/lib/utils/paginator.js#L42) into the forked module to update the text. Seems excessive for a text change but also pretty straightforward. On the other hand I didn't think including a little instruction in the prompt message was that bad. I just want to avoid people getting frustrated

PR #278, opened 2020-09-02
Parameter default file refs in env config

ryan-cahill:


src/common/dependency-manager/local-manager.ts
```diff
@@ -95,7 +95,23 @@ export default class LocalDependencyManager extends DependencyManager {
         }
         component_config.setService(service_ref, service);
       }
+
+      for (const [param_key, param_value] of Object.entries(component_config.getParameters())) {
```
tjhiggins, line 99: Do we also need to make these changes when registering a component? 
If so we should probably add tests for register that checks the posted body on the register call.

Makes me think that a purely regex replacement might be better as opposed to spot changes. Potentially would allow us to use the same code for register/deploy. I know we had talked about a regex approach did you rule that out?

tjhiggins (changes requested):

src/common/dependency-manager/local-manager.ts
ryan-cahill: I liked the idea of it because I think adding the file ref replacement in the builders because can prevent a lot of problems down the line related to relative file paths. I was pretty concerned about it though because of the weird ways that it could go wrong. Hard to say that the regexes for json and yaml are bulletproof, but I added a bunch of tests to be more sure, and will probably add more

src/dependency-manager/src/utils/files.ts
```diff
@@ -0,0 +1,37 @@
+import fs from 'fs';
+import path from 'path';
+import untildify from 'untildify';
+
+export const insertFileDataFromRefs = (file_contents: string, file_path: string, config_path: string) => {
```
tjhiggins, line 5: Any reason we need to support yml and json? I'd rather this run in the buildFromJson methods. Stringify the json, regex replace, and parse.

tjhiggins (changes requested):

src/dependency-manager/src/utils/files.ts
ryan-cahill: I had considered that at one point, but that would cause validation error line numbers to be off for yaml files. I could've looked into some kind of normalizing for that, but it seemed worse than just doing the replace for the actual file type, json or yaml

tjhiggins: Hmm this took me a sec. I would have thought the opposite. Just to confirm I would expect the following:

```yaml
parameters:
  config: file:./config.json
interfaces:
  invalid: ${{ bug }}
```

Validation error line 4

If we inline multi-line file contents won't this lead to a higher line number?

src/dependency-manager/src/component-config/builder.ts
```diff
@@ -69,25 +70,27 @@ export class ComponentConfigBuilder {
...
+        raw_config = yaml.safeLoad(updated_file_contents);
       } catch { }
     }
 
     if (!raw_config) {
       throw new Error('Invalid file format. Must be json or yaml.');
     }
 
-    return { file_path, file_contents, raw_config };
+    return { file_path, file_contents: updated_file_contents, raw_config };
```
tjhiggins, line 93: Can you try updating raw_config via insertFileDataFromRefs (stringify, insert, parse)? If we keep the file_contents as is then the line numbers should be correct (def add a test). Keep the logic in the readFromPath since we wouldn't want to allow for file: reads server-side.

ryan-cahill: I updated this from what I think you mean, but I wanted to double check that you meant `rawFromPath` instead of `readFromPath` because the line you commented on here is from `rawFromPath`

PR #280, opened 2020-09-02
adding ability to set reusable blocks in yaml env and component configs

ryan-cahill:


src/dependency-manager/src/environment-config/builder.ts
```diff
@@ -54,6 +54,12 @@ export class EnvironmentConfigBuilder {
       throw new Error('Invalid file format. Must be json or yaml.');
     }
 
+    for (const key of Object.keys(js_obj)) {
```
tjhiggins, line 57: Should both of these go in buildFromJSON? Would hopefully prevent having to have client-side changes https://gitlab.com/architect-io/architect-hub/-/merge_requests/268/diffs.

tjhiggins (changes requested):

src/dependency-manager/src/component-config/builder.ts
```diff
@@ -104,6 +104,12 @@ export class ComponentConfigBuilder {
   }
 
   static buildFromJSON(obj: any): ComponentConfig {
+    for (const key of Object.keys(obj)) {
```
tjhiggins, line 107: Both of these should go below the safeguard check `if (!(obj instanceof Object)) {`

tjhiggins (changes requested):

PR #283, opened 2020-09-09
dockerfile relative to context for register command

ryan-cahill:


src/common/docker-compose/index.ts
```diff
@@ -107,7 +107,7 @@ export const generate = async (dependency_manager: LocalDependencyManager): Prom
 
         if (build.dockerfile) {
           // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          compose.services[url_safe_ref].build!.dockerfile = build.dockerfile;
+          compose.services[url_safe_ref].build!.dockerfile = path.join(path.dirname(dependency_manager.config_path), build.context || '.', build.dockerfile);
```
tjhiggins, line 110: I think this should be component_path since the user is setting the dockerfile in the architect.yml?

tjhiggins: Can you add a test as well.

tjhiggins (changes requested):

src/common/docker-compose/index.ts
ryan-cahill: So this is actually a different problem from the one Dan was seeing. I'm still looking into what he found. Local deployments are run with environment files and that seems to be the likely place you would want to override a Dockerfile, so that's why I made this relative to those (ex. `architect deploy -l env-local.json`)

tjhiggins: We would need to support both environment/architect like volumes. His examples were in the architect.yml.

ryan-cahill: Because this wasn't Dan's original problem and I didn't want to clutter this PR so we can get it merged, I removed that code and I'm going to make a separate ticket

PR #285, opened 2020-09-15
docker compose converter

ryan-cahill:


davidthor:
This is great! Could you change the command to the following?

```sh
$ architect init --from-compose "docker-compose.yml"
```

ryan-cahill:
@davidthor any other thoughts here?

test/commands/init.test.ts
```diff
@@ -0,0 +1,336 @@
...
+
+  test
+    .do(ctx => mockAuth())
+    .finally(() => sinon.restore())
+    .stdout({ print })
+    .stderr({ print })
+    .command(['init', '--help'])
+    .it('succinctly describes the init command', ctx => {
+      expect(ctx.stdout).to.contain('Initialize an architect component from an existing docker-compose file\n')
+    });
```
davidthor, line 27: A test for help text contents might be overkill. More likely to get in the way then to be helpful.

davidthor (approved):
Looks good!

tjhiggins:
> @davidthor any other thoughts here?
On me to review. Remind me again on Monday if I don't get to it this weekend.

src/commands/init.ts
```diff
@@ -0,0 +1,258 @@
...
+      description: 'Path where the component file should be written to',
+      default: 'architect.yml',
+    }),
+    account: flags.string({
+      char: 'a',
+    }),
+    name: flags.string({
+      char: 'n',
+    }),
+    'from-compose': flags.string({
```
tjhiggins, line 36: I could have sworn that dashes didn't work, but great if they do. However, we should probably be consistent with dash or underscore.

PR #288, opened 2020-09-18
adding circleci preview orb

ryan-cahill:


circleci-orbs/preview.yml
```diff
@@ -0,0 +1,94 @@
+description: Easily register and deploy a component to a new environment on the Architect Cloud, then destroy the component and environment when ready
+
+display:
+  home_url: https://architect.io
+  source_url: https://github.com/architect-team/architect-cli/circleci-orbs
```
davidthor, line 5: This URL resolves to 404. You might be looking for this:

```
https://github.com/architect-team/architect-cli/tree/master/circleci-orbs
```

circleci-orbs/preview.yml
```diff
@@ -0,0 +1,94 @@
...
+        type: string
+      account:
+        type: string
+      environment:
+        type: string
+      component_name:
+        type: string
+    executor: preview_executor
+    steps:
+      - run: npm install -g @architect-io/cli@0.7.4
```
davidthor, line 49: @tjhiggins what're your thoughts on version pinning down to the patch level here? Seems a bit restrictive for pushing updates.

tjhiggins: We did the same thing github actions. I was torn then and now. The only comp I have is githubs own actions like `actions/cache@v2` which only major version pin. Our spec has become less volatile. Can you npm install a specific major version like `npm install -g @architect-io/cli@0.7` or even `npm install -g @architect-io/cli@0.*`?

davidthor: You can with NPM by default. See this doc:
https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept

I'm all for limiting the depth of versioning for the published actions/orbs, but since we're constantly changing the CLI with updates, I'd hate to have to keep going back to all our orbs/actions and updating version pins like this that could have been upgraded silently. What I suggested offline to @ryan-cahill was that we pin to minor versions of the CLI while in beta (e.g. @architect-io/cli@0.7) and eventually major versions which in theory are the only ones with breaking changes. 

_Here's an interesting NPM version calculator for confirmation of interpretation: https://semver.npmjs.com/_

ryan-cahill:
Minor issue with publishing actions automatically by the way

https://github.community/t/cannot-publish-an-action-with-the-api/133577

davidthor:
> Minor issue with publishing actions automatically by the way
> 
> https://github.community/t/cannot-publish-an-action-with-the-api/133577

Am I to understand that creating a release on the Github repo automatically publishes new actions versions? If so, that's probably a fine way for the actions to be updated instead of having to maintain our own actions workflow to trigger it.

.github/workflows/publish-preview.yml
```diff
@@ -0,0 +1,21 @@
...
+  publish_circleci_orb:
+    name: Publish CircleCI Orb
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: circleci-orb
+      - run: sudo snap install circleci
+      - run: circleci orb validate ./circleci-orbs/preview.yml --host https://circleci.com --token ${{ secrets.CIRCLE_TOKEN }}
+      - run: circleci orb publish ./circleci-orbs/preview.yml architect/preview@dev:master --host https://circleci.com --token ${{ secrets.CIRCLE_TOKEN }}
```
tjhiggins, line 20: How come we publish dev:master and not for new cli tags?

circleci-orbs/preview.yml
```diff
@@ -0,0 +1,95 @@
...
+        type: string
+      component_file:
+        type: string
+        default: ./architect.yml
+      deploy_flags:
+        type: string
+        default: ''
+    executor: preview_executor
+    steps:
+      - run: npm install -g @architect-io/cli@0.7
```
tjhiggins, line 30: Can we parameterize the install? Auto publishing seems kind of mute if we are pinned to a specific version of our cli.

circleci-orbs/preview.yml
```diff
@@ -0,0 +1,95 @@
...
+      - run: architect login -e << parameters.email >> -p << parameters.password >>
```
tjhiggins, line 31: use the `ARCHITECT_PASSWORD` env instead of explicitly setting -p since it will show up in logs. 

if we do this for the github action - can you create an issue to address there too?

circleci-orbs/preview.yml
```diff
@@ -0,0 +1,95 @@
...
+      - run: architect register << parameters.component_file >> -t << parameters.environment >>
+      - run: architect environment:create << parameters.environment >> -a << parameters.account >> --platform << parameters.platform >>
+      - run: architect deploy --auto_approve -a << parameters.account >> -e << parameters.environment >> << parameters.component_name >>:<< parameters.environment >> << parameters.deploy_flags >>
```
tjhiggins, line 34: I'm confused by all of the `>>` `<<` is that really necessary? Any other way to specify run commands?

circleci-orbs/preview.yml
```diff
@@ -0,0 +1,95 @@
...
+                account: example-account
+                environment: preview-$CIRCLE_BUILD_NUM
+                platform: example-platform
+                component_name: example-account/hello-world
+  destroy_preview:
+    description: Destroy the deployed component and the specified environment
+    usage:
+      version: 2.1
+      orbs:
+        preview: architect-team/preview@1.0.0
```
tjhiggins, line 83: what is `architect-team/preview@1.0.0`?

tjhiggins (changes requested):

circleci-orbs/preview.yml
ryan-cahill: Parameterize it as in change the version to be installed? What if the flags/args change?

circleci-orbs/preview.yml
ryan-cahill: This is an example of how to use the orb https://circleci.com/docs/2.0/orb-author/#providing-usage-examples-of-orbs

circleci-orbs/preview.yml
ryan-cahill: This is the interpolation syntax they use. What would you suggest instead?

circleci-orbs/preview.yml
tjhiggins: It would be queued off of the cli tag so if the flags changed we would have had to update the orb since its in the same repo. 

We could potentially add a test to test the orb with the current version of the cli before publishing.

circleci-orbs/preview.yml
tjhiggins: Gotcha I had no idea it looking at it - their docs don't use spaces, but we do for our own interpolation so I guess this is fine.

.github/workflows/publish-preview.yml
ryan-cahill: David wanted us to only publish new versions when this file was changed

circleci-orbs/preview.yml
ryan-cahill: Lets talk about this in our quick meeting tomorrow

PR #289, opened 2020-09-28
Arc variable updates

ryan-cahill:


dan-barrett (approved):
Tested locally, this fixes the issue @tjhiggins . Thanks @ryan-cahill

src/commands/deploy.ts
```diff
@@ -309,11 +309,20 @@ export default class Deploy extends DeployCommand {
...
 
-  getExtraEnvironmentVariables(parameters: string[]) {
+  getExtraEnvironmentVariables(parameters: string[], env_config?: EnvironmentConfig) {
     const extra_env_vars: { [s: string]: string | undefined } = {};
-    for (const [param_name, param_value] of Object.entries(process.env || {})) {
-      if (param_name.startsWith('ARC_')) {
-        extra_env_vars[param_name.substring(4)] = param_value;
+
+    if (env_config) {
+      for (const param_name of Object.keys(env_config.getParameters())) {
```
tjhiggins, line 316: I don't think we want to look at the env config parameters. If I deploy
```
components:
  examples/react-demo: latest
```

It will error if there are required parameters. I'd then add `ARC_xyz` to my ci env and then re-run. I wouldn't want to have to edit the environment yaml and add the missing parameters.

tjhiggins (changes requested):

src/commands/deploy.ts
dan-barrett: Ah, hah. I noticed that and briefly discussed this with @ryan-cahill . We decided this is what you had been referring to last week regarding the environment parameters and "auto-passthrough".

I assumed this was intentional and that it would be more of a feature change to modify the current behavior.

I'm behind modifying this behavior in theory, but I think you get into sticky questions with colliding parameter names (ie if you had two components in the environment spec).

tjhiggins: Gotcha - I'll create another ticket to try and capture all of the edge cases.

PR #292, opened 2020-10-16
Cpu and memory

ryan-cahill:


src/dependency-manager/src/service-config/v1.ts
```diff
@@ -253,6 +253,14 @@ export class ServiceConfigV1 extends ServiceConfig {
   @Type(() => BuildSpecV1)
   build?: BuildSpecV1;
 
+  @IsOptional({ always: true })
+  @IsNumber()
+  cpu?: number;
```
tjhiggins, line 258: Everything needs to be a string because of interpolation

ex
`cpu: ${{ parameters.cpu }}`

replace IsNumber with `@Type(() => String)`
Probably should do the same for memory

src/common/docker-compose/index.ts
```diff
@@ -86,6 +83,15 @@ export const generate = async (dependency_manager: LocalDependencyManager): Prom
           ...compose.services[url_safe_ref],
         };
       }
+
+      const cpu = node.node_config.getCpu();
+      const memory = node.node_config.getMemory();
+      if (cpu || memory) {
+        const service = compose.services[url_safe_ref];
+        service.deploy = { resources: { limits: {} } };
+        if (cpu) { service.deploy.resources.limits.cpus = cpu.toString(); }
```
tjhiggins, line 92: Do these limits apply if you don't set the `--compatability` flag for the compose
verify via `docker stats`

tjhiggins (changes requested):

PR #294, opened 2020-10-20
Compose environment variables in commands

ryan-cahill:


src/common/docker-compose/index.ts
```diff
@@ -75,8 +75,12 @@ export const generate = async (dependency_manager: LocalDependencyManager): Prom
     }
 
     if (node instanceof ServiceNode) {
-      if (node.node_config.getCommand().length) compose.services[url_safe_ref].command = node.node_config.getCommand();
-      if (node.node_config.getEntrypoint().length) compose.services[url_safe_ref].entrypoint = node.node_config.getEntrypoint();
+      if (node.node_config.getCommand().length) { // docker-compose expects environment variables used in commands/entrypoints to be prefixed with $$, not $ in order to use variables local to the container
+        compose.services[url_safe_ref].command = node.node_config.getCommand().map(command_part => command_part.replace(/"\$/g, '"$$$'));
+      }
+      if (node.node_config.getEntrypoint().length) {
+        compose.services[url_safe_ref].entrypoint = node.node_config.getEntrypoint().map(entrypoint_part => entrypoint_part.replace(/"\$/g, '"$$$'));
```
tjhiggins, line 82: Did you confirm that this was a problem for entrypoints as well?

src/common/docker-compose/index.ts
```diff
@@ -75,8 +75,12 @@ export const generate = async (dependency_manager: LocalDependencyManager): Prom
     }
 
     if (node instanceof ServiceNode) {
-      if (node.node_config.getCommand().length) compose.services[url_safe_ref].command = node.node_config.getCommand();
-      if (node.node_config.getEntrypoint().length) compose.services[url_safe_ref].entrypoint = node.node_config.getEntrypoint();
+      if (node.node_config.getCommand().length) { // docker-compose expects environment variables used in commands/entrypoints to be prefixed with $$, not $ in order to use variables local to the container
+        compose.services[url_safe_ref].command = node.node_config.getCommand().map(command_part => command_part.replace(/"\$/g, '"$$$'));
```
tjhiggins, line 79: This might be too targeted. I'd expect the following to work as well.
`echo "Test: $XYZ"`

Not sure what to do about the following:
`echo $XYZ`
This would work with escaping on docker-compose and already works on k8s. Not sure why it doesn't work on ECS.

tjhiggins (changes requested):

src/common/docker-compose/index.ts
ryan-cahill: If that's only a problem with ECS, should we just update within the API and not worry about quotes here?

src/common/docker-compose/index.ts
ryan-cahill: Yeah it was also a problem with entrypoints

src/common/docker-compose/index.ts
tjhiggins: Specifically `echo $XYZ` doesn't work on ECS. `echo "$XYZ"` does work according to Dan. We still need to escape $ for docker compose regardless of that edge case.

PR #295, opened 2020-10-21
enforcing strings as dependency tags

ryan-cahill:


test/dependency-manager/components.test.ts
```diff
@@ -319,7 +319,7 @@ describe('components spec v1', function () {
           }
         },
         dependencies: {
-          'concourse/ci': 6.2
+          'concourse/ci': '6.2'
```
tjhiggins, line 322: Does this break any of our examples?

src/dependency-manager/src/utils/validators/dictionary_types.ts
```diff
@@ -0,0 +1,29 @@
+import { registerDecorator, ValidationArguments, ValidationOptions } from "class-validator";
+
+/** Require each value of the object is of one of the types specified */
+export const DictionaryTypes = (types: string[], validationOptions?: ValidationOptions) => {
```
tjhiggins, line 4: We probably don't need to support a list of types.
I like the solution though. Better to be explicit.

test/dependency-manager/components.test.ts
ryan-cahill: Nope, all of our examples use the `latest` tag

PR #296, opened 2020-10-27
Reject circular dependencies

ryan-cahill:


src/dependency-manager/src/manager.ts
```diff
@@ -127,13 +128,16 @@ export default abstract class DependencyManager {
...
-        await this.loadComponent(graph, dep_component, component_map);
-      }
+    for (const [dep_key, dep_value] of Object.entries(component.getDependencies())) {
+      const dep_name = dep_value.includes(':') ? `${dep_key}:latest` : `${dep_key}:${dep_value}`;
+      const dep_extends = dep_value.includes(':') ? dep_value : `${dep_key}:${dep_value}`;
+      const dep_component = ComponentConfigBuilder.buildFromJSON({ extends: dep_extends, name: dep_name });
+
+        if (dep_component.getRef() in component_map) { // Reject circular dependencies
+          if (loaded_by_env.includes(dep_component.getRef())) continue;
+          throw new Error(`Circular component dependency detected (${ component.getRef() })`);
```
tjhiggins, line 138: Can the error include the other component as well?
```ts
throw new Error(`Circular component dependency detected (${ component.getRef() } <> ${ dep_component.getRef() })`);
```

src/dependency-manager/src/manager.ts
```diff
@@ -127,13 +128,16 @@ export default abstract class DependencyManager {
...
-        const dep_component = ComponentConfigBuilder.buildFromJSON({ extends: dep_extends, name: dep_name });
-        await this.loadComponent(graph, dep_component, component_map);
-      }
+    for (const [dep_key, dep_value] of Object.entries(component.getDependencies())) {
+      const dep_name = dep_value.includes(':') ? `${dep_key}:latest` : `${dep_key}:${dep_value}`;
+      const dep_extends = dep_value.includes(':') ? dep_value : `${dep_key}:${dep_value}`;
+      const dep_component = ComponentConfigBuilder.buildFromJSON({ extends: dep_extends, name: dep_name });
+
+        if (dep_component.getRef() in component_map) { // Reject circular dependencies
+          if (loaded_by_env.includes(dep_component.getRef())) continue;
```
tjhiggins, line 137: Instead of `loaded_by_env` can you check the dep_component.getDependencies() directly since its already loaded? I think we only one to reject direct dependencies not implicit.

Add a test to make sure we don't reject the following: 
i.e. 
Component A depends on Component B and C
Component B depends on C

tjhiggins (changes requested):

src/dependency-manager/src/manager.ts
```diff
@@ -127,13 +126,19 @@ export default abstract class DependencyManager {
...
-        const dep_component = ComponentConfigBuilder.buildFromJSON({ extends: dep_extends, name: dep_name });
-        await this.loadComponent(graph, dep_component, component_map);
+    for (const [dep_key, dep_value] of Object.entries(component.getDependencies())) {
+      const dep_name = dep_value.includes(':') ? `${dep_key}:latest` : `${dep_key}:${dep_value}`;
+      const dep_extends = dep_value.includes(':') ? dep_value : `${dep_key}:${dep_value}`;
+      const dep_component = ComponentConfigBuilder.buildFromJSON({ extends: dep_extends, name: dep_name });
+
+      if (component_map[dep_component.getRef()]) {
+        const first_side_dependency = ComponentVersionSlugUtils.parse(component.getRef()).namespaced_component_name in component_map[dep_component.getRef()].getDependencies();
+        const second_side_dependency = ComponentVersionSlugUtils.parse(dep_component.getRef()).namespaced_component_name in component.getDependencies();
```
tjhiggins, line 136: Pretty sure `second_side_dependency` is redundant since you're looping over the component dependencies.

tjhiggins (changes requested):

PR #297, opened 2020-10-29
add error handling for register

dan-barrett:
Closes: https://gitlab.com/architect-io/hub-api/-/issues/185

### DIFF
--- a/src/common/utils/account.ts
+++ b/src/common/utils/account.ts
@@ -20,6 +20,9 @@ export class AccountUtils {
     let account: Account;
     if (account_name) {
       account = (await api.get(`/accounts/${account_name}`)).data;
+      if (!account) {
+        throw new Error(`Could not find account=${account_name}`);
+      }
     } else {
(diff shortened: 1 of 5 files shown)


dan-barrett:
@tjhiggins these are good to go

src/common/utils/account.ts
```diff
@@ -20,6 +20,9 @@ export class AccountUtils {
     let account: Account;
     if (account_name) {
       account = (await api.get(`/accounts/${account_name}`)).data;
+      if (!account) {
+        throw new Error(`Could not find account=${account_name}`);
```
tjhiggins, line 24: Is it possible for no account to be returned? I would have assumed the api would throw a 404

tjhiggins (changes requested):

src/common/utils/account.ts
dan-barrett: Yes, potentially unnecessary. I added it to cover all bases to guarantee that this method returned an account or threw (so its callers don't have to check if `account` is defined).

But if you think we can rely on the API for that in this case I can remove.

PR #298, opened 2020-10-29
Single component deployment interfaces for dependencies

ryan-cahill:


test/commands/deploy.test.ts
```diff
@@ -168,4 +344,103 @@ describe('remote deploy environment', function () {
...
+        "echo-dos": "main"
+      }
+    }
+  ];
+
+  let deploy_spy: SinonSpy;
+  mockArchitectAuth.stub(DeployCommand, 'POLL_INTERVAL', () => { return 0 })
+    .do(ctx => {
+      deploy_spy = sinon.fake.returns(undefined);
+      sinon.replace(DeployCommand.prototype, 'deployRemote', deploy_spy);
```
tjhiggins, line 424: Don't use sinon in the do since the restore might not fire if an expect fails. 
You can use stub without defining outside of the test scope:
```ts
mockArchitectAuth
    .stub(docker, 'buildImage', sinon.stub().returns('repostory/account/some-image:1.0.0'))
    ...
    .it('test file: replacement', ctx => {
      const buildImage = docker.buildImage as sinon.SinonStub;
      expect(buildImage.notCalled).to.be.true;
    });
```

src/commands/deploy.ts
```diff
@@ -243,19 +244,20 @@ export default class Deploy extends DeployCommand {
       dependency_manager = await LocalDependencyManager.create(this.app.api);
       dependency_manager.environment = env_config;
 
+      this.setExtraEnvironmentVariables(flags.parameter, env_config);
+
       const extra_interfaces = this.getExtraInterfaces(flags.interface);
-      this.updateEnvironmentInterfaces(env_config, extra_interfaces, parsed_component_version.namespaced_component_name);
+      const edges = (await dependency_manager.getGraph()).edges;
+      this.updateEnvironmentInterfaces(dependency_manager.environment, extra_interfaces, edges);
```
tjhiggins, line 251: I would have expected the interfaces to be set in the env config before the dep-manager/graph is created. Editing the edges directly doesn't feel right.

tjhiggins (changes requested):

src/commands/deploy.ts
ryan-cahill: How else would we get dependencies and their interfaces to map back to without generating the graph?

PR #299, opened 2020-10-29
Custom domains

ryan-cahill:


src/dependency-manager/src/service-config/v1.ts
```diff
@@ -72,6 +72,11 @@ export class InterfaceSpecV1 extends BaseSpec {
 
   @IsOptional({ always: true })
   url?: string;
+
+  @IsOptional({ always: true })
+  @IsArray({ always: true })
+  @IsUrl({}, { always: true, each: true })
+  domains?: Set<string>;
```
tjhiggins, line 79: Is the type here actually a set?
I think you need something like:
`@Type(() => Set)`

tjhiggins: Even using the `@Type` might not work since the domain is being set outside the plainToClass
https://github.com/architect-team/architect-cli/pull/299/files#diff-de9be30d94b13b89791f6b04e91deb40f0729a7000d7e3cc1327c61a4183832eR96

tjhiggins: Maybe add a quick test to make sure its actually a set?

tjhiggins (changes requested):

PR #303, opened 2020-11-02
add tasks to component spec

dan-barrett:
@tjhiggins - how does this look directionally? Once you think it's safe to merge, I can get to work on the API.

I've left a number of `TODO:84` scattered here that we're going to want to return to (mostly interfaces and references), but theoretically what I have in this PR should give us enough to go on to get TaskNodes deploying remotely.

### DIFF
--- a/src/dependency-manager/src/component-config/v1.ts
+++ b/src/dependency-manager/src/component-config/v1.ts
@@ -34,7 +34,6 @@ interface ServiceContextV1 {
 
 interface TaskContextV1 {
   environment: Dictionary<string>;
-  interfaces: Dictionary<InterfaceSpec>;
 }
 
 export interface ComponentContextV1 {
@@ -72,9 +71,6 @@ export function transformServices(input?: Dictionary<object | ServiceConfigV1>):
 export function transformTasks(input?: Dictionary<object | TaskConfigV1>): Dictionary<TaskConfigV1> {
   if (!input) {
     return {};
-  }
-  if (!(input instanceof Object)) {
-    return input;
   }
 
   const output: any = {};
@@ -177,8 +173,8 @@ export class ComponentConfigV1 extends ComponentConfig {
   @Transform((value) => !value ? {} : value)
   services?: Dictionary<ServiceConfig>;
 
-  @IsOptional({ groups: ['operator'] })
-  @IsOptional({ groups: ['developer'] }) //TODO:84: should this become optional? should (tasks||services) be required?
+  @IsOptional()
+  @IsObject()
   @Transform((value) => !value ? {} : value)
   tasks?: Dictionary<TaskConfig>;
 
@@ -351,15 +347,7 @@ export class ComponentConfigV1 extends ComponentConfig {
 
     const tasks: Dictionary<TaskContextV1> = {};
     for (const [tk, tv] of Object.entries(this.getTasks())) {
-      const interfaces: Dictionary<InterfaceSpec> = {};
-      for (const [ik, iv] of Object.entries(tv.getInterfaces())) {
-        interfaces[ik] = {
-          ...interface_filler,
-          ...iv,
-        };
-      }
       tasks[tk] = {
-        interfaces,
         environment: tv.getEnvironmentVariables(),
       };
     }
--- a/src/dependency-manager/src/component-config/v1.ts
+++ b/src/dependency-manager/src/component-config/v1.ts
@@ -1,12 +1,14 @@
 import { plainToClass, serialize, Transform } from 'class-transformer';
-import { Allow, IsBoolean, IsNotEmptyObject, IsObject, IsOptional, IsString, Matches, ValidatorOptions } from 'class-validator';
+import { Allow, IsBoolean, IsObject, IsOptional, IsString, Matches, ValidatorOptions } from 'class-validator';
 import { ParameterValue, ServiceConfig } from '..';
 import { InterfaceSpec } from '../service-config/base';
 import { InterfaceSpecV1, ServiceConfigV1, transformParameters } from '../service-config/v1';
+import { TaskConfig } from '../task-config/base';
+import { TaskConfigV1 } from '../task-config/v1';
 import { BaseSpec } from '../utils/base-spec';
 import { Dictionary } from '../utils/dictionary';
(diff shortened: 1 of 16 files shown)


src/dependency-manager/src/component-config/v1.ts
```diff
@@ -312,7 +380,10 @@ export class ComponentConfigV1 extends ComponentConfig {
     const expanded = this.expand();
     errors = await validateDictionary(expanded, 'parameters', errors, undefined, options, new RegExp(`^${Slugs.ComponentParameterRegexBase}$`));
     errors = await validateDictionary(expanded, 'services', errors, undefined, { ...options, groups: (options.groups || []).concat('component') }, new RegExp(`^${Slugs.ArchitectSlugRegexNoMaxLength}$`));
+    errors = await validateDictionary(expanded, 'tasks', errors, undefined, { ...options, groups: (options.groups || []).concat('component') }, new RegExp(`^${Slugs.ArchitectSlugRegexNoMaxLength}$`));
     errors = await validateDictionary(expanded, 'interfaces', errors, undefined, options);
+    errors = await validateCrossDictionaryCollisions(expanded, 'services', 'tasks', errors); // makes sure services and tasks don't have any common keys
```
tjhiggins, line 385: Not a huge fan of this - why do we need this enforcement if the refs are different for task nodes?

src/dependency-manager/src/component-config/v1.ts
```diff
@@ -141,17 +173,22 @@ export class ComponentConfigV1 extends ComponentConfig {
   parameters?: Dictionary<ParameterValueSpecV1>;
 
   @IsOptional({ groups: ['operator'] })
-  @IsNotEmptyObject({ groups: ['developer'] })
+  @IsOptional({ groups: ['developer'] }) //TODO:84: should this become optional? should (tasks||services) be required?
   @Transform((value) => !value ? {} : value)
   services?: Dictionary<ServiceConfig>;
 
+  @IsOptional({ groups: ['operator'] })
+  @IsOptional({ groups: ['developer'] }) //TODO:84: should this become optional? should (tasks||services) be required?
```
tjhiggins, line 181: Add `@IsObject()`
Then you don't need the last if check in `transformTasks`

src/dependency-manager/src/component-config/v1.ts
```diff
@@ -297,11 +349,27 @@ export class ComponentConfigV1 extends ComponentConfig {
...
+    const tasks: Dictionary<TaskContextV1> = {};
+    for (const [tk, tv] of Object.entries(this.getTasks())) {
+      const interfaces: Dictionary<InterfaceSpec> = {};
+      for (const [ik, iv] of Object.entries(tv.getInterfaces())) {
+        interfaces[ik] = {
+          ...interface_filler,
+          ...iv,
+        };
+      }
+      tasks[tk] = {
```
tjhiggins, line 361: I don't know if we want the tasks in the interpolation context

src/dependency-manager/src/component-config/v1.ts
```diff
@@ -297,11 +349,27 @@ export class ComponentConfigV1 extends ComponentConfig {
       };
     }
 
+    const tasks: Dictionary<TaskContextV1> = {};
+    for (const [tk, tv] of Object.entries(this.getTasks())) {
+      const interfaces: Dictionary<InterfaceSpec> = {};
```
tjhiggins, line 354: This probably isn't necessary since we will block interfaces for tasks

src/dependency-manager/src/component-config/v1.ts
```diff
@@ -30,11 +32,17 @@ interface ServiceContextV1 {
   interfaces: Dictionary<InterfaceSpec>;
 }
 
+interface TaskContextV1 {
+  environment: Dictionary<string>;
+  interfaces: Dictionary<InterfaceSpec>;
```
tjhiggins, line 37: remove interfaces

src/dependency-manager/src/manager.ts
```diff
@@ -359,6 +375,7 @@ export default abstract class DependencyManager {
         context[normalized_ref].architect = architect_context;
       }
 
+      // TODO:84:add support for references to tasks
```
tjhiggins, line 378: Add another todo to the register cmd to build/push images if there is a build context.

src/dependency-manager/src/service-config/v1.ts
```diff
@@ -246,7 +246,7 @@ export class ServiceConfigV1 extends ServiceConfig {
   environment?: Dictionary<string>;
 
   @IsOptional({ groups: ['operator', 'debug'] })
-  @IsObject({ groups: ['developer'], message: 'interfaces must be defined even if it is empty since the majority of services need to expose ports' })
+  @IsObject({ groups: ['developer'], message: 'interfaces must be defined even if it is empty since the majority of services need to expose ports' }) //TODO:84: this assumption becomes less applicable with tasks
```
tjhiggins, line 249: Maybe we have a base config that is shared between task/service? Then we can define Service specific fields at a higher level. This would probably be a nightmare for versioning. @davidthor  thoughts?

src/dependency-manager/src/component-config/v1.ts
```diff
@@ -141,17 +173,22 @@ export class ComponentConfigV1 extends ComponentConfig {
   parameters?: Dictionary<ParameterValueSpecV1>;
 
   @IsOptional({ groups: ['operator'] })
-  @IsNotEmptyObject({ groups: ['developer'] })
+  @IsOptional({ groups: ['developer'] }) //TODO:84: should this become optional? should (tasks||services) be required?
```
tjhiggins, line 176: You can make it optional. Consolidate it down to:
`@IsOptional()`

tjhiggins (changes requested):

src/dependency-manager/src/component-config/v1.ts
tjhiggins: Changed my mind - add tasks to the interpolation context

dan-barrett:
@tjhiggins this is good to merge again; still has some TODO:84's that we can follow on tomorrow, but this should unstick me from API.

dan-barrett:
@tjhiggins  the cut-release pipeline failed on one OS test (passed the others). I re-kicked the pipeline and it's failing on the same test on a different OS. Re-kicked it again to see what happens.

FYI - This happened with my other PR the other day for a different story and tests passed on second try. I'll create a bug with details and look after we get Tasks out.

PR #304, opened 2020-11-03
export TaskNode and TaskConfig

dan-barrett:
@tjhiggins we also need to export the TaskNode/Config.

Try merging this one in and let's hope tests pass. If not, I think we need to resolve (or remove) that nondeterministic test `deploy.test.ts#describe('local deploy environment', ...)` because the release will block me this AM. I'm still moving forward on the terraform side but soon I'll want to bridge them.

### DIFF
--- a/src/dependency-manager/src/index.ts
+++ b/src/dependency-manager/src/index.ts
@@ -7,7 +7,9 @@ export * from './environment-config/base';
 export * from './environment-config/builder';
 export * from './graph/node';
 export * from './graph/node/service';
+export * from './graph/node/task';
 export * from './service-config/base';
+export * from './task-config/base';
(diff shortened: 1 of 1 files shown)


dan-barrett:
🔥 🔥 🔥

dan-barrett:
@tjhiggins  heads up- all checks passing here but looks like the master pipeline is failing again with the same test (on Windows 11 only). I'm realizing the package is being published anyway, so I'm not going to be blocked on Tasks for now. But I think that one test is still a problem.

dan-barrett:
Created this issue. I will look into it after I finish Tasks:

https://gitlab.com/architect-io/architect-cli/-/issues/156

PR #307, opened 2020-11-04
add resource spec interface to share between TaskConfig and ServiceCo…

dan-barrett:
@tjhiggins this refactors the TaskConfig and the ServiceConfig to both implement a ResourceSpec interface with common properties, rather than having the TaskConfig extend the ServiceConfig.

I'm open to alternative OO patterns here to achieve the competing goals of using inheritance for versioning but also wanting to keep track of common properties.

### DIFF
--- a/src/dependency-manager/src/task-config/v1.ts
+++ b/src/dependency-manager/src/task-config/v1.ts
@@ -1,37 +1,45 @@
 import { Type } from 'class-transformer';
-import { Allow, IsEmpty, IsInstance, IsOptional, IsString } from 'class-validator';
-import { ServiceConfigV1 } from '../service-config/v1';
+import { IsEmpty, IsInstance, IsOptional, IsString, ValidatorOptions } from 'class-validator';
+import { ResourceConfigV1 } from '../common/v1';
+import { validateDictionary, validateNested } from '../utils/validation';
 import { TaskConfig } from './base';
 
-//TODO:84:Tasks: let's rethink our abstraction patterns a bit here. There's a chance we don't want TaskConfig extending ServiceConfig
-export class TaskConfigV1 extends ServiceConfigV1 implements TaskConfig {
-  @Allow({ always: true })
-  __version?: string;
-
-  // TODO:84:Tasks: are there any cron parsers that we can use to validate?
-  @IsOptional({ always: true })
-  @IsString({ always: true })
-  schedule?: string;
-
+export class TaskConfigV1 extends ResourceConfigV1 implements TaskConfig {
   @Type(() => TaskConfigV1)
   @IsOptional({ always: true })
   @IsInstance(TaskConfigV1, { always: true })
   @IsEmpty({ groups: ['debug'] })
   debug?: TaskConfigV1;
 
-  getSchedule(): string {
-    return this.schedule || '';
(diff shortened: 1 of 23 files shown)


src/dependency-manager/src/task-config/v1.ts
```diff
@@ -1,30 +1,136 @@
...
   @IsString({ always: true })
-  schedule?: string;
+  @Matches(/^[a-zA-Z0-9-_]+$/, {
+    message: 'Names must only include letters, numbers, dashes, and underscores',
+  })
+  name?: string;
+
+  @IsOptional({ always: true })
+  @IsString({ always: true })
+  description?: string;
```
tjhiggins, line 24: Would prefer to not have all of this duplication of validation annotations. Can we have a ResourceSpecV1?

tjhiggins (changes requested):

src/dependency-manager/src/task-config/v1.ts
dan-barrett: **TLDR;** I'm open to alternatives, but I explored your proposed one and it leads to multiple inheritance.

The problem with the above is we have competing design concerns here between using inheritance for versioning (`*Spec extends *V1`) and using inheritance to re-use properties/annotations (`Task/Service* extends Resource*`):

ie:
```
// to re-use properties/annotations:
ResourceConfig extends BaseConfig
ServiceConfig extends ResourceConfig
TaskConfig extends ResourceConfig

// to version:
ResourceConfig extends BaseConfig
ResourceConfigV1 extends ResourceConfig

// which means we're stuck with:
TaskConfigV1 extends ?
ServiceConfigV1 extends ?
```
We might want TaskConfigV1 to extend both ResourceV1 _and_ TaskConfig but Typescript doesn't support multiple inheritance (I think for good reason... this is problematic):

```
TaskConfigV1 extends TaskConfig, ResourceConfigV1
ServiceConfigV1 extends ServiceConfig, ResourceConfigV1
```

So my solution was to use an interface to at least track the shared properties, and stick with using inheritance for versioning:
```
// use inheritence for versioning:
ServiceConfigV1 extends ServiceConfig
TaskConfigV1 extends TaskConfig

// use interface for tracking shared properties
interface ResourceConfig

ServiceConfig implements ResourceConfig
TaskConfig implements ResourceConfig
```

tjhiggins: Can we do?
`ServiceConfigV1 extends ResourceConfigV1  implements ServiceConfig`
`TaskConfigV1 extends ResourceConfigV1  implements TaskConfig`

dan-barrett:
@tjhiggins good to merge again pending tests.

There is one remaining TODO around interfaces/edges, but I think that is going to be a bigger discussion for tomorrow AM.

I want to get this in to fix API. This commit is stable and follows your outlined solution above. I'm relatively happy with this solution.

dan-barrett:
Rerunning tests... upping the timeout of one of our non-deterministically failing tests. Creating an issue for that also.

dan-barrett:
https://gitlab.com/architect-io/architect-cli/-/issues/159

dan-barrett:
@tjhiggins this is good to review and merge again.

PR #310, opened 2020-11-09
fixing bug when a numerical extends is used

ryan-cahill:


src/dependency-manager/src/spec/component/component-transformer.ts
```diff
@@ -16,6 +16,11 @@ export const transformComponents = (input?: Dictionary<any>, parent?: any): Dict
   // eslint-disable-next-line prefer-const
   for (let [key, value] of Object.entries(input)) {
     if (!value) value = {};
+
+    if (typeof value?.extends === 'number') {
+      value.extends = value.extends.toString();
```
tjhiggins, line 21: Does this fix the issue? Feels like we should still run into an issue of yaml dropping a 0. Can you add a test for the following to confirm:

```yml
components:
  examples/hello-world: 1.0
```

tjhiggins: Also noticed that extends has a IsString() validator. How does that not fail before getting to transformComponents?

```
@IsOptional({ always: true })
  @IsString({ always: true })
  @Matches(/^(?!file:).*$/g, { groups: ['developer'], message: 'Cannot hardcode a filesystem location when registering a component' })
  extends?: string;
```

tjhiggins (changes requested):

src/dependency-manager/src/spec/component/component-transformer.ts
ryan-cahill: There's a point here where 
```
components:
  ryan-cahill-a0/hello-world:
    extends: 1.10
```
becomes
```
components: {
  'ryan-cahill-a0/hello-world': { extends: 'ryan-cahill-a0/hello-world:1.10' }
},
```

I feel like now that you've asked that though, we shouldn't be updating the value to a string. It would be strange for this to accept a number when the other case throws a validation error. Should I throw a validation error in this case as well?

tjhiggins: Can you add a test case? I'd be okay rejecting - still unclear how 1.10 maintains the trailing zero. I thought that was a problem with the json parser.

ryan-cahill: Sorry, in that example above I was only demonstrating the transformation in the extends where we end up with the extends being the entire component/tag combo. You're right that the trailing zero would have been removed there though. I added a test case though where we do retain the trailing zeros

PR #329, opened 2020-12-15
Fixing kafka example

ryan-cahill:


examples/kafka/architect.yml
```diff
@@ -3,39 +3,67 @@ name: examples/kafka
 parameters:
   TOPIC_NAME:
     default: architect
+  KAFKA_LISTENERS:
```
tjhiggins, line 6: Not super familiar with kafka. Why do these need to be parameters? 

When would I override the following:
- KAFKA_LISTENERS
- INTERNAL_KAFKA_ADDR
- EXTERNAL_KAFKA_ADDR

examples/kafka/architect.yml
```diff
@@ -3,39 +3,67 @@ name: examples/kafka
...
+    default: INTERNAL://:9092,EXTERNAL://:9093
+  KAFKA_ADVERTISED_LISTENERS:
+    default: INTERNAL://:9092,EXTERNAL://${{ services.kafka.interfaces.external.host }}:${{ services.kafka.interfaces.external.port }}
+  KAFKA_LISTENER_SECURITY_PROTOCOL_MAP:
+    default: INTERNAL:PLAINTEXT,EXTERNAL:PLAINTEXT
+  KAFKA_ADVERTISED_HOST_NAME:
```
tjhiggins, line 12: What is KAFKA_ADVERTISED_HOST_NAME?

examples/kafka/architect.yml
```diff
@@ -3,39 +3,67 @@ name: examples/kafka
...
+    required: false
+  SUBSCRIBE_DELAY:
```
tjhiggins, line 14: @davidthor Not a huge fan of the single timeout on startup - should our default behavior more mimic ecs/k8s

Looks like we can set delays/max restarts:
https://docs.docker.com/compose/compose-file/#restart_policy

examples/kafka/architect.yml
ryan-cahill: `KAFKA_LISTENERS` is overridden for local. The other two don't need to be overridden, but they would have been defined in two places otherwise. I guess that doesn't make it appropriate for a parameter though, so I'll update it

examples/kafka/architect.yml
ryan-cahill: That's overridden for local and is the host address. I'll change it in the config to be the address people using macOS would use though

examples/kafka/architect.yml
tjhiggins: Yeah setting the values in the debug block might make more sense.

examples/kafka/architect.yml
ryan-cahill: The restart policy seems like it will restart and delay a container when it dies? Shouldn't we be preventing it from dying in the first place with something like a check to see if we can hit some kind of endpoint related to the topic?

tjhiggins: True it's not ideal. The only way I can think of is to append a wait-for-it.sh call to the command or entrypoint. We could bundle with the cli and mount as a volume to each container. A little wary of editing the command though. Open to alternatives.

davidthor: I haven't thought of a good, non-invasive way to handle this either. I have the feeling that long term it's not unreasonable for the different platforms to support `depends_on` like fields that adhere to health checks to verify availability, but we might be waiting a while for consistency on that one. 

I do see a lot of projects that have configured retry/backoff policies for connecting to their own databases, so perhaps its not entirely unreasonable for that logic to have to live inside the services for now. I wish we could do more, but hard to do so without risking failure.

ryan-cahill: I haven't found a great way to do this with this example. There are scripts which come bundled with Kafka that would get us the data we need, but an installation of Java is also required to be able to run them. Seems heavy for our example. I guess I could write a wait-for-it.js where I check for topics before starting the main script? But I hardly feel like that's any better

davidthor: I think something like the wait-for-it script is pretty much standard and I'd be fine including it in a project explicitly. I'm just not sold on automating it from our deployments to check the health of dependencies. If you want to try instrumenting it explicitly in this sample that wouldn't be horrible.

tjhiggins: Sold on being explicit in this example

examples/kafka/subscriber/wait-for-it.js
```diff
@@ -0,0 +1,17 @@
...
+  admin.listTopics((err, res) => {
+    if (res[1].metadata.architect) {
+      console.log('Kafka topic created');
+      clearInterval(interval_id);
+      child_process.execSync('npm start', { stdio: 'inherit' });
+    } else {
+      console.log('Waiting for Kafka topic to be created');
+    }
+  });
+}, 1000);
```
davidthor, line 17: Just fyi, this may have been more intuitive as a standalone script that exits zero or one for completion or failure. That would have allowed it to be generalized to something like this in the `CMD` field:

```sh
wait-for-kafka.sh && npm start
```

Hardcoding `npm start` into that script would make it harder to find where the actual start script is being called for editing and management. That being said, this is just an example so I'll approve since it all works locally now.

davidthor (approved):

PR #330, opened 2020-12-17
Added typescript interface representing the component spec

davidthor:
This includes everything needed to generate a json schema file from a typescript interface that I've included in the repo. 

Some things that could be done once this gets merged in:
* Publish a PR to the schemastore that uses the generated schema spec (must be done manually, and I will do once this gets merged in)
* Extend the interface that powers this schema gen as part of our config classes so that the interface becomes functionally useful
* Use something like AJV to validate incoming yaml files instead of having to use the class-validator annotations. AJV uses json schema and has sponsorship from Mozilla ensuring its stability: https://github.com/ajv-validator/ajv/

### DIFF
--- a/src/dependency-manager/README.md
+++ b/src/dependency-manager/README.md
@@ -1,4 +1,38 @@
-architect dependency graph
+Architect dependency-manager
 =============
 
-Library used to consolidate service dependencies into a consistent graph format.
+Library used for ingesting component specs and converting them to application graphs.
+
+## Generating and using json schema
+
+This project uses the [JSON schema](https://json-schema.org/) spec to help validate `architect.yml` files describing components. In order to make this schema easier to manage, we use the [typescript-json-schema]() library to generate the schema definition from a typescript interface.
+
+After you've made changes to the typescript files representing the schema(s), run the associated generate:schema command:
+
+```sh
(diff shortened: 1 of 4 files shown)


davidthor:
@tjhiggins this PR doesn't have any functional implications on the dependency manager, so I mostly wanted to make sure that the interface for the ComponentSpecV1 has all the fields we support.

tjhiggins:
Not a huge fan of having duplicate definitions. Wanted to point out that class-validator can also take a json-schema. 

I'm fine merging, but long term we should probably consolidate to only use json-schema and not class-validator decorators. 

Thoughts?

davidthor:
> Not a huge fan of having duplicate definitions. Wanted to point out that class-validator can also take a json-schema.
> https://github.com/typestack/class-validator/blob/99eec62eb9a80a5e8495fcd151ac636d5812ce90/README.md#defining-validation-schema-without-decorators
> 
> I'm fine merging, but long term we should probably consolidate to only use json-schema and not class-validator decorators.
> 
> Thoughts?

I def think it makes sense to consolidate into json schema. It to have significantly more support than class-validators unique approach via annotations. I just thought it'd be easier for someone else to take up that task once this base interface existed since it was somewhat tedious to write.

PR #332, opened 2020-12-22
Architect yml local deployments

ryan-cahill:


src/common/dependency-manager/local-manager.ts
```diff
@@ -27,12 +29,19 @@ export default class LocalDependencyManager extends DependencyManager {
...
-  static async createFromPath(api: AxiosInstance, env_config_path: string, linked_components: Dictionary<string> = {}): Promise<LocalDependencyManager> {
-    const dependency_manager = new LocalDependencyManager(api, env_config_path, linked_components);
-    const env_config = dependency_manager.config_path
-      ? await EnvironmentConfigBuilder.buildFromPath(dependency_manager.config_path)
-      : EnvironmentConfigBuilder.buildFromJSON({});
-
+  static async createFromPath(api: AxiosInstance, component_config_path: string, linked_components: Dictionary<string> = {}): Promise<LocalDependencyManager> {
+    let env_config;
+    const dependency_manager = new LocalDependencyManager(api, component_config_path, linked_components);
+    if (component_config_path.endsWith('environment.yml') || component_config_path.endsWith('environment.json') || !component_config_path) { // TODO: remove when environment configs are offically gone
```
ryan-cahill, line 35: @tjhiggins how broad do you want to make this condition? I ask because I think some examples that we've created in the past for customers have env configs named things like "env-mock-dev.yml", "dev-environment.yml", etc. There's also a file in our own examples for GCP microservices called "environment.local.yml

PR #333, opened 2020-12-28
Local deploys with values files

ryan-cahill:


src/dependency-manager/src/manager.ts
```diff
@@ -447,10 +450,18 @@ export default abstract class DependencyManager {
...
   }
 
-  async interpolateComponents(graph: DependencyGraph, interpolated_environment: EnvironmentConfig, component_map: Dictionary<ComponentConfig>) {
+  async interpolateComponents(graph: DependencyGraph, interpolated_environment: EnvironmentConfig, component_map: Dictionary<ComponentConfig>, values_dictionary: { [s: string]: { [s: string]: string } }) {
     const environment_components = interpolated_environment.getComponents();
     // Prefix interpolation expressions with components.<name>.
     const prefixed_component_map: Dictionary<ComponentConfig> = {};
+
+    // pre-sort values dictionary to properly stack/override any colliding keys
+    const sorted_values_keys = Object.keys(this.values_dictionary).sort();
```
tjhiggins, line 459: I don't understand this bit. The order of the top level keys in the values.yml matters. They should get applied top down.

Can you add a test to ensure we are respecting order?

Something like:
```yaml
'examples/react-app:*':
  log_level: warn
'*':
  log_level: override_all
```

src/dependency-manager/src/manager.ts
```diff
@@ -447,10 +450,18 @@ export default abstract class DependencyManager {
     return context;
   }
 
-  async interpolateComponents(graph: DependencyGraph, interpolated_environment: EnvironmentConfig, component_map: Dictionary<ComponentConfig>) {
+  async interpolateComponents(graph: DependencyGraph, interpolated_environment: EnvironmentConfig, component_map: Dictionary<ComponentConfig>, values_dictionary: { [s: string]: { [s: string]: string } }) {
```
tjhiggins, line 453: remove the arg - ref this.values_dictionary directly

src/common/dependency-manager/local-manager.ts
```diff
@@ -25,11 +25,11 @@ export default class LocalDependencyManager extends DependencyManager {
     this.linked_components = linked_components;
   }
 
-  static async create(api: AxiosInstance) {
-    return this.createFromPath(api, '');
+  static async create(api: AxiosInstance, values_dictionary: { [s: string]: { [s: string]: string } } = {}) {
```
tjhiggins, line 28: `values_dictionary: Dictionary<Dictionary<string>>`
Update all refs

tjhiggins (changes requested):

src/dependency-manager/src/spec/common/volume-v1.ts
```diff
@@ -11,10 +11,6 @@ export class VolumeSpecV1 extends ValidatableConfig {
     groups: ['debug'],
     message: 'Debug volumes require a host path to mount the volume to',
   })
-  @IsEmpty({
```
tjhiggins, line 14: I don't think we can remove this check. The host_path won't exist for remote deploys. There is an open issue for next milestone to address.
https://gitlab.com/architect-io/architect-cli/-/issues/116#note_473026607

test/commands/deploy.test.ts
```diff
@@ -381,6 +530,89 @@ describe('local deploy environment', function () {
...
+    .nock(MOCK_API_HOST, api => api
+      .get(`/accounts/examples/components/react-app/versions/latest`)
+      .reply(200, local_component_config_dependency))
+    .stub(Deploy.prototype, 'runCompose', sinon.stub().returns(undefined))
+    .stdout({ print })
+    .stderr({ print })
+    .command(['deploy', '-l', './examples/hello-world/architect.yml', '-i', 'test:hello', '-v', './examples/hello-world/values.yml'])
+    .it('Create a local deploy with a basic component, a dependency, and a values file', ctx => {
+      const runCompose = Deploy.prototype.runCompose as sinon.SinonStub;
+      expect(runCompose.firstCall.args[0].services['examples--hello-world--api--latest--d00ztoyu'].environment.a_required_key).to.equal('some_value');
```
tjhiggins, line 611: Try not to hardcode the generated refs... If they were to change it'll be a pain to update all of the tests.

src/dependency-manager/src/manager.ts
```diff
@@ -30,9 +32,10 @@ export default abstract class DependencyManager {
     this.__graph_cache = {};
   }
 
-  async init(environment_config?: EnvironmentConfig): Promise<void> {
+  async init(environment_config?: EnvironmentConfig, values_dictionary: Dictionary<Dictionary<string>> = {}): Promise<void> {
```
tjhiggins, line 35: I've been wondering is it safe to assign a mutable object in the constructor? Not that we modify values_dictionary in the class object, but I'm worried this is an anti-pattern if javascript behaves the same way as python.

src/dependency-manager/src/manager.ts
```diff
@@ -451,8 +454,17 @@ export default abstract class DependencyManager {
     const environment_components = interpolated_environment.getComponents();
     // Prefix interpolation expressions with components.<name>.
     const prefixed_component_map: Dictionary<ComponentConfig> = {};
+
+    // pre-sort values dictionary to properly stack/override any colliding keys
+    const sorted_values_keys = Object.keys(this.values_dictionary).sort();
```
tjhiggins, line 459: Can you add a new values.ts file to the dependency manager that has a validate function? Make the method take the values json so we can use it api side as well. Then call the validate method in the dep manager. 

Things to validate:
1. Structure - double nested
2. Keys only have a wildcard at the end of the string
3. ?

tjhiggins (changes requested):

src/dependency-manager/src/spec/common/volume-v1.ts
ryan-cahill: Sure, I've added redundant debug volumes to our configs in the meantime

src/dependency-manager/src/manager.ts
ryan-cahill: Did you have any alternatives in mind? We've made it like a year without messing up the environment config that's stored in the same way, so I wasn't too worried about it

tjhiggins: I meant the constructor arg. I don't think you're supposed to set mutable defaults. Def an anti pattern in python.

Arg = values_dictionary?
this.values_dictionary = values_dictionary || {}

tjhiggins: Ignore me. Different than python.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Evaluated_at_call_time

src/dependency-manager/src/manager.ts
ryan-cahill: I added that class and added tests, I'm not too sure where we'd like this to be called from though. I'm not sure that calling it in `init` feels right, but it's also the first place that the file hits the manager, so that does seem right. I'm a little torn. Also the fact that we need the original file contents in order to produce line numbers for validation errors makes me wonder if we should run the validation in the local manager instead. Seems strange if we were to pass file contents to the base manager

src/commands/deploy.ts
```diff
@@ -237,9 +241,32 @@ export default class Deploy extends DeployCommand {
     await execa('docker-compose', compose_args, { stdio: 'inherit' });
   }
 
+  private readValuesFile(values_file_path: string | undefined) {
+    let component_values = {};
+    if (values_file_path && fs.statSync(values_file_path)) {
+      const values_file_data = fs.readFileSync(values_file_path);
+      try {
+        component_values = JSON.parse(values_file_data.toString('utf-8'));
```
tjhiggins, line 249: remove support for JSON - create a generic loadYaml method?

tjhiggins (changes requested):

src/commands/deploy.ts
tjhiggins: feels like we have a couple custom loadYaml methods throughout the code base for no reason

ryan-cahill: Should that also check that the file exists? Or do we want to limit that function just to reading and returning contents?

tjhiggins: It shouldn't check if the file exists.
- If the user didn't pass a values param don't try to read
- if the user did pass the values param assume the file exists and fail if it doesn't

PR #337, opened 2021-01-11
adding homepage component config property and updating example component configs

ryan-cahill:


src/dependency-manager/src/spec/component/component-v1.ts
```diff
@@ -67,6 +67,10 @@ export class ComponentConfigV1 extends ComponentConfig {
   @IsString({ always: true })
   author?: string;
 
+  @IsOptional({ always: true })
+  @IsString({ always: true })
```
tjhiggins, line 71: Should we make it IsURL?

tjhiggins (changes requested):

tjhiggins:
Also reminder to update microservices submodule now that it's merged. Looking forward to this.

PR #338, opened 2021-01-11
fixing fusionauth example account

ryan-cahill:


tjhiggins:
Need to update test mock

PR #340, opened 2021-01-19
Tutorials/rabbit mq 2

dan-barrett:


dan-barrett:
@tjhiggins - there are two branches here that I'd like to keep around. There are a couple links in the tutorial that point to the source code from these branches (with earlier versions of the file before we Dockerize and parameterize them).

The branches start with `tutorials/*` to hopefully distinguish them.

tjhiggins (approved):

PR #349, opened 2021-02-10
Updated @Transform() method to match new contract

davidthor:


tjhiggins:
Looks like some tests are failing? Really confused how none of our tests caught this issue. Are we sure we shouldn't just revert?

ryan-cahill:
@tjhiggins https://github.com/architect-team/architect-cli/pull/350

PR #352, opened 2021-02-15
adding external interface to base manager

ryan-cahill:


src/dependency-manager/src/manager.ts
```diff
@@ -580,5 +580,51 @@ export default abstract class DependencyManager {
...
+      for (const component_interface_name of Object.keys(component_interfaces)) {
+        const tagless_component_name = component_config.getName();
+
+        let to_node = graph.getNodeByRef(`${component_name}-interfaces`);
+        let current_interface_name = component_interface_name;
+        let dependent_nodes = graph.getDependentNodes(to_node);
+        while (dependent_nodes.length) {
+          const dependent_node = dependent_nodes.pop();
+          const edge = graph.edges.find(edge => edge.from === dependent_node.ref && edge.to === to_node.ref);
+          if (edge instanceof IngressEdge && all_external_interfaces[current_interface_name]) {
```
tjhiggins, line 606: Can we use node.is_external instead of refing the components_map? 

If not it sounds more like a problem in the graph generation if we don't have the information required.

src/dependency-manager/src/manager.ts
```diff
@@ -580,5 +580,51 @@ export default abstract class DependencyManager {
...
+            }
+          }
+        }
+      }
+    }
+
+    return ingresses;
+  }
+
+  abstract getArchitectContext(graph: DependencyGraph, components_map: Dictionary<ComponentConfig>): Promise<any | undefined>;
```
tjhiggins, line 629: You could probably make this not abstract and then you don't need to define in the local.

tjhiggins (changes requested):

src/dependency-manager/src/manager.ts
ryan-cahill: Sorry, this was just bad naming

davidthor:
@ryan-cahill is there anything else outstanding on this one?

ryan-cahill:
Not that I can think of. This is ready for another look @tjhiggins

PR #365, opened 2021-03-30
use triggering branch for CLI release

dan-barrett:


.github/workflows/publish.yml
```diff
@@ -19,8 +19,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-        with:
```
tjhiggins, line 22: Do we also need to change these lines? I see a ref to master in the publish steps. 

https://github.com/architect-team/architect-cli/blob/master/.github/workflows/cli-rc.yml#L67
https://github.com/architect-team/architect-cli/blob/master/.github/workflows/dependency-manager-rc.yml#L42

dan-barrett: Those are only for publishing RCs. Do we want to publish RCs for hotfixes? Maybe.

PR #373, opened 2021-04-01
install applications on platform create

dan-barrett:
Closes: https://gitlab.com/architect-io/architect-cli/-/issues/216

### DIFF
--- a/src/commands/platforms/create.ts
+++ b/src/commands/platforms/create.ts
@@ -11,7 +11,7 @@ import { CreatePlatformInput } from '../../common/utils/platform';
 import { Slugs } from '../../dependency-manager/src';
 
 export default class PlatformCreate extends Command {
-  static aliases = ['platform:create', 'platforms:create'];
+  static aliases = ['platforms:register', 'platform:create', 'platforms:create'];
   static description = 'Register a new platform with Architect Cloud';
 
   static args = [{
@@ -23,6 +23,7 @@ export default class PlatformCreate extends Command {
   static flags = {
     ...Command.flags,
     ...AccountUtils.flags,
+    auto_approve: flags.boolean(),
     type: flags.string({ char: 't', options: ['KUBERNETES', 'kubernetes', 'ECS', 'ecs'] }),
     host: flags.string({ char: 'h' }),
     kubeconfig: flags.string({ char: 'k', default: '~/.kube/config', exclusive: ['service_token', 'cluster_ca_cert', 'host'] }),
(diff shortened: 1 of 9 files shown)


src/commands/platforms/create.ts
```diff
@@ -69,6 +70,13 @@ export default class PlatformCreate extends Command {
     const created_platform = await this.post_platform_to_api(platform_dto, account.id);
     cli.action.stop();
 
+    this.log(`Hang tight! We're completing the setup of your platform by installing applications that will broker networking. This could take as long as 15m, so feel free to grab a cup of coffee while you wait.`);
+    this.log(`Logs here: ${this.app.config.app_host}/examples/platforms/new?platform_id=${created_platform.id}`);
+    cli.action.start(chalk.blue('Installing platform applications'));
```
tjhiggins, line 75: @davidthor I'd assume you want a confirmation here?

davidthor: Yes please. Just like in the UI.

dan-barrett: Should we support an `--auto-approve` flag here to skip the prompt?

tjhiggins: You can add, but its prob not going to be used often.

davidthor: Seconded. I don't see how/why platform registration would be a recurring, automated process. Environments will spin up and down, but the platform will only ever be registered once.

src/commands/platforms/destroy.ts
```diff
@@ -7,8 +7,8 @@ import { AccountUtils } from '../../common/utils/account';
 import { PlatformUtils } from '../../common/utils/platform';
 
 export default class PlatformDestroy extends Command {
-  static aliases = ['platform:destroy', 'platforms:destroy'];
-  static description = 'Destroy a platform';
+  static aliases = ['platforms:deregister', 'platform:destroy', 'platforms:destroy'];
+  static description = 'Deregister a platform from Architect';
```
dan-barrett, line 11: @davidthor , @tjhiggins : proposing these changes to align the language of the inverse command ("Register a new platform...")

Plus I think it's scary to say we're "destroying" a platform that they brought.

tjhiggins: I'm not opposed on the condition we are consistent and update to environment:deregister too. Wait for David to respond before updating.

davidthor: I'm cool with it since we aren't actually changing anything in the environment when we remove it.

dan-barrett: Do we no longer remove all resources in the environment when we do an environment:destroy?

I liked environment:destroy because an Environment is an Architect construct with architect-created resources, whereas the platform felt like something that they were bringing to the table and we weren't destroying it, just deregistering it.

tjhiggins: Ah right - that is a confusing disconnect from the UI where removing "deregisters".

tjhiggins: Ha wait - `architect destroy` handles the destruction of actual cloud resources. `architect environment:destroy` just deregisters the same as platform.

PR #374, opened 2021-04-01
Updating curl command URL for task

ryan-cahill:
Wait for next production push to merge

### DIFF
--- a/examples/scheduled-tasks/architect.yml
+++ b/examples/scheduled-tasks/architect.yml
@@ -11,10 +11,7 @@ tasks:
     command:
       - sh
       - -c
-      - curl $SERVER_URL
-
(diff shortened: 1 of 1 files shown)


examples/scheduled-tasks/architect.yml
```diff
@@ -11,10 +11,7 @@ tasks:
     command:
       - sh
       - -c
```
tjhiggins, line 13: Why the ingress url? Is this because the consul sidecar for the task wasn't working as expected?

ryan-cahill: Well there currently isn't a sidecar for k8s tasks, so the external endpoint will have to do until that work is completed

PR #377, opened 2021-04-13
Fixing task lookup for local task execs

ryan-cahill:


src/commands/task.ts
```diff
@@ -82,7 +82,8 @@ export default class TaskExec extends Command {
     let service_name;
     const slug = ServiceVersionSlugUtils.build(parsed_slug.component_account_name, parsed_slug.component_name, args.task, parsed_slug.tag);
     const ref = Refs.safeRef(slug);
```
tjhiggins, line 84: Use `ComponentConfig.getNodeRef(slug)` that should match 1:1 with the compose.services keys

tjhiggins (changes requested):

src/commands/task.ts
```diff
@@ -81,9 +81,8 @@ export default class TaskExec extends Command {
 
     let service_name;
     const slug = ServiceVersionSlugUtils.build(parsed_slug.component_account_name, parsed_slug.component_name, args.task, parsed_slug.tag);
-    const ref = Refs.safeRef(slug);
-    const service_hash = ref.substring(ref.lastIndexOf('-') + 1);
-    const matching_names = Object.keys(compose.services).filter(name => name.endsWith(service_hash));
+    const ref = ComponentConfig.getNodeRef(slug)
+    const matching_names = Object.keys(compose.services).filter(name => name.includes(ref));
```
tjhiggins, line 85: Shouldn't it be `name === ref`? Also lint warning.

tjhiggins (changes requested):

PR #381, opened 2021-04-16
Sticky interfaces

ryan-cahill:


src/dependency-manager/src/spec/common/interface-v1.ts
```diff
@@ -27,6 +27,9 @@ export class InterfaceSpecV1 extends ValidatableConfig {
   @IsOptional({ always: true })
   url?: string;
 
+  @IsOptional({ always: true })
+  sticky?: string;
```
tjhiggins, line 31: add IsBooleanString?

src/dependency-manager/src/spec/common/interface-spec.ts
```diff
@@ -6,6 +6,7 @@ export interface InterfaceSpec {
   protocol?: string;
   username?: string;
   password?: string;
+  sticky?: string;
```
tjhiggins, line 9: Add the traefik labels to the docker-compose if sticky is set and add a test case to confirm it gets set.

tjhiggins (changes requested):

PR #384, opened 2021-04-27
add docs for depends_on; tweak instance_id for components

dan-barrett:
@tjhiggins note: I tweaked the instance_id. I didn't realize this gets used for the component graph too, not just on local deployments. I don't think it matters but I figured `*-component` makes more sense than `*-local`

### DIFF
--- a/src/dependency-manager/src/manager.ts
+++ b/src/dependency-manager/src/manager.ts
@@ -12,7 +12,6 @@ import { ComponentConfig } from './spec/component/component-config';
 import { Dictionary } from './utils/dictionary';
 import { flattenValidationErrors, ValidationErrors } from './utils/errors';
 import { interpolateString, replaceBrackets } from './utils/interpolation';
-import { Refs } from './utils/refs';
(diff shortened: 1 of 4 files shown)


src/dependency-manager/src/manager.ts
```diff
@@ -29,7 +29,7 @@ export default abstract class DependencyManager {
         local_path: component.getLocalPath(),
         artifact_image: component.getArtifactImage(),
       });
-      node.instance_id = `${Refs.safeRef(component.getRef())}-local`;
+      node.instance_id = `${Refs.safeRef(component.getRef())}-component`;
```
tjhiggins, line 32: Missed this in the initial review. Remove setting node.instance_id here. Set in the local-manager getGraph method when we loop over the nodes for a component. 

https://github.com/architect-team/architect-cli/blob/master/src/common/dependency-manager/local-manager.ts#L45

tjhiggins (changes requested):

dan-barrett:
@tjhiggins updated

dan-barrett:
@tjhiggins just a reminder on this

PR #386, opened 2021-04-30
adding sticky sessions for interfaces to the docs

ryan-cahill:


docs/1-configuration/1-services.md
```diff
@@ -83,6 +83,9 @@ interfaces:
 
     # (optional) A basic auth password required to access the interface
     password: ${{ parameters.API_PASSWORD }}
+
+    # (optional) Requests made to this interface, if made external, will use sticky sessions
+    sticky: true
```
tjhiggins, line 88: Defaults to false

tjhiggins: mention that it defaults to false*

PR #392, opened 2021-06-02
Architect auth integration

ryan-cahill:


src/app-config/auth.ts
```diff
@@ -155,7 +222,11 @@ export default class AuthClient {
       return; // Don't refresh if a token doesn't exist
     }
 
-    this.auth_results = await this.performOauthRefresh(token.refresh_token) as AuthResults;
+    if (this.config.oauth_domain === 'auth.architect.io') {
```
tjhiggins, line 225: Was really hoping to avoid this level of segmentation between the two flows. 

My impression was creating the AuthorizationCode client would be the split and could be passed to all of the functions.

```ts
let auth_config;
if (auth0) {
  auth_config = ...
} else {
  auth_config = ...
}
const auth_client = new AuthorizationCode(auth_config);
loginWithBrowser(auth_client);
```

I would have assumed something similar for the refresh token flow. Based on:
https://www.npmjs.com/package/simple-oauth2#access-token

tjhiggins (changes requested):

src/app-config/auth.ts
ryan-cahill: I'll have a look into it and see what I can find out

PR #396, opened 2021-06-29
Add support for passing dynamic build args

tjhiggins:


src/commands/register.ts
```diff
@@ -26,10 +27,14 @@ export default class ComponentRegister extends Command {
 
   static flags = {
     ...Command.flags,
+    arg: flags.string({
+      char: 'a',
```
ryan-cahill, line 31: They're very different commands but I feel like it might be confusing that we also use `-a` for the account flag in so many other places

src/commands/register.ts
```diff
@@ -147,7 +154,15 @@ export default class ComponentRegister extends Command {
...
-        build_args = Object.entries(service_config.build?.args).map(([key, value]) => `${key}=${value}`);
+        const build_args_map: Dictionary<string> = service_config.build?.args || {};
+        for (const arg of flags.arg || []) {
+          const [key, value] = arg.split('=');
+          if (!value) {
+            throw new Error('--arg must be in the format key=value');
+          }
+          build_args_map[key] = value;
+        }
+        build_args = Object.entries(build_args_map).map(([key, value]) => `${key}=${value}`);
```
ryan-cahill, line 165: Can't we just set `build_args` straight to `service_config.build?.args || []` after passing validation? `build_args_map` doesn't seem necessary, though the validation does of course

src/commands/register.ts
```diff
@@ -147,7 +154,15 @@ export default class ComponentRegister extends Command {
       }
       let build_args: string[] = [];
       if (service_config.build?.args) {
-        build_args = Object.entries(service_config.build?.args).map(([key, value]) => `${key}=${value}`);
+        const build_args_map: Dictionary<string> = service_config.build?.args || {};
+        for (const arg of flags.arg || []) {
+          const [key, value] = arg.split('=');
+          if (!value) {
+            throw new Error('--arg must be in the format key=value');
```
ryan-cahill, line 161: It could help if this was specific about which key/value pair failed

src/commands/register.ts
tjhiggins: true - I actually meant to not include the shorthand. good call

src/commands/register.ts
tjhiggins: sure

PR #401, opened 2021-07-19
Unsupported node version message

ryan-cahill:


bin/check-version.js
```diff
@@ -0,0 +1,7 @@
+const semver = require('semver');
```
tjhiggins, line 1: We should probably add semver as a dependency

PR #404, opened 2021-07-20
Error for unsupport node versions

ryan-cahill:


package.json
```diff
@@ -52,6 +52,7 @@
     "@types/mustache": "^4.1.1",
     "@types/node": "14.14.31",
     "@types/opener": "^1.4.0",
+    "@types/semver": "^7.3.7",
```
tjhiggins, line 55: add semvar

src/hooks/init/check-version.ts
```diff
@@ -0,0 +1,12 @@
+import { Hook } from '@oclif/config';
+import semver from 'semver';
+
+const hook: Hook<'init'> = async function (options) {
+  const version = require('../../../package').engines.node;
+  if (!semver.satisfies(process.version, version)) {
+    console.log(`The required node version ${version} for the Architect CLI is not satisfied by your current node version ${process.version}.`);
+    process.exit(1);
```
tjhiggins, line 8: You should be able to use `this.error` or throw an error

PR #406, opened 2021-07-21
adding service scaling to spec

ryan-cahill:


src/dependency-manager/src/spec/common/scaling-v1.ts
```diff
@@ -0,0 +1,27 @@
...
+export class ScalingSpecV1 extends ValidatableConfig {
+  @IsOptional({ always: true })
+  @Type(() => String)
+  min_replicas?: string;
+
+  @IsOptional({ always: true })
+  @Type(() => String)
+  max_replicas?: string;
+
+  @IsOptional({ always: true })
```
tjhiggins, line 24: Are any of these optional assuming they added a scaling block? I think we should require. I could see cpu or memory being optional, but at least one has to be set. Unsure if we should have defaults for those.

tjhiggins (changes requested):

PR #410, opened 2021-08-03
adding environment ttl flag

ryan-cahill:


src/commands/environments/create.ts
```diff
@@ -57,6 +60,7 @@ export default class EnvironmentCreate extends Command {
       name: environment_name,
       description: flags.description,
       platform_id: platform.id,
+      ttl: flags.ttl,
```
tjhiggins, line 63: To prevent breaking changes with on-prem:

```
if flags.ttl:
  dto.ttl = flags.ttl
```

src/commands/environments/create.ts
```diff
@@ -18,6 +18,9 @@ export default class EnvironmentCreate extends Command {
     description: flags.string({
       description: 'Environment Description',
     }),
+    ttl: flags.string({
```
tjhiggins, line 21: Lets validate ttl

Add https://www.npmjs.com/package/ms

Then make sure `ms(flags.ttl) > 0`

tjhiggins (changes requested):

src/commands/environments/create.ts
tjhiggins: Edit: Lets actually validate api side

PR #411, opened 2021-08-04
Register diff

ryan-cahill:


package.json
```diff
@@ -19,6 +19,8 @@
     "class-validator": "^0.13.1",
     "cli-table3": "0.6.0",
     "cli-ux": "5.4.9",
+    "colors": "^1.4.0",
```
tjhiggins, line 22: We already use [chalk](https://www.npmjs.com/package/chalk) for coloring logs. Search `chalk.green`. Surprised to see its not in our package.json dependencies. We should probably add.

src/commands/register.ts
```diff
@@ -119,10 +121,36 @@ export default class ComponentRegister extends Command {
...
+      previous_config_data = await this.getComponentConfig(selected_account.name, component_name, tag);
+    /* eslint-disable-next-line no-empty */
+    } catch {}
+
     cli.action.start(chalk.blue(`Registering component ${raw_config.name}:${tag} with Architect Cloud...`));
     await this.app.api.post(`/accounts/${selected_account.id}/components`, component_dto);
     cli.action.stop();
     this.log(chalk.green(`Successfully registered component`));
+
+    this.log(chalk.blue(`Begin component config diff`));
```
tjhiggins, line 136: Can you put the diff before:
```ts
cli.action.start(chalk.blue(`Registering component ${raw_config.name}:${tag} with Architect Cloud...`));
```

tjhiggins (changes requested):

src/commands/register.ts
ryan-cahill: Didn't you want the config to compare to be pulled from the API? If we're going to pull the new config to use after it's pushed, we have to register first

tjhiggins: Why do we need to pull the new config? We post the new config to register the component so we should already have it?

ryan-cahill: Never mind, I see what you mean

PR #412, opened 2021-08-05
262 flags use dashes

brahmlower:


src/base-command.ts
```diff
@@ -16,6 +20,15 @@ export default abstract class extends Command {
     help: flags.help({ char: 'h' }),
   };
 
+  checkFlagDeprecations(flags: any, flagDefinitions: any) {
+    Object.keys(flags).forEach((flagName: string) => {
+      const description: string = flagDefinitions[flagName].description
+      if (description?.startsWith(DEPRECATED_LABEL)) {
```
tjhiggins, line 26: Fix lint warnings

src/base-command.ts
```diff
@@ -16,6 +20,15 @@ export default abstract class extends Command {
     help: flags.help({ char: 'h' }),
   };
 
+  checkFlagDeprecations(flags: any, flagDefinitions: any) {
```
tjhiggins, line 23: Something I didn't cover yet, but our coding style is:

snakecase for vars
camelcase for functions

src/base-command.ts
```diff
@@ -16,6 +20,15 @@ export default abstract class extends Command {
...
+    Object.keys(flags).forEach((flagName: string) => {
+      const description: string = flagDefinitions[flagName].description
+      if (description?.startsWith(DEPRECATED_LABEL)) {
+        this.warn(`Flag --${flagName} is deprecated.${description.split(DEPRECATED_LABEL)[1]}`);
+      }
+    });
+  }
+
```
tjhiggins, line 31: I see my pseudo code didn't work first time. Use the following so you don't have to call checkFlagDeprecations in every command.

```ts
async init() {
    const { flags } = this.parse(this.constructor as any);
    const flag_definitions = (this.constructor as any).flags;
    this.checkFlagDeprecations(flags, flag_definitions);
    ...
```

src/commands/platforms/create.ts
```diff
@@ -24,24 +24,79 @@ export default class PlatformCreate extends Command {
...
+    }),
+    aws_region: flags.string({
+      exclusive: ['awsconfig', 'kubeconfig', 'service-token', 'cluster-ca-cert', 'host'],
+      description: `${Command.DEPRECATED} Please use --aws-region.`,
+      hidden: true,
+    }),
+    ['aws-region']: flags.string({
+      exclusive: ['awsconfig', 'kubeconfig', 'service-token', 'cluster-ca-cert', 'host']
+    }),
+    service_token: flags.string({
```
tjhiggins, line 63: Remove unused flags service_token/cluster_ca_cert and their new dashed versions

tjhiggins (changes requested):

src/base-command.ts
tjhiggins: I spoke too soon - appreciate the self review :)

src/base-command.ts
brahmlower: doah, that helps a bunch, thanks! Will get that switched over

src/base-command.ts
brahmlower: oh thanks- I forgot to ask about this earlier. I did another quick review and I think this function had the only styling violations. Should be fixed in the latest commit.

src/commands/platforms/create.ts
brahmlower: If it's okay with you, I'd really like to remove these in a separate commit, that way the resulting merge commits are relatively atomic, even though it kinda sucks to maintain the dead weight of these couple extra flags through this PR 😬

tjhiggins: Fine by me as long as we don't forget

brahmlower: Gotcha- I recognize I still haven't made a ticket for taking care of that. Meant to make it after wrapping up this PR, but this has taken a fair bit longer than expected 😅

PR #415, opened 2021-08-09
Added --force flag to platform:destroy command

brahmlower:


src/commands/platforms/destroy.ts
```diff
@@ -63,7 +68,11 @@ export default class PlatformDestroy extends Command {
     const { data: account_platform } = await this.app.api.get(`/accounts/${account.id}/platforms/${platform.name}`);
 
     cli.action.start(chalk.blue('Deregistering platform'));
-    await this.app.api.delete(`/platforms/${account_platform.id}`);
+    await this.app.api.delete(`/platforms/${account_platform.id}`, {
+      params: {
+        force: answers.force ? 1 : 0,
```
tjhiggins, line 73: Only set params if force is set. Prevents breaking changes for on-prem.

```ts
const params = {};
if (answers.force) {
  params.force = 1;
}
```

tjhiggins (changes requested):

src/commands/platforms/destroy.ts
brahmlower: Yikes, good catch! Will do 👍

brahmlower:
There are a couple additional changes I'd like to make to the API before we make this functionality available via the CLI. So I'm going to put this PR on hold for the moment.

PR #422, opened 2021-08-18
281: Support filtering env & platforms by account name

brahmlower:


brahmlower:
Need to update docs for these changes

src/commands/environments/index.ts
```diff
@@ -16,9 +18,25 @@ export default class Environments extends Command {
...
+      q: args.query || '',
+    };
+    if (account) {
+      query_params['account_id'] = account.id;
+    }
+
+    // Reduces an object into a string of query parameters
+    const query_string = Object.entries(query_params).map(key_val => key_val.join('=')).join('&');
+
+    const { data: { rows: environments } } = await this.app.api.get(`/environments?${query_string}`);
```
tjhiggins, line 39: axios has an option for query params.

```ts
const params = {
  q: args.query || '',
  account_id: account?.id
}
const { data: { rows: environments } } = await this.app.api.get(`/environments`, { params });
```

tjhiggins (changes requested):

src/commands/environments/index.ts
brahmlower: 🤦 oof, I didn't realize that. Will make that change 👍

PR #423, opened 2021-08-19
[WIP] Deploy command feedback

ryan-cahill:


package.json
```diff
@@ -27,6 +27,7 @@
     "is-ci": "^3.0.0",
     "is-windows": "^1.0.2",
     "js-yaml": "4.1.0",
+    "listr": "^0.14.3",
```
tjhiggins, line 30: I don't remember why we stopped using listr. Doesn't look like its maintained anymore for one. 

@dan-barrett can you remember why we switched to cli-ux?

dan-barrett: Nope, I don't remember. I don't think I added that. But my guess would be that we did it to kee using [oclif](https://github.com/oclif/cli-ux) wherever we can.

ryan-cahill: I see that it hasn't been worked on for a while, but I used if because oclif still recommends it at least for spinners. Assumed the whole thing was fair game

https://oclif.io/docs/spinner

src/commands/deploy.ts
```diff
@@ -441,14 +440,74 @@ export default class Deploy extends DeployCommand {
...
+        title: `Service ${service_name}`,
+        task: () => new Promise((resolve, reject) => {
+          let service_health_poll_count = 0;
+          const service_health_poll = setInterval(async () => {
+            if (service_health_poll_count > 180) {
+              clearInterval(service_health_poll);
+              return reject(new Error(`Service ${service_name} deployment timed out`));
+            }
+
+            const environment_health: EnvironmentHealth = (await this.app.api.get(`/environments/${environment.id}/health`)).data;
```
dan-barrett, line 476: I'm wondering if we can do this loop on the server so we only have to poll a single endpoint instead of polling for each service. The endpoint could return the health of each service environment?

tjhiggins: > instead of polling for each service
The endpoint used is to get the health of the environment which includes all services. 

> service environment
service in the environment?

package.json
tjhiggins: Looks like listr is pretty dead - I noticed there is a listr2 with some traction:
https://www.npmjs.com/package/listr2

PR #428, opened 2021-08-20
check if JSON.parse returns object in interoplation

dan-barrett:


dan-barrett:
@tjhiggins FYI- I'm merging this into master, tests passing and tested the `register` again.

Checking for a `\n` in the interpolation was insufficient for cases where there was a JSON string on one line. I decided to try a JSON.parse() and if it resulted in an object, then we do a JSON.stringify() instead.

PR #432, opened 2021-08-22
Added component-schema.json to version control

davidthor:
...so it can be referenced by Schema Store: https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md

### DIFF
--- a/src/dependency-manager/component-schema.json
+++ b/src/dependency-manager/component-schema.json
@@ -1,0 +1,1823 @@
+{
+  "$id": "https://architect.io/.schemas/component.schema.json",
+  "title": "JSON Schema for Architect.io configuration",
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "properties": {
+    "name": {
+      "pattern": "^(?=.{1,32}(\\/|:|$))[a-z0-9]+(?:-[a-z0-9]+)*\\/(?=.{1,32}(\\/|:|$))[a-z0-9]+(?:-[a-z0-9]+)*(?:@[\\w][\\w\\.-]{0,127})?$",
+      "type": "string"
+    },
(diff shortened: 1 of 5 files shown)


dan-barrett:
@davidthor I'm going to close this because it has a bunch of conflicts with my current working branch, I already had the SchemaStore work underway. But good call on moving `ts-node` to dev dependencies, i'll add that change to my branch.

Also, I saw this change: `$id: 'https://architect.io/.schemas/component.schema.json', did you plan on hosting the schema from our marketing website? In my branch I'm just using the raw https://raw.githubusercontent.com link to our repo.

davidthor:
> @davidthor I'm going to close this because it has a bunch of conflicts with my current working branch, I already had the SchemaStore work underway. But good call on moving `ts-node` to dev dependencies, i'll add that change to my branch.
> 
> Also, I saw this change: `$id: 'https://architect.io/.schemas/component.schema.json',` did you plan on hosting the schema from our marketing website? In my branch I'm just using the raw https://raw.githubusercontent.com link to our repo.

The `$id` field in the spec is supposed to just be a unique identifier, but I think I may have misunderstood and it may need to be a resolvable URL for other schemas to be able to reference it. In either case it's no big deal because we don't have any current usage for referencing and extending the component spec.

dan-barrett:
Yeah no I also the `$id` too, I'm just pointing it here: `https://raw.githubusercontent.com/architect-team/architect-cli/master/src/dependency-manager/schema/architect.schema.json` instead. (ie the master copy of our schema will live in the raw source directory on github, rather than on the marketing site.)

If you'd prefer to host it on https://architect.io/.schemas/component.schema.json let me know, if not we're all good here.

PR #433, opened 2021-08-24
286 schemastore

dan-barrett:


dan-barrett:
@tjhiggins this is good for merge. This includes a bunch of refactoring I did this weekend, adds the jsonschema version file, and does docs generation.

One open question: How should we trigger schema generation? Currently it's in an npm script: `npm run generate:schema`. 

Both outputs of this script needs to live in source control:
* JSONSchema will live at https://raw.githubusercontent.com/architect-team/architect-cli/master/src/dependency-manager/schema/architect.schema.json to be picked up by SchemaStore
* generated Spec Docs also need to live in source control to be picked up by Gatsby for the marketing website.

Options:
* we add some pre-commit hook to run `npm run generate:schema` on the developer's machine
* we rely on the developer to run it if they make changes to the schema spec
* we add a Github action that runs and then subsequently re-commits

dan-barrett:
@tjhiggins this has conflicts now as of your most recent commit. Hold off on merging this and I'll resolve the conflicts. But then we should get this one in next because it touches most of the new schema files.

dan-barrett:
@tjhiggins this is good to merge again

PR #434, opened 2021-08-24
Support object interpolation

tjhiggins:
Make consumers an array of strings
Add test of parameter with object default

### DIFF
--- a/src/dependency-manager/src/schema/spec/component-spec.ts
+++ b/src/dependency-manager/src/schema/spec/component-spec.ts
@@ -64,8 +64,8 @@ export class ParameterDefinitionSpec {
   description?: string;
 
   @IsOptional()
-  @JSONSchema(AnyOf('boolean', 'number', 'object', 'string', 'null'))
-  default?: boolean | number | string | null;
+  @JSONSchema(AnyOf('array', 'boolean', 'number', 'object', 'string', 'null'))
+  default?: boolean | number | object | string | null;
 }
 
 export class ComponentSpec {
--- a/src/dependency-manager/src/schema/spec/component-spec.ts
+++ b/src/dependency-manager/src/schema/spec/component-spec.ts
@@ -64,7 +64,7 @@ export class ParameterDefinitionSpec {
   description?: string;
 
   @IsOptional()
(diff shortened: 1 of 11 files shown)


src/dependency-manager/src/schema/spec/component-spec.ts
```diff
@@ -64,7 +64,7 @@ export class ParameterDefinitionSpec {
   description?: string;
 
   @IsOptional()
-  @JSONSchema(AnyOf('boolean', 'number', 'string', 'null'))
+  @JSONSchema(AnyOf('boolean', 'number', 'object', 'string', 'null'))
   default?: boolean | number | string | null;
```
dan-barrett, line 68: We should add `object` to the `default` typescript type here too.

src/dependency-manager/src/schema/spec/resource-spec.ts
```diff
@@ -47,8 +47,8 @@ export class VolumeSpec {
   readonly?: boolean | string;
 }
 
-export type EnvironmentSpecValue = boolean | null | number | string;
-export const EnvironmentDictiory = DictionaryOfAny('boolean', 'null', 'number', 'string');
+export type EnvironmentSpecValue = boolean | null | number | object | string;
```
dan-barrett, line 50: Likewise, we should add `array` to the `EnvironmentSpecValue ` typescript type too.

tjhiggins: arrays are objects

src/dependency-manager/src/schema/spec/transform/resource-transform.ts
```diff
@@ -48,7 +48,12 @@ export const transformResourceSpecEnvironment = (environment: Dictionary<Environ
     if (v === undefined || v === null) {
       continue;
     }
-    output[k] = `${v}`;
+
+    if (v instanceof Object) {
```
dan-barrett, line 52: Also I don't think this is sufficient to check if it's an array. And I'm assuming we'll also want to stringify arrays, correct?

tjhiggins: it is

PR #451, opened 2021-09-07
Task schedule is validated as a cron pattern

brahmlower:


dan-barrett:
@brahmlower looks good, but I just noticed one more thing that would be good to test.

In `examples-validation.test.ts`, we run all of the `examples/*.architect.yml` through validation which is a really a nice gut check. We've always used `examples/scheduled-tasks` as our `tasks` block sample, but it looks like the `schedule` field got removed somewhere along the way. We should add that back in to check our new regex. Adding the `schedule` field back into that `architect.yml` just as it used to be should pass continue to pass `examples-validation.test.ts`.

dan-barrett:
@brahmlower thanks! Reviewed and merged.

PR #454, opened 2021-09-09
remove unnecessary console.clear in deploy

brahmlower:


src/commands/deploy.ts
```diff
@@ -194,8 +194,6 @@ export default class Deploy extends DeployCommand {
       await execa('docker-compose', ['-f', compose_file, '-p', project_name, 'build'], { stdio: 'inherit' });
     }
 
-    console.clear();
```
tjhiggins, line 197: My thinking was if the build succeeds then you don't want to see that noise and only the deployment logs.

brahmlower: I definitely agree about the noise from the build, but I noticed that the clear was actually clearing some output from previous commands I had that I wanted to later reference, but that was cleared away the `console.clear`. This video demonstrates the behavior I'm seeing

https://user-images.githubusercontent.com/3588764/132752609-249a6c33-ca79-44ac-90fe-5325f362f009.mov

tjhiggins: I guess mine does that too. Looks like I might have wanted to do:
`process.stdout.write('\x1Bc');` however I don't know if that works in all shells/os.

https://www.codegrepper.com/code-examples/javascript/clear+terminal+node+js

PR #456, opened 2021-09-13
Path based routing

ryan-cahill:


src/dependency-manager/src/graph/node/interfaces.ts
```diff
@@ -1,18 +1,27 @@
...
   ref!: string;
   slug!: string;
+  component_interface_configs!: Dictionary<ComponentInterfaceConfig>;
 
-  constructor(ref: string, slug: string) {
+  constructor(ref: string, slug: string, component_interface_configs: Dictionary<ComponentInterfaceConfig>) {
     super();
     this.ref = ref;
     this.slug = slug;
+    this.component_interface_configs = component_interface_configs;
```
tjhiggins, line 17: be consistent and call it `config`

src/dependency-manager/src/manager.ts
```diff
@@ -154,6 +155,7 @@ export default abstract class DependencyManager {
           subdomain = interpolateStringOrReject(subdomain, dep_component.context);
           // eslint-disable-next-line no-empty
         } catch { }
+        subdomain = `${subdomain}->${interface_name}`;
```
tjhiggins, line 158: Not a fan of this approach. We should keep subdomain unmodified because the subdomain is still `www` even if it has a path. Also this isn't guaranteed to be unique across components.

src/common/docker-compose/index.ts
```diff
@@ -253,19 +255,29 @@ export class DockerComposeUtils {
...
+          if (interface_subdomain === '@') {
             // @ is an invalid service name for traefik
-            interface_name = '__at__';
+            interface_subdomain = '__at__';
+          }
+
+          const interfaces_node = graph.getUpstreamNodes(node_to).find(n => n instanceof InterfacesNode);
+          const component_interface = (interfaces_node as InterfacesNode).component_interfaces[component_interface_name];
+          if (interfaces_node && component_interface?.ingress?.path) {
+            service_to.labels.push(`traefik.http.routers.${interface_subdomain}.rule=Host(\`${host}\`) && Path(\`${component_interface.ingress.path}.*\`)`);
```
tjhiggins, line 271: Worried this might be a bad assumption:
`${component_interface.ingress.path}.*`

This is essentially traefik's path prefix:
https://doc.traefik.io/traefik/routing/routers/#rule

Lets poll Brahm/Dan on default behavior.

Wondering if we should require that someone specify the `.*`

src/dependency-manager/src/manager.ts
tjhiggins: Is traefik the only place where we run into conflicts between subdomains?

tjhiggins (changes requested):

src/common/docker-compose/index.ts
ryan-cahill: I guess I'd prefer that as long as we make sure that people understand what that means under the hood

src/dependency-manager/src/manager.ts
ryan-cahill: The end of the flow where we create the traefik config is the only place I've seen this really turn into a problem, but we make assumptions in various places in the CLI and the manager that the subdomain is unique. The fact that it's used as a key turns into a problem even starting as early as [here](https://github.com/architect-team/architect-cli/pull/456/files#diff-38950282aef5cea0f96cba5414c96fc6597942377b028fddf996ad9b63649519R338) before we get into the manager

tjhiggins: I think its fine for the first pass to not allow conditional exposing of interface paths. If you do `-i www` it would just expose all subdomains with `www` even if they have different paths.

tjhiggins: If you make the traefik service name a ref of component/interface then it'd be unique. That's what we do on the api side. See `consulRef(service_node.ref, interface_name)`

ryan-cahill: I can do that for the traefik service name. What about things like [this](https://github.com/architect-team/architect-cli/pull/456/files#diff-61da828261689d3f251ab9c24859dadff3bb043a5b45e54093c6f944cc045943R172) though? What should happen in a case like this if two subdomains are the same?

PR #459, opened 2021-09-13
Local deploy maps all available interfaces

brahmlower:
This will automatically map any unmapped interfaces when deployed locally. Implicit interface mapping is disabled if you deploy locally with the `--production` flag. Deploying multiple instances of a component will also cause implicit interface mappings to be disabled since we can't map a single interface to multiple components.


brahmlower:
do not merge- I've just found this breaks support for ingress subdomains on interfaces 😕

brahmlower:
Okay- this is ready for review again 👍

PR #461, opened 2021-09-15
adding component interface configs to interface nodes

ryan-cahill:


tjhiggins:
Missed this - already merged these same changes

PR #462, opened 2021-09-16
273 spec sidecar support

brahmlower:
This PR adds sidecars to the service definition. Sidecars were originally added to the component level here, but that functionality has been removed. It does not start the sidecars during local deployment


src/dependency-manager/src/spec/utils/json-schema.ts
```diff
@@ -114,7 +116,7 @@ const mergeDebugSpec = (definitions: Record<string, SchemaObject>): Record<strin
     delete definition.allOf;
     debug_definitions[`${DEBUG_PREFIX}${key}`] = definition;
 
-    if (definition?.properties?.debug && (key === task_spec_name || key === service_spec_name)) {
+    if (definition?.properties?.debug && (key === task_spec_name || key === service_spec_name || key === sidecar_spec_name)) {
```
tjhiggins, line 119: Feels weird to support a debug block for sidecars. Can you undo these changes and remove the debug property on ResourceSpec. Then add an explicit debug property on both Task/ServiceSpec.

tjhiggins (changes requested):

src/dependency-manager/src/spec/utils/json-schema.ts
```diff
@@ -101,12 +102,18 @@ const recursivelyReplaceDebugRefs = (obj: SchemaObject) => {
...
   const service_spec_name = ServiceSpec.name;
+  const sidecar_spec_name = SidecarSpec.name;
   const task_spec_name = TaskSpec.name;
   const debug_field = 'debug';
 
   const definitions_copy = JSON.parse(JSON.stringify(definitions)) as Record<string, SchemaObject>;
   const debug_definitions: Record<string, SchemaObject> = {};
   for (const [key, definition] of Object.entries(definitions_copy)) {
+    // Don't create or reference a debug version of a sidecar property
+    if (key === sidecar_spec_name) {
```
tjhiggins, line 113: Instead of ignoring here - I think we should remove `debug` from the base ResourceSpec.

```ts
@IsOptional()
@ValidateNested()
debug?: Partial<ResourceSpec>;
```

and define on both ServiceSpec/TaskSpec explictly.
ex.
```ts
@IsOptional()
@ValidateNested()
debug?: Partial<ServiceSpec>;
```

tjhiggins (changes requested):

src/dependency-manager/src/spec/utils/json-schema.ts
brahmlower: Debug was removed from the ResourceSpec in the [latest commit](https://github.com/architect-team/architect-cli/pull/462/commits/23eaad1ff2d03759a1bc92902d26858911d89139?file-filters%5B%5D=.ts#diff-af2afb004594711094b0e6eb4a40aa7af8e9e06f7e12d14b857049cd755f4246L175-L178). The ServiceSpec and TaskSpec already had the debug property defined, which is why they're not showing up in the commit diffs.

Even with that change though, the sidecars are included in the definitions that `mergeDebugSpec` iterates over.

PR #463, opened 2021-09-23
adding ip whitelist to ingress

ryan-cahill:


src/dependency-manager/src/spec/component-spec.ts
```diff
@@ -24,6 +24,13 @@ export class IngressSpec {
     description: 'The subdomain that will be used if the interface is exposed externally (defaults to the interface name)',
   })
   subdomain?: string;
+
+  @IsOptional()
+  @JSONSchema({
+    ...ArrayOf('string'),
+    description: 'IP addresses that are allowed to access the interface',
+  })
+  whitelist?: string[];
```
tjhiggins, line 33: @davidthor thoughts on naming here? Debating if we include ip `ipwhitelist` or `ip_whitelist`

tjhiggins: Rename to `ip_whitelist`

test/dependency-manager/interpolation.test.ts
```diff
@@ -1315,6 +1315,60 @@ describe('interpolation spec v1', () => {
...
+      '/stack/architect.yml': component_config,
+    });
+
+    const manager = new LocalDependencyManager(axios.create(), {
+      'examples/hello-world': '/stack/architect.yml',
+    });
+    const config = await manager.loadComponentConfig('examples/hello-world');
+    const graph = await manager.getGraph(
+      await manager.loadComponentConfigs(config),
+      // @ts-ignore
```
tjhiggins, line 1350: Remove ts-ignore and make values.yml support more than just strings? Curious for your thoughts here.

tjhiggins (changes requested):

test/dependency-manager/interpolation.test.ts
ryan-cahill: I hadn't considered this case. Do you see any reasons not to enable something like this besides the fact that this might be changing very soon? I mostly imagined this feature being hardcoded though because it's not something that someone would use for local development, and I don't really expect it to change frequently, if at all, for remote deployments

PR #465, opened 2021-09-27
ip whitelist docs

ryan-cahill:


docs/1-components/8-ingress-rules.md
```diff
@@ -34,3 +34,25 @@ The above component declares a configurable ingress rule to listen on a subdomai
 # The following command will expose the interface at https://app.dev.example.arc.domains
 $ architect deploy example/component:latest -a example -e dev
 ```
+
+### Whitelisting IP Addresses
+
+Ingresses of Architect components can be whitelisted by IP addresses on an individual basis. An example of a component where only users at the IP address 100.100.100.100 can access the `app` endpoint is shown below.
```
tjhiggins, line 40: Can you mention it accepts ipv4 or cidr? Add an example of a cidr.

tjhiggins (changes requested):

PR #471, opened 2021-10-05
Added slugs for sidecars

brahmlower:


src/dependency-manager/src/spec/transform/sidecar-transform.ts
```diff
@@ -3,12 +3,15 @@ import { SidecarConfig } from '../../config/sidecar-config';
...
 import { transformResourceSpec } from './resource-transform';
+import { ComponentSlugUtils, ServiceSidecarVersionSlugUtils } from '../utils/slugs';
 
 export const transformSidecarSpec = (key: string, spec: SidecarSpec, component_ref: string, tag: string, instance_metadata?: ComponentInstanceMetadata): SidecarConfig => {
   const resource_config = transformResourceSpec(key, spec, component_ref, tag, instance_metadata);
+  const { component_account_name, component_name } = ComponentSlugUtils.parse(component_ref);
 
   return {
     ...resource_config,
+    ref: ServiceSidecarVersionSlugUtils.build(component_account_name, component_name, resource_config.name, key, tag, instance_metadata?.instance_name),
```
tjhiggins, line 14: Out of curiosity - why do sidecars need a ref?

brahmlower: The ref can be used to give the sidecar container a unique name when deploying to the platform. Containers for services are currently named based on their ref, so I figured maybe it was worth doing the same for the sidecars for consistency. Maybe there's more meaning to the ref though that I'm unaware of 🤔

brahmlower:
Closing this PR- after further conversation we think we can avoid needing a ref for sidecars.

PR #473, opened 2021-10-08
Cleanup resource config volumes

brahmlower:
Removed Volumes from the base Resources and added it only to Services since volumes are not supported by Sidecars nor Tasks.

### DIFF
--- a/src/dependency-manager/src/spec/transform/resource-transform.ts
+++ b/src/dependency-manager/src/spec/transform/resource-transform.ts
@@ -78,22 +78,6 @@ export const transformBuildSpec = (build: BuildSpec | undefined, image?: string)
   };
 };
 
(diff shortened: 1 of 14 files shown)


src/dependency-manager/src/spec/transform/resource-transform.ts
```diff
@@ -78,21 +78,21 @@ export const transformBuildSpec = (build: BuildSpec | undefined, image?: string)
...
-      description: volume.description,
-      readonly: volume.readonly,
-    };
-  } else {
-    return {
-      host_path: volume,
-    };
-  }
-};
+// export const transformVolumeSpec = (key: string, volume: VolumeSpec | string): VolumeConfig => {
```
tjhiggins, line 81: remove?

tjhiggins (changes requested):

PR #476, opened 2021-10-14
Graph edges for outputs

brahmlower:
This PR adds graph edges to represent a dependency on component outputs. Prior to this, there could only be a single edge between two nodes, but this PR makes it possible for multiple edges to exist between two nodes (scaling with the number of edge types). Just as a ServiceEdge can have multiple interface mappings associated with it, an OutputEdge can have multiple outputs. This means right now we will see a maximum of two edges between nodes.


src/dependency-manager/src/manager.ts
```diff
@@ -635,9 +656,15 @@ export default abstract class DependencyManager {
       const interpolated_config = interpolateConfigOrReject(component_config, [''], false);
       nodes = nodes.concat(this.getComponentNodes(interpolated_config));
 
-      if (Object.keys(component_config.interfaces).length) {
+      const has_interfaces = Object.keys(component_config.interfaces).length > 0
```
tjhiggins, line 659: fix lint warnings

src/dependency-manager/src/graph/node/interfaces.ts
```diff
@@ -1,16 +1,16 @@
 import { DependencyNode } from '.';
-import { ComponentInterfaceConfig } from '../../config/component-config';
+import { ComponentNodeConfig } from '../../config/component-config';
 import { ServiceInterfaceConfig } from '../../config/service-config';
 import { Dictionary } from '../../utils/dictionary';
 
-export default class InterfacesNode extends DependencyNode {
+export default class ComponentNode extends DependencyNode {
```
tjhiggins, line 6: Nice - should we rename this file to component.ts?

PR #478, opened 2021-10-21
Added docs detailing multi-tenant deployments

brahmlower:


docs/2-deployments/7-multitenant-deployments.md
```diff
@@ -0,0 +1,44 @@
+---
+title: Multi-Tenant Deployments
+---
+
+# Multi-Tenant Deployments
+
+Multi-tenant deployments are when
```
dan-barrett, line 7: are when...?

docs/2-deployments/7-multitenant-deployments.md
```diff
@@ -0,0 +1,44 @@
...
+
+By default, deploying a component will create it as a singleton in the environment, where it is the default instance.
```
dan-barrett, line 9: This section might need some small note or aside as to how this is different from horizontal scaling. (ie, we probably don't want to leave the customer confused and think they need to deploy multiple instances in order to horizontally scale).

docs/2-deployments/7-multitenant-deployments.md
```diff
@@ -0,0 +1,44 @@
...
+
+```
+architect-local deploy examples/react-app@app1
+```
+
+This process works for updating existing deployments as well. If an "app1" tenant of the `examples/react-app` component had already been deployed, the above command would have deployed to that same tenant, leaving any other instances of the component untouched.
+
+## Deployments and Management Lifecycle
+
+When you view your deployed components, you'll see that the react-app has the name "app1" whereas the typical deployment without a tenant name will show only the component. In this example we've deployed the example React App twice with the tenant names "app1" and "app2", and once without a tenant name.
```
dan-barrett, line 21: "When you view your multi-tenant components..."

docs/2-deployments/7-multitenant-deployments.md
```diff
@@ -0,0 +1,44 @@
...
+Configuring individual tenants works the same way as deploying those tenants: by providing the tenant name. Secrets use the component scopes to specify which values will be applied to which components, so setting a secret for a tenant is as simple as providing the tenant name in the component scope.
+
+In the case of the example React App, the following will override the `world_text` parameter for just the `sandiego` tenant.
+
+```yaml
+'examples/react-app@sandiego':
+  world_text: San Diego
+```
+
+This enables minimal dupliation of configuration, as you can specify any necessary values in a generic component scope, but then override specific values on a tenant-by-tenant basis. In the following example, the `planetearth` tenant would have the `foo` parameter as configured in the first scope, but would have the overwritten values for `world_text` set in the second scope.
```
dan-barrett, line 36: *duplication

docs/2-deployments/7-multitenant-deployments.md
brahmlower: 😱  oh my god that's embarrassing- thanks for catching this 🤦

docs/2-deployments/7-multitenant-deployments.md
brahmlower: I just meant to refer to the list of components that are deployed to the environment, not just the multi-tenant deployments. I could rephrase the start of this to say The "Environment's components list page" instead if you think that communicates the point better?

PR #482, opened 2021-10-29
Actionable deployment error messages

brahmlower:
This PR changes deployment error messages to include the pipeline ID and links to any failed deployments.

```
% architect-local deploy examples/scheduled-tasks -a brahm-testing -e test -p 'schedule=0 0 * * ?'
Creating pipelines... done
Pipeline ready for review: http://cloud.arc.localhost:81/brahm-testing/environments/test/pipelines/a09f82ac-49d8-4de4-9f49-e22d3b393793
? Would you like to apply? Yes
Deploying... !
 ›   Error: Error
 ›   Pipeline a09f82ac-49d8-4de4-9f49-e22d3b393793 failed because 1 deployment failed:
 ›   - http://cloud.arc.localhost:81/brahm-testing/environments/test/deployments/71716bf8-e1a7-4ba2-b583-577732dcd821
 ›       at Timeout.<anonymous> (/Users/brahmlower/development/architect-cli/src/common/utils/pipeline.ts:59:18)
 ›       at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

The transition from the list of URLs to the stacktrace looks kinda odd, but I'm not sure how big of a problem that will really be.

This PR includes a utility class for easily building string templates that include mostly static values, and dynamic values.


src/common/utils/pipeline.ts
```diff
@@ -1,17 +1,62 @@
...
+
+    const deployment_string = deployment_links.length > 1
+      ? `${deployment_links.length} deployments`
+      : '1 deployment';
+    const listified_link_string = deployment_links.map((s: string) => `- ${s}`).join('\n');
+    this.message = `Pipeline ${pipeline_id} failed because ${deployment_string} failed:\n${listified_link_string}`;
+  }
+}
+
+const deploymentUrlBuilder = new StringTemplateBuilder(
```
tjhiggins, line 24: I'm not convinced we need a string template builder. Is the problem that we don't want to hardcode api endpoints  in the cli?

Feels like more code to define the same url:
```ts
const environment_url = `${app_host}/${account.name}/environments/${environment.name}`;
const deployment_url = `${environment_url}/deployments/${deployment.id}`;
```

If we wanted to take out the guess work we could create controller methods like the cloud/api:
`app.api.pipelines.getDeployments()`
or helpers to get a url
`app.cloud.deployments.getUrl(deployment)`

src/common/utils/pipeline.ts
```diff
@@ -1,17 +1,62 @@
...
+            account: pipeline.environment.account.name,
+            environment: pipeline.environment.name,
+          });
           clearInterval(poll);
-          reject(new Error('Pipeline failed'));
+          // Query deployments for pipline to determine cause of failure
+          const { data: deployments } = await app.api.get(`/pipelines/${pipeline.id}/deployments`);
+
+          // Check if the deployment failed due to a user aborting the deployment
+          const aborted_deployments = deployments.filter((d: any) => d.aborted_at);
```
tjhiggins, line 49: This check would need to be that every deployment is either applied_at or aborted_at. Since if a deployment hard fails the rest of the deployments that depend on it get aborted.

```ts
// common/utils/deployment.ts
const { data: deployments } = await app.api.get(`/pipelines/${pipeline.id}/deployments`) as Deployment[];
const pipeline_aborted = deployments.every(d => d.applied_at || d.aborted_at);
```

src/common/utils/pipeline.ts
```diff
@@ -1,17 +1,62 @@
...
         if (pipeline.failed_at || poll_count > 180) {  // Stop checking after 30min (180 * 10s)
+          const deploymentUrl = deploymentUrlBuilder.with({
+            app_host: app.config.app_host,
+            account: pipeline.environment.account.name,
+            environment: pipeline.environment.name,
+          });
           clearInterval(poll);
-          reject(new Error('Pipeline failed'));
+          // Query deployments for pipline to determine cause of failure
+          const { data: deployments } = await app.api.get(`/pipelines/${pipeline.id}/deployments`);
```
tjhiggins, line 46: Need to add a special case for platform pipelines. Since a pipeline can be for either an environment or a platform.

```ts
if (pipeline.platform) {
... // Unclear what we link to besides the platform page
}
```

tjhiggins (changes requested):

src/common/utils/pipeline.ts
brahmlower: I don't think that's quite right- if every deployment was successful and none were aborted, `pipeline_aborted` would still evaluate to `true` in your example. If a deployment fails and the following deployments are all aborted, then should it be find to just filter by `aborted_at` and check if the list is empty or not

src/common/utils/pipeline.ts
brahmlower: The problem was reducing code repetition and improving readability. In the case of the deployment urls, there are 4 variables that need to be templated, which got relatively messy to write out inline:

```
const deploymentUrl = `${app.config.app_host}/${pipeline.environment.account.name}/environments/${pipeline.environment.name}/deployments/${aborted_deployments[0].id}`
```

Abstracting this out to a function helped a bit:

```
const buildDeploymentUrl = (app_host, account, environment, deployment) =>
  `${app_host}/${account}/environments/${environment}/deployments/${deployment}`;

...

const deploymentUrl = buildDeploymentUrl(
  app.config.app_host,
  pipeline.environment.account.name,
  pipeline.environment.name
  aborted_deployments[0].id
);
```

But this is still a lengthy function call, and we need to build this URL in multiple spots which added a lot of visual heft to the `pollPipeline` function overall.

Between the two calls though the only thing that was changing was the final deployment id, which made me want to just curry those first 3 function inputs. Though the template builder is using the builder pattern, it achieves the same ergonomics as currying in this case, letting us set the host, account and env name just once and then build various deployment urls as needed without having to repeat inputs

src/common/utils/pipeline.ts
tjhiggins: It'd only work in this block since the pipeline.failed_at is set

src/common/utils/pipeline.ts
tjhiggins: Yeah I definitely agree that its ugly in a one liner. Still not sold on the need for string templating since typescript does that already via `${...}`. 

I'd rather hide away the building of the urls.  That way a dev doesn't need to know how to build. Something like:
```ts
app.cloud.deployments.getUrl(deployment);

getUrl(deployment: Deployment) {
  const environment = deployment.pipeline.environment;
  if (environment) {
    return `${this.config.app_host}/${environment.account.name}/environments/${environment.name}/deployments/${deployment.id}`
  } else if (platform) {
    return ...
  }
}
```

PR #516, opened 2022-01-16
Bring deploy -l into its own command

mueschm:
## Overview
This diff takes the deploy command and splits it into two separate commands. Deploy is now responsible for deploying an architect application to a platform. Dev is responsible for local deployments. Now ```architect deploy``` will only kick off architect pipelines, where as ```architect dev``` will keep a stream of all component logs in the current terminal session.

Functionally this changes nothing! The only goal is to lay the groundwork for future changes.

This also includes updates to the docs for the new command.

## Testing
Run unit tests
```
npm run test
```
Test the dev command
```
cd examples/react-app
architect dev -e dev
```
Test the deploy command
```
architect deploy examples/react-app
```


src/commands/dev.ts
```diff
@@ -0,0 +1,297 @@
...
+    return false;
+  }
+
+  static description = 'Run your stack locally';
+
+  static flags = {
+    ...DevCommand.flags,
+    ...EnvironmentUtils.flags,
+
+    compose_file: flags.string({
```
tjhiggins, line 47: Since its a new command can we remove the deprecated flags?

src/commands/deploy.ts
```diff
@@ -70,42 +60,16 @@ export abstract class DeployCommand extends Command {
...
 
-  static description = 'Create a deploy job on Architect Cloud or run stacks locally';
+  static description = 'Create a deploy job on Architect Cloud';
 
   static flags = {
     ...DeployCommand.flags,
     ...AccountUtils.flags,
     ...EnvironmentUtils.flags,
 
-    local: flags.boolean({
```
tjhiggins, line 84: Could we support --local for a period of time? Add a deprecated label `${Command.DEPRECATED} ..`.

It says its an anti-pattern, but it'd be temporary.

```ts
if (flags.local) {
  // https://oclif.io/docs/running_programmatically#calling-commands-directly
  await DevCommand.run([...])
}
```

src/commands/deploy.ts
```diff
@@ -70,42 +60,16 @@ export abstract class DeployCommand extends Command {
...
-      char: 'l',
-      description: 'Deploy the stack locally instead of via Architect Cloud',
-      exclusive: ['account', 'auto-approve', 'auto_approve', 'refresh'],
-    }),
-    production: flags.boolean({
```
tjhiggins, line 89: Sorry for the merge conflicts from:
https://github.com/architect-team/architect-cli/commit/73f5aa2036c9e543c3b544f54f2e259dd9e3a310

tjhiggins (changes requested):

src/commands/dev.ts
mueschm: Yeah we can deprecate on this command, and for the proxy on the other I can just translate to the new one.

src/commands/deploy.ts
mueschm: This makes sense and is easy enough. If we do this though, I would prefer to add an update by timeframe. This would also let us put a notice for when users need to upgrade their usage of the command by. For this specific instance I would say a month is more than adequate.

src/commands/dev.ts
mueschm: So turns out proxying the commands is not as simple as I thought. Instead I added to warning all deprecated flags will be removed and have them hidden.

src/commands/deploy.ts
```diff
@@ -489,6 +263,7 @@ export default class Deploy extends DeployCommand {
   }
 
   async run(): Promise<void> {
+    this.log(JSON.stringify(this.argv));
```
tjhiggins, line 266: debug code?

tjhiggins (changes requested):

src/commands/deploy.ts
mueschm: No, clearly this is very important code! (Already removed thanks for the catch!)

PR #517, opened 2022-01-20
Updated all the docs and examples that used deploy -l

mueschm:
## Overview
This PR follows up the changes made to the CLI to remove the --local flag. 

All docs/examples have been updated to remove the usage of the ``` architect deploy --local``` command and replace it with the ```architect dev``` command. At the same time I removed some of the deprecated flags such as ```--production```.

## Testing
Viewed each change in a Markdown viewer to make sure the formatting looked correct. Otherwise discussed any possible confusion a user may have over the intent of the documentation with @tjhiggins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,9 +189,36 @@ ALIASES
 
 _See code: [src/commands/config/view.ts](https://github.com/architect-team/architect-cli/blob/v1.13.1/src/commands/config/view.ts)_
 
+## `architect dev [CONFIGS_OR_COMPONENTS]`
+
+Run your stacks locally
+
+```
+USAGE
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -189,9 +189,36 @@ ALIASES
 
 _See code: [src/commands/config/view.ts](https://github.com/architect-team/architect-cli/blob/v1.13.1/src/commands/config/view.ts)_
 
+## `architect dev [CONFIGS_OR_COMPONENTS]`
```
tjhiggins, line 192: For future ref - this should be autogenerated by oclif during a release. See the `<!-- commands -->` block. Oclif uses that as a marker to insert generated commands.

PR #518, opened 2022-01-20
Adding flags to recreate containers and volumes for local

ryan-cahill:


mueschm:
What problem is this solving? I can understand ```--renew-anon-volumes``` but ```--force-recreate``` does not make sense to me. If the image has not changed why rebuild? Though I do appreciate you are not using the shorthand for the commands.

ryan-cahill:
Agreed, `--force-recreate` wasn't actually necessary

src/common/docker-compose/index.ts
```diff
@@ -203,7 +203,8 @@ export class DockerComposeUtils {
               compose.volumes[spec.key] = { external: true };
               volume = `${spec.key}:${service_volume}${spec.readonly ? ':ro' : ''}`;
             } else {
-              volume = service_volume;
+              volume = `${key}:${service_volume}`;
```
tjhiggins, line 206: is this right? if I don't specify a host_path I'd expect docker to create a volume
https://docs.docker.com/compose/compose-file/compose-file-v3/

```
volumes:
  # Just specify a path and let the Engine create a volume
  - /var/lib/mysql
  ````

tjhiggins (changes requested):

src/common/docker-compose/index.ts
ryan-cahill: It would, but that volume would be anonymous. That wouldn't work if want volumes specified in a service to be persistent between redeploys

tjhiggins: ah that's what I was afraid of - annoying that anon volumes mean explicit and non-explicit. That makes no sense.

PR #520, opened 2022-01-25
Added support for local logs.

mueschm:
## Overview
This PR adds support for querying logs for a local service. The following assumptions are now made.

1. If an environment flag is set then we check if it matches a local env first. If it does we assume the user is looking for a local service.
2. If an account flag is set we automatically assume the user is looking for a remote service.
3. If no flags are set we ask for the account the user is looking for. We add a new option called "dev (Local Machine)" that the user can select for a local service.

NOTES: The since flag does not work locally and is ignored. While kubernetes supports since, docker does not. We could implement this feature ourselves but it did not seem worthwhile in the short term.

## Changes
1. Added the concept of a local account and the ability to check if an account is a local account.
2. Added support for getting a list of local environments.
3. Added support for detecting if an environment is a local one.
4. Added support for getting list of local services for a particular environment
5. Added support for getting logs from a local service.

## Testing
1. Setup a local env with the ```architect dev``` command.
2. Deploy a service to a remote platform with the ```architect deploy```` command.
3. Run ```architect logs``` and interact with the autocomplete options to test the functionality.
4.  Test the following flags. --tail --follow --raw --since

## Sample Output
Logs

### DIFF
--- a/src/architect/account/account.utils.ts
+++ b/src/architect/account/account.utils.ts
@@ -14,10 +14,22 @@ export default class AccountUtils {
     }),
   };
 
-  static async getAccount(app: AppService, account_name?: string, account_message?: string): Promise<Account> {
+  static getLocalAccount(): Account {
+    // Account ids are UUID so there is no chance of collision
+    return {
+      id: 'dev',
+      name: 'dev (Local Machine)',
+    };
+  }
+
+  static isLocalAccount(account: Account): boolean {
+    return account.id === "dev";
+  }
+
+  static async getAccount(app: AppService, account_name?: string, account_message?: string, ask_local_account?: boolean): Promise<Account> {
     const config_account = app.config.defaultAccount();
     // Set the account name from the config only if an account name wasn't set as cli flag
-    if (config_account && !account_name) {
+    if (config_account && !account_name && !ask_local_account) {
       account_name = config_account;
     }
 
@@ -41,6 +53,9 @@ export default class AccountUtils {
     } else {
       inquirer.registerPrompt('autocomplete', require('inquirer-autocomplete-prompt'));
       let accounts: Account[] = [];
+      if (ask_local_account) {
+        accounts.push(this.getLocalAccount());
+      }
       const answers: any = await inquirer.prompt([
         {
           type: 'autocomplete',
@@ -49,7 +64,7 @@ export default class AccountUtils {
           filter: (x) => x, // api filters
           source: async (answers_so_far: any, input: string) => {
             const { data } = await app.api.get('/accounts', { params: { q: input, limit: 10 } });
-            accounts = data.rows;
+            accounts = accounts.concat(data.rows);
             return accounts;
           },
         },
--- a/src/commands/logs.ts
+++ b/src/commands/logs.ts
@@ -215,7 +215,7 @@ export default class Logs extends Command {
     // If no account is default to local first.
     if (!flags.account && flags.environment) {
       // If the env exists locally then just assume local
-      const is_local_env = DockerComposeUtils.isLocalEnvironment(this.app.config.getConfigDir(), flags.environment);
+      const is_local_env = await DockerComposeUtils.isLocalEnvironment(this.app.config.getConfigDir(), flags.environment);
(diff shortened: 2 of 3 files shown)


src/commands/logs.ts
```diff
@@ -43,12 +47,90 @@ export default class Logs extends Command {
...
+      decodeStrings: false,
+    });
+
+    logger._transform = function (chunk, _encoding, done) {
+      chunk.toString().split('\n').forEach((line: string) => {
+        // This removes the service name so that the logger
+        // can display it the same way regardless of where the logs
+        // came from.
+        if (!flags.raw) {
+          line = line.substr(line.indexOf('|') + 1);
```
tjhiggins, line 117: substr is deprecated

src/commands/logs.ts
```diff
@@ -43,12 +47,90 @@ export default class Logs extends Command {
...
+  }
+
+  async runLocal(): Promise<void> {
+    const { args, flags } = this.parse(Logs);
+
+    const environment_name = await DockerComposeUtils.getLocalEnvironment(this.app.config.getConfigDir(), flags.environment);
+    const compose_file = DockerComposeUtils.buildComposeFilepath(this.app.config.getConfigDir(), environment_name);
+    const service_name = await DockerComposeUtils.getLocalServiceForEnvironment(environment_name, compose_file, args.resource);
+
+    const compose_args = ['-f', compose_file, '-p', environment_name, 'logs'];
```
tjhiggins, line 91: Nice job! Tested it out a bit locally and found some quirks:
1. I can get logs for services/environments that aren't running
    - Don't show environments that aren't running
    - Don't show services that aren't running
2. Not all logs showing when running with and without --follow
   - ex. architect logs -e architect cloud-api
   - I'm not seeing some logs towards the end or when I hot reload our app

src/architect/account/account.utils.ts
```diff
@@ -14,10 +14,22 @@ export default class AccountUtils {
     }),
   };
 
-  static async getAccount(app: AppService, account_name?: string, account_message?: string): Promise<Account> {
+  static getLocalAccount(): Account {
+    // Account ids are UUID so there is no chance of collision
+    return {
+      id: 'dev',
+      name: 'dev (Local Machine)',
```
tjhiggins, line 21: Nice! I don't mind this solution and I can still have a default account for every other cmd.

tjhiggins (changes requested):

src/commands/logs.ts
tjhiggins: Actually I'm not so sure about my stance on the not showing services that aren't running. Since a service might start and die and you'd want to see the logs. Maybe make it more clear that the service died or isn't running? Thoughts?

mueschm: I can pull the status field out and display that as well so it would look like
```
cloud-api (RUNNING)
```
Which would make it more clear. Though I do like the idea of showing all services. If they are part of the local environment then I most likely want to interact with them in some way shape or form and even just knowing they are dead is useful.

Will look into the other issues as I did not hit those.

src/commands/logs.ts
mueschm: Did not know that. Will update.

src/commands/logs.ts
mueschm: @tjhiggins few things
1) An environment is considered to be live if the docker compose yml file exists in the .config folder. I would prefer to cleanup that file on stop rather than do some weird checks to see whats live. Also would prefer to use the .config folder as a source of truth since other things may be running in local docker instance that we don't control.
2. I cannot replicate the log issue, can you share an example so I can get some insight?

tjhiggins: 1. Sounds good to me

2. For the logs issue I'm running:
   - `architect deploy -l auth0.architect.yml -e architect -v values.yml`
   - `architect logs --follow -e architect cloud-api`

mueschm: Found the log issue. For some reason execa is eating parts of stdout. Have no idea why, but if we use spawn everything works perfectly fine. Current version fixes the issue.

PR #542, opened 2022-03-24
Draft: Allow user to run exec without being logged in

Zach-at-Arch:
Allows users to access their locally run instances without logging in

Added the following changes:
* exec and logs function:
- set auth_required() to false to allow use when not logged in

*account.utils
- Updated getAccount() to check if the user is logged in
	- if not logged in, warn user and default to local account


mueschm:
We went over this in person. Zack is going to make some changes to move the logic into AccountUtils and we will review again after.

Zach-at-Arch:
Pushed logic to check if a user is logged in to getAccount() function in Utils. Also updated logs command so that it can run locally like exec

mueschm (changes requested):

PR #543, opened 2022-03-25
[WIP] 398 dependency manager consolidation

ryan-cahill:


.github/workflows/cli-rc.yml
```diff
@@ -1,6 +1,10 @@
 name: CLI - Cut Release Candidate
 
-on: push
+on:
```
mueschm, line 3: Why is this change needed?

mueschm (changes requested):

.github/workflows/cli-rc.yml
ryan-cahill: Part of this work is to make an update to the GitHub actions to publish CLI RCs on pull requests

PR #545, opened 2022-03-29
Add docs for exec and logging.

mueschm:
## Overview
Added a new docs page that explains how to use architect exec and logs. This provides a couple of basic examples showing how to use these commands.

## Testing
1) Used Gatsby to veirfy the style
2) Ran through Grammerly to make sure it was grammatically correct.

## Picture
image

### DIFF
--- a/docs/2-deployments/7-observability.md
+++ b/docs/2-deployments/7-observability.md
@@ -1,0 +1,59 @@
+---
+title: Observability
+---
+# Observability
+Up until now, you have used Architect to develop and deploy your application. However, Architect's role in your development process does not end there. You still need to monitor the system and be able to dive in when a problem occurs. To help facilitate this, Architect provides some helpful commands that will allow you better insights into how your system is doing.
+
+Whether it be a remote deployment using [`architect deploy`](/reference/cli/#architect-deploy-configs_or_components) or a local one using [`architect dev`](http://localhost:8000/reference/cli/#architect-dev-configs_or_components) these commands will give you a deeper insight into your product.
+
+We will assume that you have deployed the example react app to a remote environment for the following examples.
+```sh
+$ architect register ./examples/react-app/
+$ Architect deploy react-app
+```
+
+## Logging
+
+As your application runs, it will generally print important information to stdout and stderr as it runs. The log command will allow you to stream those logs from a remote service to your console. For instance, let's say you wanted to see the latest logs for the API of the example react app. You would do the following.
+```sh
+$ architect logs -a my_account -e my_environment react-app.services.api
+```
+*If you are unsure what your account, environment, or service name are, you can leave them blank, and Architect will prompt you will a list of valid options.*
+The output would be
+```sh
+Logs:
+―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
+api | {"message":"GET /names","level":"info","service":"backend"}
+api | Executing (default): SELECT "id", "name", "createdAt", "updatedAt" FROM "names" AS "na
+api | me" ORDER BY "name"."id" DESC;
+```
+
+To have a constant stream of logs, you can add the `--follow` flag to the command.
+```sh
+$ architect logs -a my_account -e my_environment react-app.services.api
+```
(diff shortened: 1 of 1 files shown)


docs/2-deployments/7-observability.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Observability
+---
+# Observability
+Up until now, you have used Architect to develop and deploy your application. However, Architect's role in your development process does not end there. You still need to monitor the system and be able to dive in when a problem occurs. To help facilitate this, Architect provides some helpful commands that will allow you better insights into how your system is doing.
+
+Whether it be a remote deployment using [`architect deploy`](/reference/cli/#architect-deploy-configs_or_components) or a local one using [`architect dev`](http://localhost:8000/reference/cli/#architect-dev-configs_or_components) these commands will give you a deeper insight into your product.
```
tjhiggins, line 7: remove http://localhost:8000

docs/2-deployments/7-observability.md
```diff
@@ -0,0 +1,59 @@
...
+```sh
+$ architect register ./examples/react-app/
+$ Architect deploy react-app
+```
+
+## Logging
+
+As your application runs, it will generally print important information to stdout and stderr as it runs. The log command will allow you to stream those logs from a remote service to your console. For instance, let's say you wanted to see the latest logs for the API of the example react app. You would do the following.
+```sh
+$ architect logs -a my_account -e my_environment react-app.services.api
```
tjhiggins, line 19: Should we also showcase how to use this commands for local dev?

tjhiggins (changes requested):

docs/2-deployments/7-observability.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Observability
+---
+# Observability
+Up until now, you have used Architect to develop and deploy your application. However, Architect's role in your development process does not end there. You still need to monitor the system and be able to dive in when a problem occurs. To help facilitate this, Architect provides some helpful commands that will allow you better insights into how your system is doing.
+
+Whether it be a remote deployment using [`architect deploy`](/reference/cli/#architect-deploy-configs_or_components) or a local one using [`architect dev`](http://localhost:8000/reference/cli/#architect-dev-configs_or_components) these commands will give you a deeper insight into your product.
+
+We will assume that you have deployed the example react app to a remote environment for the following examples.
```
ryan-cahill, line 9: Maybe also link to the react app example here?

docs/2-deployments/7-observability.md
```diff
@@ -0,0 +1,59 @@
...
+```sh
+$ architect register ./examples/react-app/
+$ Architect deploy react-app
+```
+
+## Logging
+
+As your application runs, it will generally print important information to stdout and stderr as it runs. The log command will allow you to stream those logs from a remote service to your console. For instance, let's say you wanted to see the latest logs for the API of the example react app. You would do the following.
```
ryan-cahill, line 17: `as it runs` at the end is redundant

ryan-cahill, line 17: `logs` command rather than log command

docs/2-deployments/7-observability.md
```diff
@@ -0,0 +1,59 @@
...
+Which would output
+```sh
+Dockerfile         node_modules       package.json
+logs               package-lock.json  src
+```
+While this is great for quick one-off commands, sometimes a more interactive approach would be preferred. Instead, you can create an interactive session with the remote container using the `-ti` flags.
+```sh
+$ architect exec -ti -a my_account -e my_environment react-app.services.api -- sh
+```
+Which would provide us with an interactive terminal.
```
ryan-cahill, line 53: "Which would provide us" to "Which would provide you"

ryan-cahill (changes requested):

docs/2-deployments/7-observability.md
mueschm: Added in a new section that goes over the difference between Remote and Local. Also moved the guidance around what to do if you have no idea about what accounts or envs you have there.

ryan-cahill (approved):

PR #546, opened 2022-03-29
DRAFT: Added support for semantic release.

mueschm:
## Overview
This converts us over to semantic release. With this, we want the following.
1) Allow our commit messages to generate a changelog.
2) Auto-update the version of the package based on the commit type.
3) Add both an RC and ARC branch that allow for more generic testing as well as release candidate testing.
4) Remove the push-button release and move everything into occurring based on Merge Requests.
5) Support both NPM and Github Releases.
6) This also removes the current rc process and now lint and tests are run for everything. This will force a sanity check at each step.

## Testing
While I have tested most of this in a dummy repo. This is really something that can only be tested live. So the test plan is as follows.
1) Have someone review this PR
2) Land into master
3) Merge a small change into the RC branch
4) Verify everything worked as expected.


.github/workflows/publish.yml
```diff
@@ -1,84 +1,81 @@
...
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - run: npm ci
+      - run: npm run lint
+      - run: npm run check-circular
+  test:
+    strategy:
+      matrix:
+        node-version: [12.x, 14.x, 16.x]
```
ryan-cahill, line 38: Wanted to note that node 12 EOL is the end of this month, and node 17 is out. We might want to take this chart into consideration

https://nodejs.org/en/about/releases/

mueschm: We should do this as another PR but good callout.

ryan-cahill: Sure, made a ticket for that

.github/workflows/publish.yml
```diff
@@ -1,84 +1,81 @@
...
-      - name: Archive artifact
-        uses: actions/upload-artifact@v2
-        with:
-          name: dist
-          path: dist
       - name: npm publish
         run: |
           echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_PUBLISH_TOKEN }}" > .npmrc
-          npm publish
+          cross-env GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD) npm run semantic-release -- --dry-run
```
tjhiggins, line 79: Does cross-env exist in the image? Might make more sense to `npm install --save-dev cross-env` as part of the cli.

Then the package.json cmd would be:
`"semantic-release": "cross-env GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD) ./node_modules/semantic-release/bin/semantic-release.js"`

release.config.js
```diff
@@ -0,0 +1,68 @@
...
+    {
+      "name": "rc",
+      "prerelease": true
+    },
+    {
+      "name": "arc-*",
+      "prerelease": true
+    }
+  ],
+  plugins: branch === 'main' ? main_plugins : plugins,
```
tjhiggins, line 67: LGTM

.github/workflows/publish.yml
```diff
@@ -1,84 +1,81 @@
...
 on:
-  release:
-    types:
-      - published
+  push:
     branches:
-      - master
-    tags:
-      - v*.*.*
+      - main
```
tjhiggins, line 6: Like this a lot. Nice job consolidating the github actions into one.

tjhiggins (changes requested):

.github/workflows/publish.yml
tjhiggins: That's probably fine. Still my q applies. Is cross-env installed in the image for this job?

PR #550, opened 2022-04-04
Add restart for local services based on exit status or health

mueschm:
## Problem
On a remote deployment, kubernetes will restart  a service if there are issues with its health checks or if the container fails to start. Locally this does not happen. If a container fails to start then it just crashes. Health checks are generally ignored. The goal is to bring local deployments into parity with remote deployments.

## Changes
1) Add a new function in the CLI that monitors each running service and checks their health and running state.
2) If a service fails we restart it and reattach the logging for it.
3) If the service becomes unhealthy we 

## Testing
I created a sample app called Health that is as follows.
```
const express = require('express')
const app = express()
const port = process.env.PORT;

let stopService = false;

console.log(port.toString());

let pingCount = 0;

app.get('/ping', (req, res) => {
  pingCount++;
  console.log("PING" + pingCount);
  res.send("PONG");
  if (pingCount == 3) {
    stopService = true;
  }
})

app.get('/', (req, res) => {
  console.log("ROOT");
  res.send('Hello World!')
})

const server = app.listen(port, () => {
  console.log(`Example app listening on port ${port}`)
})


const main = async () => {
  while (!stopService) {
    await new Promise(r => setTimeout(r, 1000));
  }
  console.log("CLOSING APP 2");
  // The server is closed but the app is still running
  // Service is not healthy
  server.close();

  // Close the app after 10 seconds so the app is exited
  await new Promise(r => setTimeout(r, 10000000));
}

main();

```
architect.yml
```
name: health

services:
  health:
    build:
      context: .
    interfaces:
      main: &api-port 8081
    liveness_probe:
      path: /ping
      port: *api-port
      interval: 10s
      failure_threshold: 3
    environment:
      PORT: *api-port
```
Dockerfile
```
FROM node:16

WORKDIR /usr/src/app

COPY package*.json ./

RUN npm install

COPY . .

CMD [ "node", "index.js" ]
```

## Edge Cases
When a container is restarted or brought back online docker compose removes it from the log monitoring. It will only restart the logs if the container had a restart policy it could handle itself. This requires us to add the logs back by calling docker compose logs.

## Restart on Unhealthy
The liveness_probe is used to determine if a service is unhealthy. We essentially have two options for how to handle this.
1) Restart directly when a service becomes unhealthy
2) Wait for multiple checks of unhealthy before restarting
Based on the amount of time kubernetes takes to restart a service it seemed like number 2 was the best parity for remote deployments. So that is what I added.

## Pictures
image

### DIFF
--- a/src/commands/dev.ts
+++ b/src/commands/dev.ts
@@ -207,7 +207,9 @@ export default class Dev extends BaseCommand {
       compose_args.push('-d');
     }
 
-    await DockerComposeUtils.dockerCompose(compose_args, { stdio: 'inherit', env: { COMPOSE_IGNORE_ORPHANS: 'true' } });
+    const dockerComposeRunnable = DockerComposeUtils.dockerCompose(compose_args, { stdio: 'inherit', env: { COMPOSE_IGNORE_ORPHANS: 'true' } });
+    await DockerComposeUtils.watchContainersHealth(compose_file, project_name);
+    await dockerComposeRunnable;
     fs.removeSync(compose_file);
   }
 
--- a/src/commands/dev.ts
+++ b/src/commands/dev.ts
@@ -211,8 +211,7 @@ export default class Dev extends BaseCommand {
     try {
       await DockerComposeUtils.watchContainersHealth(compose_file, project_name);
     } catch (err) {
-      docker_compose_runnable.kill();
-      throw err;
+      console.error(err);
     }
     await docker_compose_runnable;
     fs.removeSync(compose_file);
--- a/src/commands/dev.ts
+++ b/src/commands/dev.ts
@@ -207,9 +207,14 @@ export default class Dev extends BaseCommand {
       compose_args.push('-d');
     }
 
-    const dockerComposeRunnable = DockerComposeUtils.dockerCompose(compose_args, { stdio: 'inherit', env: { COMPOSE_IGNORE_ORPHANS: 'true' } });
-    await DockerComposeUtils.watchContainersHealth(compose_file, project_name);
-    await dockerComposeRunnable;
+    const docker_compose_runnable = DockerComposeUtils.dockerCompose(compose_args, { stdio: 'inherit', env: { COMPOSE_IGNORE_ORPHANS: 'true' } });
+    try {
+      await DockerComposeUtils.watchContainersHealth(compose_file, project_name);
+    } catch (err) {
+      docker_compose_runnable.kill();
+      throw err;
+    }
+    await docker_compose_runnable;
     fs.removeSync(compose_file);
   }
 
--- a/src/commands/dev.ts
+++ b/src/commands/dev.ts
@@ -208,11 +208,7 @@ export default class Dev extends BaseCommand {
     }
 
     const docker_compose_runnable = DockerComposeUtils.dockerCompose(compose_args, { stdio: 'inherit', env: { COMPOSE_IGNORE_ORPHANS: 'true' } });
-    try {
-      DockerComposeUtils.watchContainersHealth(compose_file, project_name);
-    } catch (err) {
-      console.error(err);
-    }
+    DockerComposeUtils.watchContainersHealth(compose_file, project_name);
     await docker_compose_runnable;
     fs.removeSync(compose_file);
   }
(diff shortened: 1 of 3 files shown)


ryan-cahill (commented):
Does this also take into account services that can take a long time to become ready on their first run?

mueschm:
@ryan-cahill yes it will. So when a container states with a health check it's health will be set to starting. Once the health check has passed the first time it moves into healthy. This PR will only restart a container if it is deemed unhealthy or has exited. So a long startup time is fine.

ryan-cahill (approved):

src/common/docker-compose/index.ts
```diff
@@ -444,4 +444,72 @@ export class DockerComposeUtils {
...
+
+          service_data.restarts++;
+          service_data.last_restart_ms = Date.now();
+          if (service_data.restarts > max_restarts) {
+            throw new Error(`Unable to start service ${service_name}`);
+          }
+          console.log(chalk.red(`ERROR: Service ${service_name} has encountered an error and is being restarted.`));
+          console.log(chalk.red(`Retry attempt ${service_data.restarts} of ${max_restarts}`));
+          await restart(id);
+          DockerComposeUtils.dockerCompose(['-f', compose_file, '-p', environment_name, 'logs', full_service_name, '--follow', '--since', new Date(service_data.last_restart_ms).toISOString()], { stdout: 'inherit' });
```
tjhiggins, line 522: image

This seems to be creating multiple log streams for a service when it fails.

tjhiggins (changes requested):

src/common/docker-compose/index.ts
tjhiggins: I don't think this is necessary. I removed the line and my logs still show for the restarted containers.

src/commands/dev.ts
```diff
@@ -207,7 +207,14 @@ export default class Dev extends BaseCommand {
       compose_args.push('-d');
     }
 
-    await DockerComposeUtils.dockerCompose(compose_args, { stdio: 'inherit', env: { COMPOSE_IGNORE_ORPHANS: 'true' } });
+    const docker_compose_runnable = DockerComposeUtils.dockerCompose(compose_args, { stdio: 'inherit', env: { COMPOSE_IGNORE_ORPHANS: 'true' } });
+    try {
+      await DockerComposeUtils.watchContainersHealth(compose_file, project_name);
```
tjhiggins, line 212: Should this be an await? I got in a bad state where I couldn't ctrl-c because of the while loop (I think). Had to manually kill the processes.

tjhiggins: Nevermind that might be the kill call I added. Was having a separate issue where the docker-compose process never ended if watchContainersHealth threw an error.

src/common/docker-compose/index.ts
tjhiggins: Yeah I removed and still got the api logs
image

mueschm: Interesting for me the streams would be gone. Need to deep dive. Can I get your docker/kubernetes version number just to verify?

src/commands/dev.ts
mueschm: It actually does not have to be so will remove this.

src/common/docker-compose/index.ts
tjhiggins:
```
Client: Docker Engine - Community
 Cloud integration: v1.0.22
 Version:           20.10.12
 API version:       1.41
 Go version:        go1.16.12
 Git commit:        e91ed57
 Built:             Mon Dec 13 11:45:48 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.12
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.12
  Git commit:       459d0df
  Built:            Mon Dec 13 11:43:56 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.12
  GitCommit:        7b11cfaabd73bb80907dd23182b9347b4245eb5d
 runc:
  Version:          1.0.2
  GitCommit:        v1.0.2-0-g52b36a2
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

Zach-at-Arch (approved):
Looks good! I tried spinning up with 5 instances of the health app and it would keep healing itself every time an app died. Also played around with different time intervals for both the health probe and how quickly the app crashed. Overall, looks good!

PR #553, opened 2022-04-10
Updated workflows to fix bugs.

mueschm:
## Fixes
1) `npm run pack` is now handled by the release config. Only being called once now.
2) Added waits on the publish command for the lint and test command.
3) Only main will post to Github. This was due to a configuration issue.


.github/workflows/publish.yml
```diff
@@ -18,14 +18,27 @@ jobs:
     name: Publish CLI
     runs-on: ubuntu-latest
     steps:
+      - name: Wait for tests to succeed
+        uses: lewagon/wait-on-check-action@v1.0.0
```
tjhiggins, line 22: I'd still prefer to have this all in one workflow. Did you try using a conditional on the job to check the branch name?

```yaml
on: push

jobs:
  lint:
    ...
  test: 
    ...
  publish_cli:
    if: github.ref_name == 'main' || github.ref_name == 'rc' || startsWith(github.ref_name, 'arc-')
    needs: [lint, test]
```

tjhiggins (changes requested):

.github/workflows/publish.yml
mueschm: I did and that would work, I just really like the branch requirements up to so was trying to hold on to that. Will change it to this.

PR #556, opened 2022-04-12
Multiple secret files

Zach-at-Arch:
Extend Deploy.utils to allow multiple secret-files to be ingested

The last file provided will always upsert any existing secrets, and any secrets passed via CLI will upsert all secret-files in the '*' key.

**Testing:**
With two secret files, you can run:
$ architect deploy --secret-file secrets1.yml --secret-file secrets2.yml
OR
$ architect deploy --secret-file secrets1.yml secrets2.yml
OR
$ architect deploy --secret-file secrets1.yml --secret-file secrets2.yml -s secret_key=secret_value

### DIFF
--- a/src/commands/dev.ts
+++ b/src/commands/dev.ts
@@ -216,16 +216,19 @@ export default class Dev extends BaseCommand {
   }
 
   private async runLocal() {
-    const { args, flags } = await this.parse(Dev);
+    const { args, flags, raw } = await this.parse(Dev);
     await Docker.verify();
 
     if (!args.configs_or_components || !args.configs_or_components.length) {
       args.configs_or_components = ['./architect.yml'];
     }
 
+    // Ensure we capture all secret-files provided
+    const all_secret_files = DeployUtils.getAllSecretFiles(raw);
+
     const interfaces_map = DeployUtils.getInterfacesMap(flags.interface);
-    const component_secrets = DeployUtils.getComponentSecrets(flags.secret, flags['secret-file']);
-    const component_parameters = DeployUtils.getComponentSecrets(flags.parameter, flags['secret-file']);
+    const component_secrets = DeployUtils.getComponentSecrets(flags.secret, all_secret_files);
+    const component_parameters = DeployUtils.getComponentSecrets(flags.parameter, all_secret_files);
 
     const linked_components = this.app.linkedComponents;
     const component_versions: string[] = [];
--- a/src/commands/dev.ts
+++ b/src/commands/dev.ts
@@ -47,6 +47,8 @@ export default class Dev extends BaseCommand {
     }),
     'secret-file': Flags.string({
       description: 'Path of secrets file',
+      multiple: true,
+      default: [],
     }),
     secret: Flags.string({
       char: 's',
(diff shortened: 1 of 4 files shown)


src/commands/dev.ts
```diff
@@ -216,16 +216,19 @@ export default class Dev extends BaseCommand {
   }
 
   private async runLocal() {
-    const { args, flags } = await this.parse(Dev);
+    const { args, flags, raw } = await this.parse(Dev);
```
mueschm, line 219: I do not think we need to use the raw data here. Flags can be set with the multiple option to true.
https://oclif.io/docs/flags

```
    'secrets-file': Flags.string({
      description: 'Path to a secrets file',
      multiple: true,
      default: [],
    }),
```

Then if you went to access the flag you can just do. `flags['secrets-file']` to get a string array. This will also make sure the flag will appear in our readme.

src/common/utils/deploy.utils.ts
```diff
@@ -58,13 +59,36 @@ export default class DeployUtils {
...
+      }
+    }
+    return all_secret_files;
+  }
+
+  static getComponentSecrets(individual_secrets: string[], secrets_file?: string | string[]): Dictionary<Dictionary<string | number | null>> {
+    // Check to see if there are multiple secret files; else, just read the single secret file
+    let component_secrets: any = {};
+    if (Array.isArray(secrets_file)) {
+      for (const secret_file of secrets_file) {
```
mueschm, line 76: With the changes above this should become an array by default and the check unneeded. We should just assume that secrets file is an array no matter the size and work with it that way.

src/common/utils/deploy.utils.ts
```diff
@@ -58,13 +59,36 @@ export default class DeployUtils {
     return flags;
   }
 
-  static getComponentSecrets(individual_secrets: string[], secrets_file?: string): Dictionary<Dictionary<string | number | null>> {
-    const component_secrets = DeployUtils.readSecretsFile(secrets_file);
+  static getAllSecretFiles(raw_argv: any): string[] {
```
mueschm, line 62: This function would probably just become.
```
const secret_files = flags.values.concat(flags['secret-files']);
```

mueschm (changes requested):
Idea is definitely right, but we need to make to lean into the oclif standards since we rely on them for core features like Readme generation.

src/common/utils/deploy.utils.ts
Zach-at-Arch: Ended up removing since we can pass multiple secret files in

src/common/utils/deploy.utils.ts
```diff
@@ -58,13 +59,22 @@ export default class DeployUtils {
     return flags;
   }
 
-  static getComponentSecrets(individual_secrets: string[], secrets_file?: string): Dictionary<Dictionary<string | number | null>> {
-    const component_secrets = DeployUtils.readSecretsFile(secrets_file);
+  static getComponentSecrets(individual_secrets: string[], secrets_file: string | string[]): Dictionary<Dictionary<string | number | null>> {
+    // Check to see if there are multiple secret files; else, just read the single secret file
+    let component_secrets: any = {};
+    for (const secret_file of secrets_file) {
```
mueschm, line 65: `secrets_file` can be both a string or string array. This code would work fine for a string array but not a string. Can we make secrets_files only an array here?

mueschm (changes requested):
Please also add the test back in.

PR #558, opened 2022-04-14
Remove orphaned containers on startup.

mueschm:
## Overview
Make sure that when `docker compose up` is run that orphaned containers are destroyed.

## Changes
1) Remove the env that ignored orphans
2) Add the parameter that removes orphans

## Test
Used my health test from https://github.com/architect-team/architect-cli/pull/550

Added a service called `health2` and ran that. Stopped it then removed the service from the yml file. Started it again and only one instance was running.


Zach-at-Arch:
Confirmed the same - also tried running with different number of instances and removing and would clean each time.

Zach-at-Arch (approved):

PR #561, opened 2022-04-22
Hello world dockerfile

ryan-cahill:


examples/hello-world/architect.yml
```diff
@@ -9,7 +9,8 @@ keywords:
 
 services:
   api:
-    image: heroku/nodejs-hello-world
```
tjhiggins, line 12: Looks like there are other places we use this image in the cli. Search `heroku/nodejs-hello-world`

tjhiggins (changes requested):

PR #562, opened 2022-04-24
Better support exiting dev

mueschm:
## Overview
When a user hits ctrl-c multiple times it prevents `docker compose` from being able to clean up after itself. This will leave all the containers still running and just exit the `dev` command. This intercepts the exit signal and makes sure we can gracefully shutdown.

Originaly the plan was also to update `execa` to the latest version. We could have use dynamic imports to pull it in even though its not directly compatible with CommonJS. However I ran into a problem where we used the `Options` type from the import quite a bit. Since dynamic imports require an `await` there is no clean way to continually use this type without making a lot of code async when it does not need to be. The only other options were.
1) Create an abstraction around execa. We do not get really anything from this.
2) Copy all the types into our project so we can include `execa` async but have the types sync. This seemed like a really hacky work around.

Since the current version meets our needs I saw no reason to pursue either of these not great options.

## Changes
1) Intercept the `ctrl-c` command and stop the application from exiting.
2) When `ctrl-c` is pressed send the signal to shutdown `docker compose`
3) Fixed an issue where the healthcheck was now causing issues because containers had exited status even though they were left over from a previous run.
4) `docker compose up` will display a message telling users that they can hit `ctrl-c` again to force exit. We no longer allow this so I intercepted the text stream and override the message. This did lose us color support so had to hack something back in.

## Testing
You can run the react-app to see it working.
```
cd examples/react-app
architect-local dev .
````
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,6 +38,7 @@
         "matcher": "^4.0.0",
         "opener": "^1.5.2",
         "p-limit": "^3.1.0",
+        "readline": "^1.3.0",
         "reflect-metadata": "0.1.13",
         "semver": "^7.3.5",
         "shell-quote": "^1.7.2",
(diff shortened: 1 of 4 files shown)


Zach-at-Arch (approved):
All changes look good!

Tried out the new feature with both the health application and another app. Both worked well, with expected output when hitting `ctrl-c`.

Screen Shot 2022-04-25 at 8 37 27 AM
 
Additionally, I tried running one of the apps and hit `ctrl-c` multiple times, and the application still successfully closed all running instances.

PR #564, opened 2022-04-26
Adding reserved names

ryan-cahill:


tjhiggins:
Talked with Ryan and we are going to get this ready for release without supporting interpolation for the reserved name field.

src/dependency-manager/manager.ts
```diff
@@ -44,7 +44,7 @@ export default abstract class DependencyManager {
     // Load component tasks
     for (const [task_name, task_config] of Object.entries(component.tasks)) {
       const node = new TaskNode({
-        ref: buildNodeRef(component, 'tasks', task_name),
+        ref: task_config.reserved_name || buildNodeRef(component, 'tasks', task_name),
```
tjhiggins, line 47: Couldn't we move this logic into `buildNodeRef`? 

something like:
`component_config[resource_type][resource_name].reserved_name`

src/dependency-manager/spec/utils/slugs.ts
```diff
@@ -34,6 +34,10 @@ export class Slugs {
   public static ComponentSecretDescription = 'must contain alphanumeric character ([a-z0-9A-Z]), could contain dashes (-), underscores (_), and alphanumerics between.';
   public static ComponentSecretRegexBase = `[a-zA-Z0-9_-]+`;
   public static ComponentSecretValidator = new RegExp(`^${Slugs.ComponentSecretRegexBase}$`);
+
+  public static ReservedNameDescription = `must contain only alphanumeric characters ([a-z0-9A-Z]), could contain dashes (-), underscores (_), and alphanumerics between; max length ${Slugs.SLUG_CHAR_LIMIT}`;
```
tjhiggins, line 38: I don't think underscores are allowed for k8s resource names. Hadn't considered this case though...

I also don't want to allow bypassing the regex enforcement for service names from ArchitectSlugValidator.

tjhiggins (changes requested):

PR #565, opened 2022-04-26
temporarily setting secrets back to parameters

ryan-cahill:


mueschm:
fix(examples): Examples now use parameters instead of secrets

PR #566, opened 2022-04-26
Register command with M1 chip

mtgoncalves1:


src/commands/register.ts
```diff
@@ -118,40 +127,61 @@ export default class ComponentRegister extends Command {
     const project_name = `register.${resourceRefToNodeRef(component_spec.name)}.${tag}`;
     const compose_file = DockerComposeUtils.buildComposeFilepath(this.app.config.getConfigDir(), project_name);
 
-    await DockerComposeUtils.writeCompose(compose_file, yaml.dump(compose));
+    await DockerBuildXUtils.writeCompose(compose_file, yaml.dump(compose));
```
mueschm, line 130: This can still be `DockerComposeUtils` since it is a compose file. `BuildX` just happens to support compose file format. This also means we don't duplicate functionality.

src/commands/register.ts
```diff
@@ -118,40 +127,61 @@ export default class ComponentRegister extends Command {
...
-      arr.push(value);
+      arr.push("--set");
+      arr.push(`*.args.${value}`);
       return arr;
     }, [] as string[]);
 
     try {
-      await DockerComposeUtils.dockerCompose(['-f', compose_file, 'build', ...build_args], {
-        stdio: 'inherit',
+      const is_local = this.app.config.api_host.includes("localhost");
```
mueschm, line 148: Lets move all of this logic into the BuildX utils. You should just have something like.
```
await DockerBuildXUtils.createBuilder();
```

src/commands/register.ts
```diff
@@ -118,40 +127,61 @@ export default class ComponentRegister extends Command {
...
+    }
+
+    try {
+      await DockerBuildXUtils.dockerBuildX(["inspect", "--bootstrap"], {
+        stdio: "inherit",
       });
     } catch (err: any) {
       fs.removeSync(compose_file);
-      this.log(`Docker build failed. If an image is not specified in your component spec, then a Dockerfile must be present`);
+      this.log(`Docker buildx failed to inspect`);
```
mueschm, line 173: Change this too
```
Unable to start buildx builder. Please make sure docker is running.
```
Inspect may be a confusing word if people don't have context into buildx.

mueschm:
Please fix all errors as well as address comments. Though overall well done. The implementation itself is clean and all the comments I made are just some touch-ups. Nice work!

mueschm (changes requested):

mueschm:
This is looking good. Tested locally and all the fixes seem to be working now. However I still want 2 additional engineers to sign off given this is a major change to the system.

PR #573, opened 2022-05-09
Changed order of plugins to fix git happening too soon.

mueschm:
## Overview
After looking at the logs it looks like the git command is happening before some of the other commands. Since most commands occur at the same step and they seem to match the order they were put in the config file, I am changing that order. This should fix it.

## Changes
1. Update order of the plugins for semantic release.
2. Log the order to make it easier to debug in the future.

## Testing
`node release.config.js`
```
{
  branches: [
    'main',
    { name: 'rc', prerelease: true },
    { name: 'arc-*', prerelease: true }
  ],
  plugins: [
    '@semantic-release/commit-analyzer',
    '@semantic-release/release-notes-generator',
    [ '@semantic-release/exec', [Object] ],
    '@semantic-release/npm',
    [ '@semantic-release/git', [Object] ]
  ]
}
```

`GIT_BRANCH=main node release.config.js`
```
{
  branches: [
    'main',
    { name: 'rc', prerelease: true },
    { name: 'arc-*', prerelease: true }
  ],
  plugins: [
    '@semantic-release/commit-analyzer',
    '@semantic-release/release-notes-generator',
    [ '@semantic-release/changelog', [Object] ],
    [ '@semantic-release/exec', [Object] ],
    '@semantic-release/npm',
    [ '@semantic-release/git', [Object] ],
    [ '@semantic-release/github', [Object] ]
  ]
}
```

### DIFF
--- a/release.config.js
+++ b/release.config.js
@@ -1,31 +1,28 @@
 const { GIT_BRANCH: branch } = process.env;
 
-const plugins = [
-  "@semantic-release/commit-analyzer",
-  "@semantic-release/release-notes-generator",
-  [
-    "@semantic-release/git",
-    {
-      "assets": [
-        "CHANGELOG.md",
-        "README.md",
-        "package.json",
-        "package-lock.json",
-        "docs/5-reference/2-architect-yml.md",
-        "src/dependency-manager/schema/architect.schema.json",
-      ]
-    }
-  ],
-  [
-    "@semantic-release/exec",
-    {
-      "publishCmd": "npm run pack"
-    }
-  ],
-  "@semantic-release/npm"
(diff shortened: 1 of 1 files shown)


release.config.js
```diff
@@ -54,5 +71,8 @@ module.exports = {
       "prerelease": true
     }
   ],
-  plugins: branch === 'main' ? main_plugins : plugins,
+  plugins: branch === 'main' ? main_plugins : default_plugins,
 };
+
+
+console.log(module.exports);
```
tjhiggins, line 78: Did you want this console.log?

mueschm: Yeah it should just now print out in the Action which will make it easier to tell exactly what configuration options are being passed into semantic release.

PR #574, opened 2022-05-10
add user-agent default header to AppService._api

devinsag:


src/app-config/service.ts
```diff
@@ -38,7 +39,7 @@ export default class AppService {
       baseURL: this.config.api_host,
       timeout: 10000,
       headers: {
-        'Cli-Version': this.version,
```
mueschm, line 41: Lets not remove this and instead add to it.

devinsag: Changed to append a new User-Agent header instead of replacing the existing 'Cli-Version' header.

PR #577, opened 2022-05-17
Whitelist interpolation

ryan-cahill:


test/dependency-manager/interpolation.test.ts
```diff
@@ -1320,17 +1320,20 @@ describe('interpolation spec v1', () => {
...
         ingress:
-          ip_whitelist: \${{ secrets.ip_whitelist }}
+          ip_whitelist:
+            - \${{ secrets.ip_whitelist }}
       api2:
         url: \${{ services.api.interfaces.main.url }}
         ingress:
-          ip_whitelist: \${{ secrets.required_ip_whitelist }}
+          ip_whitelist:
+            - \${{ secrets.required_ip_whitelist }}
```
tjhiggins, line 1336: Shouldn't we also test for the list case as well?

`ip_whitelist: \${{ secrets.required_ip_whitelist }}`

devinsag (approved):

test/dependency-manager/interpolation.test.ts
tjhiggins: I get that the array doesn't work api side, but why would we remove the test of it working in the cli?

ryan-cahill: I just added a new one instead of updating the existing one. At first I thought it would be weird to include a test for a scenario that wouldn't work when used remotely

PR #578, opened 2022-05-17
Revert "Register command with M1 chip"

ryan-cahill:
Reverts architect-team/architect-cli#566


src/commands/register.ts
```diff
@@ -128,33 +120,40 @@ export default class ComponentRegister extends Command {
 
     await DockerComposeUtils.writeCompose(compose_file, yaml.dump(compose));
 
-    let build_args: string[] = flags.arg || [];
+    let build_args: string[] = [];
```
mueschm, line 123: Can we not revert this change as I think this is still necessary and was actually a bug from before.

tjhiggins: I don't know if that was the right bug fix since we set the build args below too. 

```ts
private async getBuildArgs(resource_spec: ResourceSpec): Promise<string[]> {
    const { flags } = await this.parse(ComponentRegister);

    const build_args_map: Dictionary<string | null> = { ...resource_spec.build?.args };
    for (const arg of flags.arg || []) {
      const [key, value] = arg.split(/=([^]+)/);
      if (!value) {
        throw new Error(`--arg must be in the format key=value: ${arg}`);
      }
      build_args_map[key] = value;
    }
    return Object.entries(build_args_map).map(([key, value]) => `${key}=${value}`);
  }
  ```

tjhiggins: Although I'm not sure that logic makes any sense either. Looks like we are setting the same build args potentially multiple times in that for loop.

PR #580, opened 2022-05-17
Register interpolation validation

tjhiggins:


ryan-cahill:
Before, we were also setting build args based on both flags and some that might be explicitly set in a service config. Why don't we use the ones set in a service config anymore? Does bake work differently?

test/dependency-manager/interpolation-validation.test.ts
```diff
@@ -0,0 +1,145 @@
...
+        `
+
+      const component_spec = buildSpecFromYml(component_config)
+
+      expect(() => {
+        registerInterpolation(component_spec, context)
+      }).to.be.throws(ValidationErrors);
+    });
+
+    it('cannot use conditional around service block with build block', async () => {
```
mueschm, line 68: If you have a local only service why could you not do this? This may be a  more advanced feature but I see no reason you could not do.
```
${{ if architect.build.args.something == 'local' }}:
```
Would also like to see some tests that include the args.

test/dependency-manager/interpolation-validation.test.ts
```diff
@@ -0,0 +1,145 @@
...
+        `
+
+      const component_spec = buildSpecFromYml(component_config)
+
+      expect(() => {
+        registerInterpolation(component_spec, context)
+      }).to.be.throws(ValidationErrors);
+    });
+
+    it('cannot use conditional around build block', async () => {
```
ryan-cahill, line 50: This seems like it would be problematic in a lot of cases as it was a replacement for the debug block. We use conditionals around `build` blocks for some of our own services as it is so I feel like it's really likely that this is something other users would need as well

test/dependency-manager/interpolation-validation.test.ts
tjhiggins: You'd be able to use conditionals anywhere with `architect.build.tag` and `architect.build.args`. The breaking change would be preventing non-build time interpolation like environment. 

Here is an example of a valid conditional in the build block:
https://github.com/architect-team/architect-cli/pull/580/files#diff-ccf91dc9ee7a6fcf782de9f37462393ebf1f28f1a01e586ecb6e3bbd97e1034bR93

test/dependency-manager/interpolation-validation.test.ts
tjhiggins: The thought is that you could still use conditionals anywhere, but not around the build block if they are deploy-time specific.

So the following wouldn't be allowed:
`${{ if architect.environment == 'local' }}:`

However, this would:
`${{ if architect.build.tag == 'local' }}:`
See: https://github.com/architect-team/architect-cli/pull/580/files#diff-ccf91dc9ee7a6fcf782de9f37462393ebf1f28f1a01e586ecb6e3bbd97e1034bR93

----

Still trying to decide if we make a special case for `architect.environment == 'local'` for some period of time.

src/dependency-manager/utils/rules.ts
```diff
@@ -0,0 +1,93 @@
...
+];
+
+export function checkRules(context_map: ContextMap, context_key: string): ValidationError | undefined {
+  if (!context_map._path) {
+    return;
+  }
+  for (const rule of rules) {
+    const maybe_error = rule.run(context_map, context_key);
+    if (maybe_error) {
+      return maybe_error;
```
ryan-cahill, line 90: What's with the `maybe` things in these new changes? Is this only an error in some cases and in other cases it can be ignored? Using `maybe` for these variables makes me question what's going on in general

test/dependency-manager/interpolation-validation.test.ts
```diff
@@ -0,0 +1,196 @@
...
+            \${{ if architect.environment == 'local' }}:
+              environment:
+                TEST: test
+        `
+
+      const component_spec = buildSpecFromYml(component_config)
+      registerInterpolation(component_spec, context)
+    });
+  });
+});
```
ryan-cahill, line 196: I think these test cases make sense, though there's one common pattern that we use ourselves that I don't see here and will be a breaking change to our own stack's config. For example, you can't do something like 

```yml
a-service:
    ...
    build:
      ...
      args:
        NODE_ENV: ${{ services.a-service.environment.NODE_ENV }}
    environment:
      NODE_ENV: production
    ${{ if secrets.environment == 'local' }}:
      environment:
        NODE_ENV: development
```

The error is

```
ValidationErrors
component: cloud
file: architect.yml:153:23
  150 |     build:
  151 |       ...
  152 |       args:
› 153 |         NODE_ENV: ${{ services.a-service.environment.NODE_ENV }}
      |                       ﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋ Cannot use ${{ services.a-service.environment.NODE_ENV }} inside of a build block. Use ${{ if architect.build.tag == 'latest' }}: for build time conditionals.
```

I'd have expected this interpolation to work since it doesn't rely on a secret. Is that a bug or intentional?

ryan-cahill (changes requested):

test/dependency-manager/interpolation-validation.test.ts
tjhiggins: Good call - let me think about this

src/dependency-manager/utils/parser.ts
```diff
@@ -236,9 +229,11 @@ export class ArchitectParser {
 
     let last_value;
     for (const match of matches(program, EXPRESSION_REGEX)) {
-      const ast = this.parseExpression(match[1], context_map, ignore_keys, _depth);
-      res = res.replace(match[0], ast.body[0].value);
-      last_value = ast.body[0].value;
+      const ast = this.parseExpression(match[1], context_map, _depth);
```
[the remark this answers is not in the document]
tjhiggins: abstract syntax tree

ryan-cahill (approved):

PR #585, opened 2022-05-18
427-remove deploy spec and add test for cli validation

devinsag:
Removed DeploySpec, DeployModuleSpec and its references in src and test.  Added a new test-case verifying use of deploy spec causes a validation failure in test/dependency-manager/validation.test


src/dependency-manager/spec/resource-spec.ts
```diff
@@ -1,59 +1,9 @@
-import { IsObject, IsOptional, IsString, ValidateNested } from 'class-validator';
+import { IsOptional, ValidateNested } from 'class-validator';
```
tjhiggins, line 1: Remove the changes to:
- docs/5-reference/2-architect-yml.md
- src/dependency-manager/schema/architect.schema.json

These files are auto-generated.

tjhiggins (changes requested):

src/dependency-manager/spec/resource-spec.ts
devinsag: reverted changes to the files:
* docs/5-reference/2-architect-yml.md
* src/dependency-manager/schema/architect.schema.json

PR #587, opened 2022-05-24
Deprecate path and port

ryan-cahill:


tjhiggins (changes requested):

docs/1-components/0-architect-yml.md
```diff
@@ -47,8 +47,7 @@ services:
         protocol: http
       admin: 8081
     liveness_probe:
-      port: 8080
-      path: /health
+      command: curl localhost:8080/health
```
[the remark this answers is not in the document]
ryan-cahill: Updated those examples and tested the react app again. Also found/fixed a bug with shell parsing that I'll add a test for

src/common/docker-compose/index.ts
```diff
@@ -134,7 +134,7 @@ export class DockerComposeUtils {
         const liveness_probe = node.config.liveness_probe;
         if (liveness_probe) {
           if (!liveness_probe.command) {
-            liveness_probe.command = ['CMD-SHELL', `curl -f http://localhost:${liveness_probe.port}${liveness_probe.path} || exit 1`];
+            liveness_probe.command = ['CMD-SHELL', `curl -f http://localhost:${liveness_probe.port}${liveness_probe.path} || exit 1`]; // deprecated
           } else {
             liveness_probe.command = ['CMD-SHELL', liveness_probe.command.join(' ')];
```
tjhiggins, line 139: Probably a question for the team. Should we be using `CMD-SHELL` or making them explicitly put `sh -c ...`

tjhiggins (changes requested):

src/common/docker-compose/index.ts
ryan-cahill: What's the downside of the current approach? I haven't heard a user complain about it

tjhiggins: I imagine we would have to make an assumption on k8s of what shell to use. /bin/sh isn't guaranteed to exist I don't think.

ryan-cahill: Fair point. Want me to include that here or add a ticket?

tjhiggins: Add a ticket and put in next sprint with triage

devinsag: Can the selection of shell be dynamic at runtime? something like..

```shell
if [ -e "$SHELL" ]
then
  CMD ["$SHELL", `curl -f http://localhost:${liveness_probe.port}${liveness_probe.path} || exit 1`]
else
  echo "WARNING: k8 must have shell to run liveness check" >&2
  exit 1
fi
```

ryan-cahill: I think that's a cool idea but I'm wary of edge cases where this kind of magic could cause a headache for users

mueschm: After looking into this I think it makes the most sense to keep using `CMD-SHELL`.  The only time this becomes a problem is if there is no shell and `SHELL` is not set by the docker container which looks to be incredibly rare.

`CMD` does not allow features like piping and behaves differently than most of our users would be used to. Unless someone explicitly needs to override this to use a kernel exec then I would vote we don't change it.

Honestly, I am more worried that `curl` or `wget` may not exist on an image and we need to have a mechanism to walk users through getting everything working.

PR #588, opened 2022-05-27
Ensure that image names are preserved when a reserved name is used

ryan-cahill:
The purpose of this PR is to restore the image name generation in the register command. Adding `reserved_name` changes the `ref` of a node to just that reserved name and that's it.Without a reserved name though, the ref would contain the component name, whether the node was a service or task node, and the name of the service or task that it represents. This information is required when creating an image to push to the registry

### DIFF
--- a/src/dependency-manager/graph/node/service.ts
+++ b/src/dependency-manager/graph/node/service.ts
@@ -1,5 +1,6 @@
 import { DependencyNode, DependencyNodeOptions } from '.';
 import { ServiceConfig, ServiceInterfaceConfig } from '../../config/service-config';
+import { Refs } from '../../utils/refs';
 
 export interface ServiceNodeOptions {
   ref: string;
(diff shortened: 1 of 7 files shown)


src/dependency-manager/graph/node/service.ts
```diff
@@ -43,4 +43,13 @@ export class ServiceNode extends DependencyNode implements ServiceNodeOptions {
   get is_local(): boolean {
     return this.local_path !== '';
   }
+
+  get architect_ref(): string {
```
mueschm, line 47: Can we abstract this logic into a util function?

mueschm:
Can you add a description to this PR?

mueschm (changes requested):

mtgoncalves1:
I got the following error when I registered with `reserved_name: my-reserved-api`.
register error with reserved_name keyword

src/dependency-manager/graph/node/service.ts
ryan-cahill: Updated by adding a function to `src/dependency-manager/utils/refs.ts`

ryan-cahill: Updated by adding a function to `src/dependency-manager/utils/refs.ts`

PR #590, opened 2022-06-07
(Sentry 1/3) 440 send exceptions to sentry

devinsag:


src/base-command.ts
```diff
@@ -92,11 +99,98 @@ export default abstract class BaseCommand extends Command {
     return super.parse(options, [...args, ...flags]);
   }
 
+  async _getFilteredMetadata(calling_class: any): Promise<any> {
```
ryan-cahill, line 102: Maybe `_getFilteredSentryMetadata` so we know it's just for Sentry

src/base-command.ts
```diff
@@ -92,11 +99,98 @@ export default abstract class BaseCommand extends Command {
...
+  }
+
+  async _logToSentry(err: any): Promise<void> {
+    const auth_result = await this.app.auth.getPersistedTokenJSON();
+    const auth_user = await this.app.auth.checkLogin();
+    const metadata = await this._getFilteredMetadata(this.constructor as any);
+
+    Sentry.init({
+      dsn: CLI_SENTRY_DSN,
+      debug: false,
```
ryan-cahill, line 134: Should this be set to true/false based on the config or something?

src/base-command.ts
```diff
@@ -92,11 +99,98 @@ export default abstract class BaseCommand extends Command {
...
+          id: auth_user.id,
+        },
+        extra: {
+          ...this.config,
+          ...this.app.config.toSentry(),
+          command_args: metadata.args,
+          command_flags: metadata.flags,
+          config_file: path.join(this.app.config.getConfigDir(), LocalPaths.CLI_CONFIG_FILENAME),
+          cwd: process.cwd(),
+          docker_info: execSync('docker version').toString(),
```
ryan-cahill, line 164: Might want to just use the `docker` util here instead

src/base-command.ts
```diff
@@ -92,11 +99,98 @@ export default abstract class BaseCommand extends Command {
...
+          linked_components: this.app.linkedComponents,
+          log_level: this.app.config.log_level,
+          node_version: process.version,
+          os_release: os.release(),
+          os_type: os.type(),
+        },
+        tags: {
+          cli: this.app.version,
+          node_runtime: process.version,
+          os: os.platform(),
```
ryan-cahill, line 174: Nice, looks like it would be pretty helpful for M1 issues

src/commands/destroy.ts
```diff
@@ -25,29 +25,43 @@ export default class Destroy extends DeployCommand {
     }),
   };
 
+  static sensitive = new Set();
+
+  static non_sensitive = new Set([...Object.keys({ ...this.flags })]);
+
   async run(): Promise<void> {
-    const { flags } = await this.parse(Destroy);
+    try {
```
ryan-cahill, line 33: Is there a way to log to Sentry without wrapping each function in a try/catch block? Maybe by adding something in the base class?

ryan-cahill (changes requested):

mtgoncalves1 (changes requested):

src/commands/destroy.ts
devinsag: I know it's not very dry, so I'm writing a class and method decorator to get the proper context/stacktrace without the try catch blocks.  This ideally will be set on the BaseCommand class only, and once I can confirm it works the same i'll remove all of the added try/catch in this PR.

src/base-command.ts
```diff
@@ -92,11 +99,98 @@ export default abstract class BaseCommand extends Command {
...
+
+    const _flags = Object.entries(flags).map(([key, value]) => {
+      if (calling_class.non_sensitive.has(key)) {
+        return { [key]: value };
+      } else if (calling_class.sensitive.has(key)) {
+        return { '[Filtered]': '**********' };
+      } else {
+        return { '[Unknown]': '**********' };
+      }
+    });
```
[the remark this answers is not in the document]
devinsag: great idea! thanks for pointing this out.

src/base-command.ts
devinsag: refactored using docker

src/commands/destroy.ts
devinsag: refactored as a class decorator

src/base-command.ts
devinsag: updated function name

src/base-command.ts
```diff
@@ -92,11 +99,88 @@ export default abstract class BaseCommand extends Command {
     return super.parse(options, [...args, ...flags]);
   }
 
+  async _filterSensitiveSentryMetadata(sentry_metadata: any): Promise<any> {
+    const context = this.constructor as any;
+    return Object.entries(sentry_metadata).map(([key, value]) => {
+      if (context.non_sensitive.has(key)) {
+        return { [key]: value };
+      } else if (sentry_metadata.sensitive.has(key)) {
```
mueschm, line 107: We do not need both lists. Either something is sensitive or its not. We should start with the assumption everything is sensitive. Then if its in the non-sensitive list we are willing to report on it.

src/commands/components/versions.ts
```diff
@@ -22,6 +29,11 @@ export default class ComponentVersions extends Command {
     name: 'component_name',
   }];
 
+  static sensitive = new Set();
+
+  static non_sensitive = new Set([...Object.keys({ ...ComponentVersions.flags }),
```
mueschm, line 34: We could make a util function in BaseCommand that would handle this. Having to do this for each command would be hard to maintain.

src/sentry.ts
```diff
@@ -0,0 +1,48 @@
...
+      const property_names = Reflect.ownKeys(target.prototype).filter(prop => prop !== 'constructor');
+      for (const property_name of property_names) {
+        const descriptor = Object.getOwnPropertyDescriptor(target.prototype, property_name);
+        const is_method = descriptor?.value instanceof Function;
+        if (!is_method) {
+          continue;
+        }
+        Object.defineProperty(target.prototype, property_name, _generateDescriptor(descriptor, error_type, handler));
+      }
+    }
```
Zach-at-Arch, line 46: bit confused about this `if` statement. I see that if a descriptor is present, it returns `_generateDescriptor`, but if not it is just defining the property, but not returning anything. Is this intended?

mueschm (changes requested):
Need to still do a depper dive, but given the core changes needed that will wait until the next revision.

src/commands/components/index.ts
```diff
@@ -28,6 +35,11 @@ export default class Components extends Command {
     description: 'Search term used to filter the results',
   }];
 
+  static sensitive = new Set();
```
Zach-at-Arch, line 38: For empty sets, just put a comment to let others know this is a placeholder

src/sentry.ts
devinsag: The decorator can be used on classes, or individual methods, so (f descriptor) means it's a method decorator and is applied to that method; essentially wrapping only that method with a try/catch.

Otherwise it's a class decorator so it is applied for every method (besides the constructor); essentially wrapping every method in the class with a try/catch dynamically.

src/commands/config/get.ts
```diff
@@ -1,8 +1,16 @@
...
 import Command from '../../base-command';
 import InvalidConfigOption from '../../common/errors/invalid-config-option';
+import { ToSentry } from '../../sentry';
 
+@ToSentry(Error,
+  (err, ctx) => {
+    const error = err as any;
+    error.stack = Error(ctx.id).stack;
+    return error;
+})
```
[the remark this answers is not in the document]
devinsag: I was seeing unexpected behavior when added to the BaseCommand combined with oclif running as binary. I can do more research on the topic.

src/commands/components/versions.ts
ryan-cahill: Yeah ideally most things here would be on the base command class

devinsag: Removed sensitive/non_sensitive sets, and added flag/arg property 'non_sensitive': boolean instead, which will be used to filter flag/arg metadata sent to Sentry from within the BaseCommand class.

src/base-command.ts
devinsag: Removed sensitive/non_sensitive sets, and added flag/arg property 'non_sensitive': boolean instead, which can be used by the BaseCommand class to filter out flags/args that have do not have the property 'non_sensitive'.

src/commands/config/get.ts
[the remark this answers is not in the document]
devinsag: Removed decorator, and instead have classes extend BaseCommand, to allow stack reporting directly from the oclif BaseCommand catch function.

src/base-command.ts
```diff
@@ -29,6 +39,110 @@ export default abstract class BaseCommand extends Command {
...
+
+    const sentry_session_metadata = {
+      command: this.id || (this.constructor as any).name,
+      environment: analytics_env,
+      email: auth_user?.email || '',
+      id: auth_login?.id || os.hostname(),
+      config_file: path.join(this.app?.config?.getConfigDir(), LocalPaths.CLI_CONFIG_FILENAME),
+      cwd: process.cwd(),
+      docker_info: sentry_session_docker_version,
+      docker_containers: sentry_session_docker_containers,
```
mueschm, line 126: Won't this only be set from when the command starts? Basically fi there is an error in the middle of the command and that command spun up some containers, we would not be able to see the new containers. This needs to happen when the error occurs and NOT on startup. The same goes for `linked_components`

src/base-command.ts
```diff
@@ -53,10 +167,45 @@ export default abstract class BaseCommand extends Command {
...
+      if (api_host === 'api.architect.io') {
+        this.oclif_env = 'production';
+      }
+      else if (api_host === 'api.dev.architect.io') {
+        this.oclif_env = 'dev';
+      } else {
+        this.oclif_env = 'local';
+      }
+      try {
+        return await this.setupAnalytics();
```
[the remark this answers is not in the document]
devinsag: removed return statement from init function

src/base-command.ts
devinsag: moved setting of metadata to occur at time of error during the catch function.

src/base-command.ts
```diff
@@ -53,10 +167,45 @@ export default abstract class BaseCommand extends Command {
         }
       }
     }
+
+    if (this.app && this.app.config) {
```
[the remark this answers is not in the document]
devinsag: moved config out of base command and instead into app

src/app-config/service.ts
```diff
@@ -45,6 +54,20 @@ export default class AppService {
     });
 
     const url = new URL(this.config.api_host);
+
+    if (process.env.TEST === '1') {
```
mueschm, line 58: This should be read in from the config file and not like this. TEST is okay, but the others are too flaky. Should default to Production but allow overriding from the config file.

devinsag: updated environment to be loaded and a part of the config file, with default values to be either test (if test) or production. Allows for overriding from the config file.

PR #591, opened 2022-06-09
Alert user on K8 version support

Zach-at-Arch:
When running `platform:create`, a message will show notifying
the user of the latest version of K8 we suport (up to v1.23).

If the user tries to register a K8 cluster v.1.24.0 or higher,
an error will be thrown

### DIFF
--- a/src/common/utils/kubernetes-platform.utils.ts
+++ b/src/common/utils/kubernetes-platform.utils.ts
@@ -71,6 +71,26 @@ export class KubernetesPlatformUtils {
       kube_context = new_platform_answers.context;
     }
 
+    /**
+     * Currently, we do not support K8 v1.24 or higher
+     * Below checks user's Server version to ensure it's under 1.24
+     */
+    const { stdout } = await execa('kubectl', [
+      ...set_kubeconfig,
+      'version', '--short=true']);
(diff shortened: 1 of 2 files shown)


src/common/utils/kubernetes-platform.utils.ts
```diff
@@ -71,6 +71,26 @@ export class KubernetesPlatformUtils {
...
+
+    const api_version = stdout.split('v').pop();
+    if (api_version) {
+      const [major_ver, minor_ver] = api_version.split('.').map(int => parseInt(int));
+
+      if (major_ver >= 1 && minor_ver >= 24) {
+        throw new ArchitectError('Architect currently does not support Kubernetes v1.24 or higher.');
+      }
+    } else {
+      throw new ArchitectError('Error reading kubectl version.');
```
ryan-cahill, line 90: Should we error here? I hesitate not to let someone continue in the flow in case we end up with some kind of edge case here with finding the version where the platform would actually be fine

Zach-at-Arch: Good call - I'll remove this

ryan-cahill (approved):

PR #597, opened 2022-06-14
Updating architect ref

ryan-cahill:
Setting the `architect_ref` in a cleaner way that shouldn't have problems with the API or frontend. This feels much cleaner than the way it was being created before as well

### DIFF
--- a/src/dependency-manager/spec/transform/resource-transform.ts
+++ b/src/dependency-manager/spec/transform/resource-transform.ts
@@ -77,6 +77,7 @@ export const transformResourceSpec = (resource_type: ResourceType, key: string, 
     metadata: {
       ...metadata,
       ref: spec.reserved_name || ResourceSlugUtils.build(component_account_name, component_name, resource_type, key, instance_name),
+      architect_ref: ResourceSlugUtils.build(component_account_name, component_name, resource_type, key, instance_name),
     },
     description: spec.description,
     image: spec.image,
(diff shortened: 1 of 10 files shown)


src/dependency-manager/spec/transform/resource-transform.ts
```diff
@@ -77,6 +77,7 @@ export const transformResourceSpec = (resource_type: ResourceType, key: string,
     metadata: {
       ...metadata,
       ref: spec.reserved_name || ResourceSlugUtils.build(component_account_name, component_name, resource_type, key, instance_name),
+      architect_ref: ResourceSlugUtils.build(component_account_name, component_name, resource_type, key, instance_name),
```
mueschm, line 80: Lets not rebuild this twice.

test/dependency-manager/schema/component-transform.unit.test.ts
```diff
@@ -17,6 +17,7 @@ describe('component transform unit test', function () {
 
     const metadata = {
       ref: 'tests/superset@instance-1',
+      architect_ref: 'tests/superset@instance-1',
```
mueschm, line 20: Make a const to make it clear they are the same value on purpose.

mueschm, line 20: Make a const to make it clear they are the same value on purpose.

mueschm (changes requested):

PR #599, opened 2022-06-17
(Sentry 2/3) 413 send transactions and exceptions to sentry

devinsag:
please see https://gitlab.com/architect-io/architect-cli/-/issues/413


src/base-command.ts
```diff
@@ -63,6 +64,16 @@ export default abstract class BaseCommand extends Command {
     }
   }
 
+  async finally(_: Error | undefined): Promise<any> {
+    if (this.app.config.environment !== ENVIRONMENT.TEST && !_) {
+      const { filtered_sentry_args, filtered_sentry_flags } = await this._getNonSensitiveSentryMetadata();
```
mueschm, line 69: This code looks duplicated, please move it to a new function that is just called from both places

mueschm (commented):
Had to merge in some of your changes which is creating a merge conflict. If you can fix that and the code duplication we can get this in ASAP.

PR #601, opened 2022-06-23
405 secrets download and upload

mtgoncalves1:
Closes https://gitlab.com/architect-io/architect-cli/-/issues/405.   

Commands to test this PR can be found in the issue above.

### DIFF
--- a/src/commands/secrets/download.ts
+++ b/src/commands/secrets/download.ts
@@ -32,8 +32,14 @@ export default class SecretsDownload extends Command {
 
   async run(): Promise<void> {
     const { flags, args } = await this.parse(SecretsDownload);
-    const secrets_file = path.resolve(untildify(args.secrets_file));
+
     const account = await AccountUtils.getAccount(this.app, flags.account);
+    const { data: user } = await this.app.api.get('/users/me');
+    const membership = user.memberships?.find((membership: any) => membership.account.id === account.id);
+    const is_admin = !!membership && membership.role !== 'MEMBER';
+    if (!is_admin) {
+      this.error('You do not have permission to download secrets. Please contact your admin.');
+    }
 
     let secrets = [];
     let environment;
@@ -49,9 +55,11 @@ export default class SecretsDownload extends Command {
       secret_yaml[secret.scope] = secret_yaml[secret.scope] || {};
       secret_yaml[secret.scope][secret.key] = secret.value;
     }
+
+    const secrets_file = path.resolve(untildify(args.secrets_file));
     await fs.writeFile(secrets_file, yaml.dump(secret_yaml), (err) => {
       if (err) {
-        throw new Error('Failed to download secrets!');
+        this.error('Failed to download secrets!');
       }
     });
 
--- a/src/commands/secrets/download.ts
+++ b/src/commands/secrets/download.ts
@@ -12,7 +12,7 @@ tmp.setGracefulCleanup();
 
 export default class SecretsDownload extends Command {
   async auth_required(): Promise<boolean> {
-    return false;
+    return true;
   }
 
   static description = 'Download secrets from an account or an environment';
@@ -27,7 +27,7 @@ export default class SecretsDownload extends Command {
   static args = [{
     name: 'secrets_file',
     description: 'Secrets filename to download secrets',
-    default: './secrets.yml',
+    required: true,
   }];
 
   async run(): Promise<void> {
@@ -42,16 +42,16 @@ export default class SecretsDownload extends Command {
     }
 
     let secrets = [];
-    let environment;
     if (!flags.environment) {
       secrets = (await this.app.api.get(`accounts/${account.id}/secrets/values`)).data;
     } else {
-      environment = await EnvironmentUtils.getEnvironment(this.app.api, account, flags.environment);
+      const environment = await EnvironmentUtils.getEnvironment(this.app.api, account, flags.environment);
       secrets = (await this.app.api.get(`environments/${environment.id}/secrets/values`, { params: { inherited: true } })).data;
     }
 
     if (secrets.length === 0) {
-      this.error('There are no secrets to be downloaded.');
+      this.log('There are no secrets to download.');
+      return;
     }
 
     const secret_yml: SecretsDict = {};
@@ -67,6 +67,6 @@ export default class SecretsDownload extends Command {
       }
     });
 
-    this.log(JSON.stringify(secret_yml, null, 4));
+    this.log(`Secrets have been downloaded to ${secrets_file}`);
   }
 }
--- a/src/commands/secrets/download.ts
+++ b/src/commands/secrets/download.ts
@@ -1,0 +1,60 @@
+import fs from 'fs-extra';
+import yaml from 'js-yaml';
+import path from 'path';
+import tmp from 'tmp';
+import untildify from 'untildify';
+import { Dictionary } from '../..';
+import AccountUtils from '../../architect/account/account.utils';
+import { EnvironmentUtils } from '../../architect/environment/environment.utils';
+import Command from '../../base-command';
+
+tmp.setGracefulCleanup();
+
+export default class SecretsDownload extends Command {
+  async auth_required(): Promise<boolean> {
+    return false;
+  }
+
+  static description = 'Download secrets from an account or an environment';
+  static aliases = ['secrets', 'secrets/get'];
+
+  static flags = {
+    ...Command.flags,
+    ...AccountUtils.flags,
+    ...EnvironmentUtils.flags,
+  };
+
+  static args = [{
+    name: 'secrets_file',
+    description: 'Secrets filename to download secrets',
+    default: './secrets.yml',
+  }];
+
+  async run(): Promise<void> {
+    const { flags, args } = await this.parse(SecretsDownload);
+    const secrets_file = path.resolve(untildify(args.secrets_file));
+    const account = await AccountUtils.getAccount(this.app, flags.account);
+
+    let secrets = [];
+    let environment;
+    if (!flags.environment) {
+      secrets = (await this.app.api.get(`accounts/${account.id}/secrets/values`)).data;
+    } else {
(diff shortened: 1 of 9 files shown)


src/commands/secrets/download.ts
```diff
@@ -0,0 +1,72 @@
...
+import AccountUtils from '../../architect/account/account.utils';
+import { EnvironmentUtils } from '../../architect/environment/environment.utils';
+import Command from '../../base-command';
+import { SecretsDict } from '../../dependency-manager/secrets/type';
+
+tmp.setGracefulCleanup();
+
+export default class SecretsDownload extends Command {
+  async auth_required(): Promise<boolean> {
+    return false;
```
mueschm, line 15: This should be true. If a user is not logged in there are no accounts to get secrets from.

src/commands/secrets/download.ts
```diff
@@ -0,0 +1,72 @@
...
+  static flags = {
+    ...Command.flags,
+    ...AccountUtils.flags,
+    ...EnvironmentUtils.flags,
+  };
+
+  static args = [{
+    name: 'secrets_file',
+    description: 'Secrets filename to download secrets',
+    default: './secrets.yml',
```
mueschm, line 30: This should not have a default and should be required for the user to specify a path. The main reason is we do not accidentally want to override their secret file if they do not have  a backup.

src/commands/secrets/download.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  async run(): Promise<void> {
+    const { flags, args } = await this.parse(SecretsDownload);
+
+    const account = await AccountUtils.getAccount(this.app, flags.account);
+    const { data: user } = await this.app.api.get('/users/me');
+    const membership = user.memberships?.find((membership: any) => membership.account.id === account.id);
+    const is_admin = !!membership && membership.role !== 'MEMBER';
+    if (!is_admin) {
+      this.error('You do not have permission to download secrets. Please contact your admin.');
```
mueschm, line 41: Really like the extra check to give useful information.

src/commands/secrets/download.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    let secrets = [];
+    let environment;
+    if (!flags.environment) {
+      secrets = (await this.app.api.get(`accounts/${account.id}/secrets/values`)).data;
```
mueschm, line 47: In the same way we have AccountUtils we should probably make a SecretUtils. Calling the `this.app.api` in a command file feels like we may be too tightly coupling implementations.

src/commands/secrets/download.ts
```diff
@@ -0,0 +1,72 @@
...
+    const account = await AccountUtils.getAccount(this.app, flags.account);
+    const { data: user } = await this.app.api.get('/users/me');
+    const membership = user.memberships?.find((membership: any) => membership.account.id === account.id);
+    const is_admin = !!membership && membership.role !== 'MEMBER';
+    if (!is_admin) {
+      this.error('You do not have permission to download secrets. Please contact your admin.');
+    }
+
+    let secrets = [];
+    let environment;
```
mueschm, line 45: This can be a `const` inside the else scope.

src/commands/secrets/download.ts
```diff
@@ -0,0 +1,72 @@
...
+import tmp from 'tmp';
+import untildify from 'untildify';
+import AccountUtils from '../../architect/account/account.utils';
+import { EnvironmentUtils } from '../../architect/environment/environment.utils';
+import Command from '../../base-command';
+import { SecretsDict } from '../../dependency-manager/secrets/type';
+
+tmp.setGracefulCleanup();
+
+export default class SecretsDownload extends Command {
```
mueschm, line 13: This should extend `BaseCommand` instead of `Command` otherwise we lose out on all our custom logic.

src/commands/secrets/upload.ts
```diff
@@ -0,0 +1,98 @@
...
+import path from 'path';
+import tmp from 'tmp';
+import untildify from 'untildify';
+import { Dictionary } from '../..';
+import AccountUtils from '../../architect/account/account.utils';
+import { EnvironmentUtils } from '../../architect/environment/environment.utils';
+import Command from '../../base-command';
+import { SecretsDict } from '../../dependency-manager/secrets/type';
+
+tmp.setGracefulCleanup();
```
mueschm, line 13: This does not seem to be used here.

src/commands/secrets/download.ts
```diff
@@ -0,0 +1,72 @@
...
+import yaml from 'js-yaml';
+import path from 'path';
+import tmp from 'tmp';
+import untildify from 'untildify';
+import AccountUtils from '../../architect/account/account.utils';
+import { EnvironmentUtils } from '../../architect/environment/environment.utils';
+import Command from '../../base-command';
+import { SecretsDict } from '../../dependency-manager/secrets/type';
+
+tmp.setGracefulCleanup();
```
mueschm, line 11: This does not seem to be used here.

src/commands/secrets/download.ts
```diff
@@ -0,0 +1,72 @@
...
+    let environment;
+    if (!flags.environment) {
+      secrets = (await this.app.api.get(`accounts/${account.id}/secrets/values`)).data;
+    } else {
+      environment = await EnvironmentUtils.getEnvironment(this.app.api, account, flags.environment);
+      secrets = (await this.app.api.get(`environments/${environment.id}/secrets/values`, { params: { inherited: true } })).data;
+    }
+
+    if (secrets.length === 0) {
+      this.error('There are no secrets to be downloaded.');
```
mueschm, line 54: Error text should generally use yellow or red in order to display information. Since this is more feedback and not a hard error, making it yellow would be useful.

src/commands/secrets/download.ts
```diff
@@ -0,0 +1,72 @@
...
+    const secret_yml: SecretsDict = {};
+    for (const secret of secrets) {
+      secret_yml[secret.scope] = secret_yml[secret.scope] || {};
+      secret_yml[secret.scope][secret.key] = secret.value;
+    }
+
+    const secrets_file = path.resolve(untildify(args.secrets_file));
+    await fs.writeFile(secrets_file, yaml.dump(secret_yml), (err) => {
+      if (err) {
+        this.error('Failed to download secrets!');
```
mueschm, line 66: This is a good example of an error that should be red.

src/commands/secrets/download.ts
```diff
@@ -0,0 +1,72 @@
...
+      }
+    });
+
+    this.log(JSON.stringify(secret_yml, null, 4));
```
mueschm, line 70: We should not print out secrets to a user's terminal. This is a breach of privacy. We should give them a green "Secrets have been download to <path>"

src/commands/secrets/upload.ts
```diff
@@ -0,0 +1,98 @@
...
+    name: 'secrets_file',
+    description: 'Secrets file to be uploaded',
+    required: true,
+  }];
+
+  async run(): Promise<void> {
+    const { flags, args } = await this.parse(SecretsUpload);
+    const account = await AccountUtils.getAccount(this.app, flags.account);
+
+    const { data: user } = await this.app.api.get('/users/me');
```
mueschm, line 49: All duplicated logic should be moved to a single file. We could have a UserUtil with a function that returns membership type for instance.

src/commands/secrets/upload.ts
```diff
@@ -0,0 +1,98 @@
...
+    // loaded secrets
+    const secrets_file = path.resolve(untildify(args.secrets_file));
+    const secrets_file_data = fs.readFileSync(secrets_file);
+    const loaded_secret_yml = yaml.load(secrets_file_data.toString('utf-8')) as Dictionary<Dictionary<string>>;
+    if (!loaded_secret_yml) {
+      this.error('There are no secrets to be uploaded.');
+    }
+
+    const existing_secret_yml: SecretsDict = {};
+    for (const secret of existing_secrets) {
```
mueschm, line 75: I thought the batch update to the server handled this for us? Lets double check because would be nice not to duplicate this logic in multiple places.

mueschm (changes requested):
Really like the overall thought put into the user experience, just need to tighten up the code a little bit.

src/commands/secrets/upload.ts
mtgoncalves1: Totally understand. From the secret-file-upload code in the frontend it does all of these steps before calling `secret-batch-update` - https://gitlab.com/architect-io/architect-hub/blob/b6fee30a357602a3dbf1fbf4f79bd7cb40bd8c32/src/components/common/secrets/secrets.vue#L384

src/commands/secrets/download.ts
mtgoncalves1: That makes sense.

mtgoncalves1: I will also make the same change in the `upload.ts` as well.

src/commands/secrets/download.ts
mtgoncalves1: This prints out to my terminal and it is white. I do agree that it shouldn't be an error so I will update it to this.log() instead.

src/commands/secrets/download.ts
mtgoncalves1: That makes sense. Removed and made it required.

src/commands/secrets/download.ts
mtgoncalves1: Updated to `true`

src/commands/secrets/download.ts
```diff
@@ -67,6 +53,6 @@ export default class SecretsDownload extends Command {
       }
```
mueschm, line 53: This code is the issue.
`fs.writeFile` uses the old callback style and not the `Promise` style of async execution. The `await` here actually does not do anything. Instead the write goes out, but because oclif has no concept of the callback function in the `writeSync` it just goes a head with the test.
Instead you can use this library to get async support.
https://www.npmjs.com/package/fs-extra

PR #607, opened 2022-06-24
(Sentry 3/3) 428 write transactions to filesystem and add doctor command

devinsag:
please see https://gitlab.com/architect-io/architect-cli/-/issues/428


src/base-command.ts
```diff
@@ -29,6 +43,119 @@ export default abstract class BaseCommand extends Command {
     });
   }
 
+  async _filterNonSensitiveSentryMetadata(non_sensitive: Set<string>, metadata: any): Promise<any> {
```
tjhiggins, line 46: Can we move all of this sentry logic into its own class that gets instantiated in the BaseCommand constructor kind of like the app/api?

tjhiggins (changes requested):

src/base-command.ts
devinsag: Moved the sentry logic out of BaseCommand and into a new class, SentryService.

src/static/doctor.html
```diff
@@ -0,0 +1,77 @@
...
+        tds.filter((cell_contents) => !(cell_contents instanceof Array || cell_contents instanceof Object))
+          .text((cell_contents) => cell_contents);
+
+        // new table if array or object
+        tds.filter((cell_contents) => (cell_contents instanceof Array || cell_contents instanceof Object))
+          .append("table").call(build_table);
+      });
+    }
+  </script>
+  </script>
```
TylerAldrich, line 75: Was reading this PR for fun and don't know what most of it does, but it looks like this is a dupe`</script>` tag 😛

devinsag: removed extra <script> tag

ryan-cahill (changes requested):

src/commands/doctor.ts
```diff
@@ -0,0 +1,142 @@
...
+    DEFAULT_VALUE: true,
+  },
+};
+
+export default class Doctor extends BaseCommand {
+  async auth_required(): Promise<boolean> {
+    return false;
+  }
+
+  history: any[] = [];
```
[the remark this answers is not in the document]
devinsag: added SENTRY_HISTORY interface to describe current metadata.

src/commands/doctor.ts
```diff
@@ -0,0 +1,142 @@
...
+          }
+        }
+      });
+      server.listen(port);
+    });
+  }
+
+  async run(): Promise<void> {
+    const inquirer = require('inquirer');
+    const inquirerPrompt = require('inquirer-autocomplete-prompt');
```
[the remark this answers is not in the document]
devinsag: moved inquirer to import instead of using require.  When trying to import inquirer-autocomplete-prompt it throws an exception. the repo shows examples using requires, found here:

https://github.com/mokkabonna/inquirer-autocomplete-prompt

src/sentry.ts
```diff
@@ -178,7 +185,45 @@ export default class SentryService {
...
+    }
+    return command_history;
+  }
+
+  async writeCommandHistoryToFileSystem(): Promise<void> {
+    const scope = Sentry.getCurrentHub()?.getScope();
+    const remove_keys = new Set(['plugins', 'pjson', 'oauth_client_id', 'credentials', '_auth_result', 'Authorization']);
+
+    // remove empty objects, empty strings, or key/values from the remove_keys list.
+    let current_output = JSON.parse(JSON.stringify(scope as any, (key, value) => {
```
ryan-cahill, line 208: Rather than stringifying then parsing `scope`, why not just iterate over key/value pairs and set values to remove equal to undefined?

src/sentry.ts
```diff
@@ -178,7 +185,45 @@ export default class SentryService {
...
+      return ((value && !remove_keys.has(key) && Object.keys(value).length)) ? value : undefined;
```
TylerAldrich, line 209: I'm not sure if this is relevant for this function in particular, but I think `Object.keys(value).length` when `value` is a Number is going to return 0 which means numbers also get filtered out if I'm reading this correctly

devinsag: good catch, thank you. updated to check if value is a number,

PR #610, opened 2022-07-06
fixing interpolation if a service config string has a -- in it

ryan-cahill:
Our interpolation was using a forward lookahead for an unlimited amount of characters before, which could catch something like an `curl --fail localhost:3000/users || exit 1` in the service config. The `--` would fail the regex match, and therefore not find the edges of a service for the graph. The regex has been updated to take this into account, and a test was added

### DIFF
--- a/src/commands/secrets/download.ts
+++ b/src/commands/secrets/download.ts
@@ -20,6 +20,7 @@ export default class SecretsDownload extends BaseCommand {
   };
 
   static args = [{
+    non_sensitive: true,
     name: 'secrets_file',
     description: 'Secrets filename to download secrets',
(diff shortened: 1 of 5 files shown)


src/commands/secrets/download.ts
```diff
@@ -20,6 +20,7 @@ export default class SecretsDownload extends BaseCommand {
   };
 
   static args = [{
+    non_sensitive: true,
```
tjhiggins, line 23: What are non_sensitive secrets?

tjhiggins: nvm I see this is used everywhere now

PR #612, opened 2022-07-07
446 project command

mtgoncalves1:


mueschm:
This is definitely on the right track but there are some changes we should make.
1. Frontend, Backend, and Database are not the only types we allow. The point of these names is that if I ask for `react` for my project it should go to that template.json and get a list of elements to ask me for, that list could point to 10 other required service types all with unique names. This gives control over to the marketing and customer success team rather than having us update our code each time.
2. The source of truth should not be in the project. We should just let users pick any project name and then look to see if that project exists on Github to kick off the flow. 
3. Looking at the way the databases work we may want to just allow custom architect.yml in the template.json file that just gets added directly into the generated architect.yml.

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,319 @@
...
+      .then(res => res.blob())
+      .then(blob => blob.text())
+      .then(yaml_as_string => yaml.load(yaml_as_string) as ComponentSpec)
+      .catch(err => {
+        throw new Error(`Failed to fetch ${url}`);
+      });
+    return component_spec;
+  }
+
+  static async getSelections(chosen_project: string): Promise<Dictionary<Selection>> {
```
mueschm, line 84: This is a god function, and we need to split up the logic more.

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,319 @@
...
+    // find project repository in our GitHub
+    const github_repos = await this.getGitHubRepos('https://api.github.com/orgs/architect-templates/repos');
+    const start_project = github_repos.find(repo => repo.name === chosen_project.toLowerCase());
+    if (!start_project) {
+      const repo_names = github_repos.map(repo => repo.name).join(', ');
+      throw new Error(`Cannot find project '${chosen_project}'. Available projects are: ${repo_names}.`);
+    }
+
+    // store component spec of the chosen project
+    const yaml_url = 'https://raw.githubusercontent.com/' + start_project.full_name + `/${GITHUB_BRANCH}/architect.yml`;
```
mueschm, line 99: So we kind of duplicate this logic. The first time we initalize the queue and the second time to keep adding new choices to the queue. If the queue instead was just a list of repos to add we could move all the logic into the while loop and not duplicate it.

So we just take the starting component name add it to the queue and then enter the loop. All logic from there just happens in the loop.

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,319 @@
...
+          continue;
+        }
+      }
+
+      let is_prompt = false;
+      if (proj.required) { 
+        is_prompt = true;
+      } else {
+        const answer = await this.prompt(['yes', 'no'], `${proj_type} is optional. Would you like to select a ` + proj_type.toLowerCase() + '?');
+        if (answer === 'yes') {
```
mueschm, line 136: One line shorthand for this stuff.
```
is_prompt = answer === 'yes'
```

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,319 @@
...
+          const service_name = Object.keys(services).find((name: any) => !services[name].image);
+          selection.depends_on = service_name;
+        }
+      }
+    }
+
+    let interfaces: Dictionary<any> = {};
+    let dependencies: Dictionary<any> = {};
+    let secrets: Dictionary<any> = {};
+    let services: Dictionary<any> = {};
```
mueschm, line 265: These should be determined generically based on the loaded architect.yml files. For instance, if we want to showcase `tasks` this code will not work.

src/commands/project/create.ts
```diff
@@ -0,0 +1,46 @@
...
+    description: 'Name of your project',
+    required: true,
+  }];
+
+  async run(): Promise<void> {
+    const { flags, args } = await this.parse(ProjectCreate);
+
+    let chosen_project;
+    if (!flags.project) {
+      const answer: { chosen: any } = await inquirer.prompt([
```
mueschm, line 30: You have a function that gets the repo name. We should just pass that in as a list of choices here since its only a few lines of code now.

mueschm (changes requested):
So there is some high level things we should fix first, once those are done I will do another deep dive pass since these require some big code changes.

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,290 @@
...
+interface Selection {
+  name: string,
+  architect_yml: Dictionary<any>,
+  depends_on: string,
+}
+
+const GITHUB_BRANCH = 'main';
+
+class Queue<T> {
+  items: any[];
```
mueschm, line 25: This should be.
```
items: T[];
```
We should avoid the use of `any` anywhere in this class.

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,290 @@
...
+    return this.items.some(item => item.hasOwnProperty(key));
+  }
+
+  size() {
+    return this.items.length;
+  }
+}
+
+export default class ProjectUtils {
+  static queue = new Queue();
```
mueschm, line 53: Static like this should be avoided. The intent is good, but the issue is that this Util class now has state that must be managed. If we are going to add state we should generally do 1 of 2 things.
1. Create a new class that stores this as instance variables and hide most of the logic away as private functions.
2. Create a data class that stores the intermediary state necessary for the operation to work and pass that through each function.

In this instance, I would opt for number 1 and create a new src folder under `src/common` called project-generation.

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,290 @@
...
+          combined_yml[key] = combined_yml[key] ? { ...combined_yml[key], ...val } : val;
+          continue;
+        }
+
+        const yml_services = val as Dictionary<ServiceSpec>;
+        const service_keys = Object.keys(yml_services);
+        if (service_keys.length === 1) {
+          const service_key = service_keys[0];
+          const service = yml_services[service_key];
+          if (service.build) {
```
mueschm, line 259: Why do we need to do this?

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,290 @@
...
+              if (selection.depends_on) {
+                combined_yml[key][service_name]['depends_on'] = [selection.depends_on];
+              }
+            }
+          }
+        }
+      }
+    }
+
+    const app_key = Object.keys(combined_yml['interfaces']).find(key => key === 'app');
```
mueschm, line 280: Is this because multiple starter projects use the same interface name?

mueschm (changes requested):
This feels like we are getting very close!

src/architect/project/project.utils.ts
mtgoncalves1: No, it is because they use different interface names. When the interfaces combine, we don't know which to choose. So, I made it look for the interface named `app`. If `app` doesn't exist, then just pick the first one interface.

src/architect/project/project.utils.ts
mtgoncalves1: It is for handling where to put the `depends_on` but since we will be updating how we handle `depends_on`, I will remove that.

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,229 @@
+import fs from 'fs-extra';
+import inquirer from 'inquirer';
+import yaml from 'js-yaml';
+import fetch from 'node-fetch';
```
tjhiggins, line 4: why are we using fetch over axios? In general I'd like to be consistent about the lib as opposed to devs have to choose and then learn both.

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,229 @@
...
+import path from 'path';
+import untildify from 'untildify';
+import { buildSpecFromPath, ComponentSpec } from '../../';
+import AppService from '../../app-config/service';
+import { EnvironmentSpecValue } from '../../dependency-manager/spec/resource-spec';
+import { Dictionary } from '../../dependency-manager/utils/dictionary';
+
+const download = require('download-git-repo');
```
tjhiggins, line 12: what is the benefit of this package as opposed to shelling out via execa?

`execa(git clone ...)`

src/commands/init.ts
```diff
@@ -68,29 +77,30 @@ export abstract class InitCommand extends BaseCommand {
...
+    if (fs.existsSync(`./${project_name}`)) {
+      console.log(chalk.red(`The folder ./${project_name} already exists. Please choose a different project name or remove the folder`));
+      return;
+    }
+    this.log(`Start the process to create your project '${project_name}'.`);
+    const selections = await ProjectUtils.getSelections();
+    await ProjectUtils.downloadGitHubRepos(selections, project_name);
+    await ProjectUtils.updateArchitectYamls(this.app, selections, project_name);
+    await ProjectUtils.linkSelections(this.app, selections, project_name);
+    this.log(`Successfully created project ${project_name}.`);
```
tjhiggins, line 96: Can we make the call to action the last thing we log out?

```
Successfully created project test.

Now try running your project: `architect dev test/react/architect.yml`
```

Also we should color the command to make it stand out. `chalk.blue('architect dev test/react/architect.yml')`. Want to make it abundantly clear what the next step should be.

tjhiggins (changes requested):

src/commands/init.ts
mtgoncalves1: That's a great idea. We write out a README file so I just point them to it. 
```
Successfully created project my-proj.
Now try running your project using the instructions in my-proj/README.md.
```

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,229 @@
...
+    const config_file = 'https://raw.githubusercontent.com/architect-team/template-configs/main/config.json';
+    const config_json = await this.fetchJsonFromGitHub(config_file) as Dictionary<any>;
+    const choices = config_json.choices;
+
+    const selections: Dictionary<Selection> = {};
+    const types = ['Full stack', 'Backend'];
+    const type = await this.prompt(types, 'Select a type you would like to proceed');
+    if (type.toLowerCase() === 'full stack') {
+      const frontend_opts = choices.filter((item: any) => item.type === 'frontend');
+      const frontend = await this.prompt(frontend_opts, 'Select a frontend');
```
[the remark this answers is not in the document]
mtgoncalves1: We can disable it if you'd like.

src/commands/init.ts
```diff
@@ -98,24 +80,48 @@ export abstract class InitCommand extends BaseCommand {
...
+        }]);
+      args.name = answer.chosen;
+    }
+
+    if (!ComponentSlugUtils.Validator.test(args.name)) {
+      if (flags['from-compose']) {
+        throw new Error(`Component name can only contain lowercase letters and dashes, and must start and end with a letter.`);
+      }
+      throw new Error(`Project name can only contain lowercase letters and dashes, and must start and end with a letter.`);
+    }
```
Zach-at-Arch, line 107: The UX is pretty jarring here - if the user doesn't create a project that adheres to the standards, it errors out and they have to run `init` again. Two things to add to this:

1. Don't Error - just warn if they don't follow the rules and allow them to re-name their project
2. Give a hint before prompting. Something like:
`Please provide a name for your project. \n use only lowercase letters and dashes, and do not end or start with a dash`

mtgoncalves1: Thanks for the review, Zach! I will work on that.

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,215 @@
...
+        }
+      }
+    }
+    return service_environment;
+  }
+
+  static switchDatabase(database_yml: ComponentSpec, backend_yml: ComponentSpec, backend_yml_path: string): void {
+    if (!backend_yml.services || !database_yml.services) {
+      return;
+    }
```
Zach-at-Arch, line 106: Is this still WIP? I tried creating multiple project w/ different database and they still all have the postgre image in them

mtgoncalves1: Hmm I'll look into it.

src/architect/project/project.utils.ts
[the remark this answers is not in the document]
Zach-at-Arch: My thoughts: 

- for V1 we'll disable it. You'll select your stack, and since we only have one selection for Front/backend, it will just default to those
- For V2 after we get more options to choose from, we'll add this feature back in to allow the user to select what they want

src/architect/project/project.utils.ts
Zach-at-Arch: @mtgoncalves1 let's switch to using `execa`, unless there was additional benifits from `download-git-repo` package. 

UX portions later in the script will also need to leverage `execa` for linking and other things

mtgoncalves1: Sorry, that was already addressed.

Zach-at-Arch (approved):

Zach-at-Arch:
@mueschm @tjhiggins Init command is ready to go and tested. As a quick note - put a [PR for the config template ](https://github.com/architect-team/template-configs/pull/3), so the latest projects won't show in the selections until after it's merged (ASP, Ruby)

src/commands/init.ts
```diff
@@ -133,6 +195,27 @@ export abstract class InitCommand extends BaseCommand {
...
+      from_path = path.resolve(untildify(flags['from-compose']));
+    } else {
+      const files_in_current_dir = fs.readdirSync('.');
+      const default_compose = files_in_current_dir.find(f => f.includes('compose') && (f.endsWith('.yml') || f.endsWith('.yaml')));
+
+      if (default_compose) {
+        from_path = default_compose;
+        if (!fs.existsSync(from_path) || !fs.statSync(from_path).isFile()) {
+          throw new Error(`The Docker Compose file ${from_path} couldn't be found.`);
+        }
```
TylerAldrich, line 218: Similar thing here, this is an async method but we're only doing synchronous operations, so we can make the method sync or use the async version of readdir/exists/stats to benefit from async-ness 😛

src/commands/init.ts
```diff
@@ -56,6 +61,60 @@ export abstract class InitCommand extends BaseCommand {
...
+    name: 'name',
+    description: 'Name of your component or project',
+    required: false,
+  }];
+
+  async doesDockerComposeYmlExist(): Promise<boolean> {
+    const files_in_current_dir = fs.readdirSync('.');
+    const default_compose = files_in_current_dir.some(f => f.includes('compose') && (f.endsWith('.yml') || f.endsWith('.yaml')));
+    return default_compose;
+  }
```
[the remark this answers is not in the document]
Zach-at-Arch: I'm going to keep this one sync since we're awaiting `doesDockerComposeYmlExist` anyway

src/commands/init.ts
Zach-at-Arch: I'm keeping this one sync as well - we need to make sure we can actually get the compose file; otherwise, it will throw undefined for the `from_path` variable and cause the file to not be found and fail the tests

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  static async getSelections(): Promise<Dictionary<Selection>> {
+    // get choices from template-configs repository
+    const config_file = 'https://raw.githubusercontent.com/architect-team/template-configs/main/config.json';
+    const config_json = await this.fetchJsonFromGitHub(config_file) as Dictionary<any>;
+
+    const choices = config_json.choices;
+
+    const project = await this.prompt(choices, 'Please select a framework/language for your project');
```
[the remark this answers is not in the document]
Zach-at-Arch: So I'm going to keep this for now since `prompt` is used in a few other places, and by casting to a specific type it ends up breaking the other use cases

src/architect/project/project.utils.ts
```diff
@@ -0,0 +1,75 @@
...
+    // download repositories
+    if (selection.repository) {
+      await execa('git', ['clone', selection.repository.toString(), project_dir + '/' + selection.name.toLowerCase()], { stdio: 'ignore' });
+      await PromptUtils.oclifTimedSpinner(
+        `Pulling down GitHub Repository`,
+         selection.name.toLowerCase(),
+        `${chalk.green('✓')} ${selection.name.toLowerCase()}`,
+      );
+    }
+    await new Promise(resolve => setTimeout(resolve, 2000));
```
TylerAldrich, line 63: Why do we wait for 2s after cloning?

Zach-at-Arch: That is a great question - I have no clue; I'll remove it.

TylerAldrich: Haha cool - was going to say if there's a reason we should add a comment since it looked random, but maybe was just random 😛. There's also a couple merge conflicts in the package.json/lock files but once fixed I think this is good to go

Zach-at-Arch: Sounds good! got everything all resolved

TylerAldrich:
Code-wise everything looks good now, I just tried out the tool and noticed a couple things that are a bit misleading that we should reword:

```
? What should the name of the component be? test-comp
```

This doesn't seem to actually be the name of the component, but rather the folder that we create to clone the project into. I tested out cloning the Go project and I end up with a file structure like:
```
test-comp/
    go/
        architect.yml
        (other files...)
```

At the end, we then say:
```
To Deploy locally, run:
	$ architect dev test-comp/architect.yml
```

This isn't accurate because we create the `test-comp` folder, then clone the `go` project within it, so the real path would be `architect dev test-comp/go/architect.yml` (and would change for every project)


Perhaps what we should do to fix this is:
1. Change "What should the name of the component be" to something that indicates it's just naming a folder (like "what's the name of your project") OR actually modify the cloned architect.yml with the name chosen (the former is easier of course, probably a good starting place because we don't need to be too fancy)
2. We should run `git clone <git repo url> .` to clone into the folder we created so the file structure is:
```
test-comp/
    architect.yml
    (other files...)
```
With the updated clone command the final message won't be a lie and the `dev` command could be copy-pasta'd successfully

TylerAldrich:
Realized Zach was out today so I fixed the things I suggested above

Zach-at-Arch:
@TylerAldrich muchas gracias

PR #625, opened 2022-07-14
453 debug block

mtgoncalves1:


src/commands/dev.ts
```diff
@@ -384,6 +391,10 @@ export default class Dev extends BaseCommand {
       dependency_manager.environment = 'local-production';
     }
 
+    if (flags.debug === 'false') {
+      dependency_manager.environment = 'local-production';
```
tjhiggins, line 395: could you add a boolean `debug: true` on the local dependency_manager?

```ts
if (flags.debug === 'false') {
  dependency_manager.debug = false;
}
```

I don't want debug to be tied to environments anymore. Then in the manager you can use `this.debug` to determine whether to merge.

tjhiggins (changes requested):

src/common/dependency-manager/local-manager.ts
```diff
@@ -109,15 +109,15 @@ export default class LocalDependencyManager extends DependencyManager {
     if (spec.metadata.file?.path && this.environment === 'local') {
```
tjhiggins, line 109: I think I want it to only merge if debug is true - the environment could be anything. Let me know if this doesn't seem right.

```ts
if (spec.metadata.file?.path && debug) {
```

tjhiggins (changes requested):

src/common/dependency-manager/local-manager.ts
mtgoncalves1: That makes sense.

PR #640, opened 2022-07-25
return empty when reading empty json

devinsag:
please see https://gitlab.com/architect-io/architect-cli/-/issues/469

### DIFF
--- a/src/app-config/service.ts
+++ b/src/app-config/service.ts
@@ -63,9 +63,12 @@ export default class AppService {
 
   private loadLinkedComponents(config_dir: string) {
     const linkedComponentsFile = path.join(config_dir, LocalPaths.LINKED_COMPONENT_MAP_FILENAME);
-    if (fs.existsSync(linkedComponentsFile)) {
+    if (!fs.existsSync(linkedComponentsFile)) {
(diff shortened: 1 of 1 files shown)


src/app-config/service.ts
```diff
@@ -63,9 +63,12 @@ export default class AppService {
...
   private loadLinkedComponents(config_dir: string) {
     const linkedComponentsFile = path.join(config_dir, LocalPaths.LINKED_COMPONENT_MAP_FILENAME);
-    if (fs.existsSync(linkedComponentsFile)) {
+    if (!fs.existsSync(linkedComponentsFile)) {
+      return {};
+    }
+    try {
       return fs.readJSONSync(linkedComponentsFile) as Dictionary<string>;
-    } else {
+    } catch {
```
tjhiggins, line 71: Can you add a warning? 
`console.warn('Failed to read linked components file.');`

tjhiggins (changes requested):

src/app-config/service.ts
devinsag: Updated to console.warn on readJSONSync exception before returning

PR #642, opened 2022-07-27
Ephemeral component versions

ryan-cahill:
Adds support for ephemeral components by allowing a user to register a component version with an environment


src/commands/deploy.ts
```diff
@@ -232,10 +236,25 @@ export default class Deploy extends DeployCommand {
...
 
-    const deployment_dtos = [];
+    const component_names: string[] = [];
     for (const component of components) {
+      if (fs.existsSync(component)) {
+        const register = new ComponentRegister([component, '-a', account.name, '-e', environment.name], this.config);
+        register.app = this.app;
+        await register.run();
+        const component_spec = buildSpecFromPath(component);
+        component_names.push(`${account.name}/${component_spec.name}:${ComponentRegister.getTagFromFlags({ environment: environment.name })}`);
```
tjhiggins, line 246: Remove the account name prefix. Since sometimes the account name is part of the name for older customers.

```ts
component_names.push(`${component_spec.name}:${ComponentRegister.getTagFromFlags({ environment: environment.name })}`);
```

src/commands/register.ts
```diff
@@ -56,6 +60,14 @@ export default class ComponentRegister extends BaseCommand {
         description: 'Directory to write build cache to. Do not use in Github Actions: https://docs.architect.io/deployments/automated-previews/#caching-between-workflow-runs',
       }),
     },
+    environment: {
```
tjhiggins, line 63: should we make this `hidden: true`?

src/commands/register.ts
```diff
@@ -20,6 +21,8 @@ import { IF_EXPRESSION_REGEX } from '../dependency-manager/spec/utils/interpolat
 
 tmp.setGracefulCleanup();
 
+export const EPHEMERAL_DELIMITER = 'architect-ephemeral';
```
tjhiggins, line 24: Should we make it targeted to the environment? Not really a delimiter - more a prefix.
```
export const ENV_TAG_PREFIX = 'architect.environment.'
```

tjhiggins (changes requested):

src/commands/register.ts
ryan-cahill: I had considered that, but I don't see why someone shouldn't be able to create an ephemeral component version directly rather than indirectly thorough the deploy command

tjhiggins: What would be the reason for them to do that? My inclination is to make it hidden with the option to unhide in the future.

ryan-cahill: Maybe they just don't want to deploy the component immediately after it's registered? I'll hide it though

src/commands/register.ts
ryan-cahill: Yeah I like that

PR #643, opened 2022-07-28
Generate yarn.lock during releases

TylerAldrich:


package.json
```diff
@@ -150,6 +150,7 @@
     "pack": "oclif pack:tarballs",
     "readme": "./bin/readme",
     "generate:schema-docs": "ts-node --project ./tsconfig.json ./src/dependency-manager/generate.ts",
+    "generate:yarn-lock": "(rm yarn.lock || true) && yarn import",
```
tjhiggins, line 153: Will yarn be installed in the pipeline? The image is `ubuntu-latest`

TylerAldrich: Mmm good question, it looks like `Yarn 1.22.19` is included in `ubuntu-latest` https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md

PR #644, opened 2022-07-28
add try/catch to file io and sentry end transactn

devinsag:
please see https://gitlab.com/architect-io/architect-cli/-/issues/471

### DIFF
--- a/src/sentry.ts
+++ b/src/sentry.ts
@@ -143,19 +143,22 @@ export default class SentryService {
 
   async endSentryTransaction(write_command_out: boolean, error?: any): Promise<void> {
     if (this.app.config.environment === ENVIRONMENT.TEST) return;
+    try {
+      await this.updateSentryTransaction(error);
 
-    await this.updateSentryTransaction(error);
-
-    if (this.file_out && write_command_out) {
-      await this.writeCommandHistoryToFileSystem();
-    }
-
-    if (this.sentry_out) {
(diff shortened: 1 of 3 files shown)


src/sentry.ts
```diff
@@ -143,19 +143,22 @@ export default class SentryService {
...
-        Sentry.withScope(scope => Sentry.captureException(error));
+      if (this.sentry_out) {
+        Sentry.getCurrentHub().getScope()?.getSpan()?.finish();
+        Sentry.getCurrentHub().getScope()?.getTransaction()?.finish();
+        if (error) {
+          Sentry.withScope(scope => Sentry.captureException(error));
+        }
       }
+    } catch {
+      // If endSentryTransaction fails, don't error.
```
tjhiggins, line 161: Add some console.debug to these - an empty catch can be really confusing

tjhiggins (changes requested):

src/sentry.ts
devinsag: Added try/catch console.debug messages to all methods in SentryService, and one in BaseCommand sentry fails to initialize for currently unknown use cases.

```typescript
try {
  this.sentry = await SentryService.create(this.app, this.constructor as any);
} catch (e) {
  console.debug("SENTRY: an error occurred creating a new instance of SentryService");
}
if (!this.sentry) {
  console.debug("SENTRY: SentryService failed to initialize");
}
```

PR #645, opened 2022-08-01
472 exit code on validation errors

devinsag:
Please see https://gitlab.com/architect-io/architect-cli/-/issues/472

### DIFF
--- a/src/base-command.ts
+++ b/src/base-command.ts
@@ -72,9 +72,6 @@ export default abstract class BaseCommand extends Command {
   }
 
   async finally(_: Error | undefined): Promise<any> {
-    if (_ instanceof ValidationErrors) {
-      process.exitCode = 1;
-    }
     return await this.endSentryTransaction();
   }
 
@@ -117,6 +114,7 @@ export default abstract class BaseCommand extends Command {
 
   async catch(err: any): Promise<void> {
     if (err.oclif && err.oclif.exit === 0) return;
+    process.exitCode = 1;
 
     try {
 
--- a/src/base-command.ts
+++ b/src/base-command.ts
@@ -72,6 +72,9 @@ export default abstract class BaseCommand extends Command {
   }
 
   async finally(_: Error | undefined): Promise<any> {
+    if (_ instanceof ValidationErrors) {
+      process.exitCode = 1;
+    }
     return await this.endSentryTransaction();
   }
 
--- a/test/dependency-manager/validation.test.ts
+++ b/test/dependency-manager/validation.test.ts
@@ -829,6 +829,7 @@ services:
       expect(errors).lengthOf(1);
       expect(err.message).includes(`services.app.labels.environment2`);
       expect(err.message).includes('must match pattern');
+      expect(process.exitCode).eq(1);
     });
 
     it('invalid labels length', async () => {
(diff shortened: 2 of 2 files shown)


src/base-command.ts
```diff
@@ -72,6 +72,9 @@ export default abstract class BaseCommand extends Command {
   }
 
   async finally(_: Error | undefined): Promise<any> {
+    if (_ instanceof ValidationErrors) {
+      process.exitCode = 1;
```
tjhiggins, line 76: Don't we want it to exit 1 for any error? Especially uncaught ones. 

Why are we calling `await this.endSentryTransaction();` in both finally and catch methods? I'd imagine it should only exist in the finally method since that will "should" always run.

Also why doesn't catch always call `super.catch(err);`? That would be the reason we aren't spitting out the correct error code.

tjhiggins (changes requested):

src/base-command.ts
tjhiggins: @mueschm you have more context and I'd def like to get a fix in for this tomorrow after the demo

test/dependency-manager/validation.test.ts
```diff
@@ -423,22 +432,23 @@ services:
...
       const errors = JSON.parse(err.message) as ValidationError[];
       expect(errors).lengthOf(3);
       expect(errors.map(e => e.path)).members([
         'services.stateful-app.depends_on',
         'services.api.depends_on',
-        'services.backend.depends_on'
-      ])
+        'services.backend.depends_on',
+      ]);
+      expect(process.exitCode).eq(1);
```
tjhiggins, line 443: This is a great call to check exit code in the tests. However, I'm not sure this is the correct place for it. We want to test e2e against the base command finally/catch. 

Include a check if the `commands/register.test.ts` and `validate.test.ts` where we are actually running the oclif commands.

tjhiggins (changes requested):

src/base-command.ts
devinsag: BaseCommand.catch attempts to send sentry metadata before calling the super.catch method in order to get running docker context written to sentry/file before any cleanup is performed.  

If the error is a validation error, or is err.oclif.exit === 0, it will not call super.catch from the catch method. I suppose this could be an artifact from when we were returning prettyValidationErrors(err) before sentry was added. 
If super.catch() is called with validation error, it will double print the error message to screen. I will re-evaluate the logic block.

src/base-command.ts
```diff
@@ -72,6 +72,9 @@ export default abstract class BaseCommand extends Command {
   }
 
   async finally(_: Error | undefined): Promise<any> {
```
tjhiggins, line 74: We are also not calling super.finally which we probably should do. See:
https://github.com/oclif/core/blob/main/src/command.ts#L268

Anytime we override a base class method you almost always want to call the super method.

src/base-command.ts
mueschm: Will help to test this, but pretty sure we do just always want to return the 1 on this finally.

src/base-command.ts
devinsag: super.finally() and super.catch() should be executed at the end of the endSentryTransaction() function.

https://github.com/architect-team/architect-cli/blob/rc/src/base-command.ts#L189-L192

devinsag:
closing in favor of PR with an updated solution here: https://github.com/architect-team/architect-cli/pull/648

PR #655, opened 2022-08-05
adding process.exitCode = 1 for errs in catch fn

devinsag:
please see https://gitlab.com/architect-io/architect-cli/-/issues/472

### DIFF
--- a/src/base-command.ts
+++ b/src/base-command.ts
@@ -115,6 +115,7 @@ export default abstract class BaseCommand extends Command {
 
   async catch(error: any): Promise<void> {
     if (error.oclif && error.oclif.exit === 0) return;
+    process.exitCode = 1;
 
     try {
       if (error.stack) {


src/base-command.ts
```diff
@@ -115,6 +115,7 @@ export default abstract class BaseCommand extends Command {
 
   async catch(error: any): Promise<void> {
     if (error.oclif && error.oclif.exit === 0) return;
+    process.exitCode = 1;
```
tjhiggins, line 118: I still find setting the exitCode a little suspect. 

Is this because we are returning for the prettyValidationErrors?

```ts
if (error instanceof ValidationErrors) {
  return prettyValidationErrors(error);
}
```

tjhiggins (changes requested):

mueschm:
What is this for? The oclif/command class already does this. Can you add a test that would have failed before but passes now?

src/base-command.ts
devinsag: I think that is correct. we are skipping the super.catch call, (https://github.com/oclif/core/blob/main/src/command.ts#L253) and the oclif cleanup doesn't appear to happen during finally; terminating with exit code 0.
Other errors seem to exit with error codes fine. Perhaps it's because super.catch is otherwise always called. I can look into this further into suppressing the output from super.catch when called and/or move this into the conditional as an edge case;

tjhiggins: If its only for the pretty errors then the solution should be more targeted.

```ts
if (error instanceof ValidationErrors) {
  prettyValidationErrors(error);
  process.exit(1);
}
```

or potentially calling super.catch, but then the resulting error wouldn't be super pretty :P

devinsag: By adding the optional property exitCode to the validationErrors object, we can return super.catch({...error}), to get the best of both worlds.  pretty validation errors, and returning the super.catch() fn (generating the expected error exit code)

updated PR.

src/dependency-manager/utils/errors.ts
```diff
@@ -96,5 +97,6 @@ export class ValidationErrors extends ArchitectError {
 
     this.message = JSON.stringify(errors, null, 2);
     this.file = file;
+    this.exitCode = exitCode || 1;
```
tjhiggins, line 100: do we actually need to set this exitCode? I would assume the super.catch would always default to exitCode=1

src/base-command.ts
```diff
@@ -115,15 +115,15 @@ export default abstract class BaseCommand extends Command {
...
-    process.exitCode = 1;
 
     try {
       if (error.stack) {
         error.stack = [...new Set(error.stack.split('\n'))].join("\n");
       }
 
       if (error instanceof ValidationErrors) {
-        return prettyValidationErrors(error);
+        prettyValidationErrors(error);
```
tjhiggins, line 125: Can we add a test to the validation oclif command test? That way we ensure that pretty printing the validation errors will always exist 1. 

Also it would be good to ensure that the output is actually pretty and doesn't include a stack trace.

src/dependency-manager/utils/errors.ts
devinsag: I'll remove the exitCode property.  It will default to exitCode = 1 by calling the super.catch() even without the exitCode set.  The initial thought was to make it more flexible by allowing for other exitCodes if specified.

src/base-command.ts
devinsag: The output to console from super.catch is suppressed and does not print the stacktrace.  I'll write more test cases.

devinsag: added test cases to the validation oclif command test to test stacktrace is suppressed, exitCode is 1, and validation is pretty printed;

src/dependency-manager/utils/errors.ts
devinsag: reverted errors.ts

test/commands/validate.test.ts
```diff
@@ -7,6 +7,9 @@ describe('architect validate component', function () {
   // set to true while working on tests for easier debugging; otherwise oclif/test eats the stdout/stderr
   const print = false;
 
+  const cr = new RegExp(/\r/, 'g');
```
mueschm, line 10: What are these used for?

mueschm (changes requested):

test/commands/validate.test.ts
devinsag: oops, these were leftover from previous invalid tests. removed both regex constants.

PR #656, opened 2022-08-09
Fix issue with commands run in no-tty but with stdin available exiting prematurely due to stdin closing

TylerAldrich:
If the process's stdin is a TTY, even if --tty is not desired as part of the exec command itself, we can use `process.stdin.pipe(inputTransform).pipe(websocket);` to send input if it exists through the websocket. This is what happens during `architect exec /bin/sh` for example. 

If the process's stdin _isnt_ a TTY, which happens with a command like `architect exec -a test-tyler-0 -e example-environment -- cat < test-script.sh` where we're piping input to exec, it's possible for stdin to close before the websocket gets the data and passes back the response. If this happens, the command just doesn't work at all, so `{ end: false }` prevents the websocket from closing until the `outputTransform` eventually gets a success status.

The other issue was related to setting `tty=true` in the request - `kubectl` only does this if both `-it` are passed (stdin + tty), and we weren't doing the same - it was possible to run exec with `--no-stdin` which still set `tty=true` and would run into issues:
```
~/code/architect/architect-cli (rc*) » architect-local exec -a test-tyler-0 -e example-environment --no-stdin -- "sh -c \"export THING='yo'; echo 'hi'; npm run test\""
hi
```
Here, the export and echo work but `npm run test` wasn't running. After fixing this, you get the correct output:
```
~/code/architect/architect-cli (exec-stdin-tty-bugs*) » architect-local exec -a test-tyler-0 -e example-environment --no-stdin -- "sh -c \"export THING='yo'; echo 'hi'; npm run test\""
hi

> node-js-sample@0.1.0 test
> echo $THING

yo
```
where `--no-stdin` will now also send a command to kubernetes with `tty=false` just as `kubectl` does.

### DIFF
--- a/src/commands/exec.ts
+++ b/src/commands/exec.ts
@@ -85,22 +85,25 @@ export default class Exec extends BaseCommand {
     const ws = await this.getWebSocket(uri);
 
     await new Promise((resolve, reject) => {
-      const duplex = createWebSocketStream(ws, { encoding: 'utf-8' });
-      duplex.pipe(this.getOutputTransform());
+      const websocket = createWebSocketStream(ws, { encoding: 'utf-8' });
+      const inputTransform = this.getInputTransform();
+      websocket.pipe(this.getOutputTransform());
 
-      if (flags.stdin) {
-        if (flags.tty) {
-          // This method is only available when stdin is a TTY as it's part of the tty.ReadStream class:
-          // https://nodejs.org/api/tty.html#readstreamsetrawmodemode
-          process.stdin.setRawMode(true);
-        }
-        process.stdin.pipe(this.getInputTransform()).pipe(duplex);
+      if (process.stdin.isTTY) {
+        // This method is only available when stdin is a TTY as it's part of the tty.ReadStream class:
+        // https://nodejs.org/api/tty.html#readstreamsetrawmodemode
+        process.stdin.setRawMode(true);
+        process.stdin.pipe(inputTransform).pipe(websocket);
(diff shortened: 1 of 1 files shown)


src/commands/exec.ts
```diff
@@ -85,22 +85,25 @@ export default class Exec extends BaseCommand {
...
+
+      if (process.stdin.isTTY) {
+        // This method is only available when stdin is a TTY as it's part of the tty.ReadStream class:
+        // https://nodejs.org/api/tty.html#readstreamsetrawmodemode
+        process.stdin.setRawMode(true);
+        process.stdin.pipe(inputTransform).pipe(websocket);
+      } else if (flags.stdin) {
+        // If stdin is not a tty, stdin can be closed before data gets received by the websocket.
+        // Passing {end: false} keeps the websocket open until it's closed when we receive the result.
+        process.stdin.pipe(inputTransform, { end: false }).pipe(websocket);
```
tjhiggins, line 100: Is it possible for us to write a test for this? Would be nice to cover both cases of tty and no-tty. If its going to be too difficult we can add a tech debt issue.

TylerAldrich: It's maybe possible but very tricky because the tests in github actions run such that stdin isn't a tty, whereas running locally in your terminal via just `npm run test` does have a tty. So, it's probably not that hard to write a test that works locally for this, but not sure how valuable that is if it wouldn't work with the current way our tests are being run automatically on each branch

PR #658, opened 2022-08-10
473 list linked components

mtgoncalves1:
Closes https://gitlab.com/architect-io/architect-cli/-/issues/473  

#### List all linked components command:
```
% architect link:list    
```


src/commands/link.ts
```diff
@@ -3,42 +3,56 @@ import path from 'path';
...
 
   static flags = {
     ...BaseCommand.flags,
   };
 
   static args = [{
     non_sensitive: true,
-    name: 'componentPath',
-    char: 'p',
+    name: 'componentPathOrList',
```
ryan-cahill, line 23: Based on the conversation in standup, let's proceed with the ticket and do the following things:

* Create a folder in `src/commands/` called `link`
* Move `link.ts` to that new folder
* Create a new file called `index.ts` in the new folder for the linked components listing command
* Use the format `architect link:list` for the listing command

So when this is done, should have two files for the list command so that this one doesn't have to be overloaded. That should keep the code cleaner and hopefully make the usage clearer to the user since one command doesn't do two very different things

ryan-cahill (changes requested):

src/commands/link/index.ts
```diff
@@ -1,28 +1,40 @@
...
+
 declare const process: NodeJS.Process;
 
 export default class Link extends BaseCommand {
   async auth_required(): Promise<boolean> {
     return false;
   }
 
-  static description = 'Link a local component to the host to be used to power local deployments.';
+  static description = 'Link a local component to the host to be used to power local deployments or list all linked components.';
```
ryan-cahill, line 15: The descriptions in this file should be changed to only be relevant to the actual linking action, and a description should be added to the `link.ts` file to describe the listing action

src/commands/link/index.ts
```diff
@@ -1,28 +1,40 @@
...
 
   static flags = {
     ...BaseCommand.flags,
   };
 
   static args = [{
-    sensitive: false,
+    sensitive: true,
```
ryan-cahill, line 22: What makes this sensitive?

src/commands/link/index.ts
```diff
@@ -1,28 +1,40 @@
...
     name: 'componentPath',
-    char: 'p',
+    description: 'Provide a component path or list all linked components using "architect link list"',
     default: '.',
   }];
 
+  static non_sensitive = new Set([...Object.keys({ ...Link.flags }), ...Link.args.map(arg => arg.name)]);
+
+  listLinkedComponents(): void {
```
ryan-cahill, line 30: Doesn't look like this function is needed in this file anymore

test/commands/link.test.ts
```diff
@@ -63,4 +66,26 @@ describe('link', () => {
     const linked_components = fs.readJSONSync(linked_components_file);
     expect(linked_components).to.have.property('hello-world', component_path);
   });
+
+  it('list linked components', async () => {
```
ryan-cahill, line 70: It would be ideal to follow the same directory structure in the `tests` folder that we do for `src`. We have different test files for different commands such as `platforms`, `environments`, and `secrets`, so we should add one for `link` now too. Then each link command would have a different file in that folder

test/commands/link.test.ts
```diff
@@ -63,4 +66,26 @@ describe('link', () => {
...
+    const component_path = path.join(__dirname, '../../examples/hello-world/').replace(/\/$/gi, '').replace(/\\$/gi, '');
+
+    const log_spy = sinon.fake.returns(null);
+    sinon.replace(Link.prototype, 'log', log_spy);
+    await Link.run([component_path]);
```
ryan-cahill, line 75: I know this test is similar to the one above, but that's pretty outdated. New tests should run the command with `.command` like this one and follow a similar format

https://github.com/architect-team/architect-cli/blob/rc/test/commands/dev.test.ts#L511

ryan-cahill (changes requested):

test/commands/link.test.ts
mtgoncalves1: Sounds great! I'll update that

ryan-cahill (changes requested):

ryan-cahill (approved):

PR #662, opened 2022-08-12
no local selection when docker is unavailable

devinsag:
Please see: https://gitlab.com/architect-io/architect-cli/-/issues/486

### DIFF
--- a/src/architect/account/account.utils.ts
+++ b/src/architect/account/account.utils.ts
@@ -31,7 +31,7 @@ export default class AccountUtils {
     } catch (e: any) {
       console.log(chalk.yellow(`Unable to provide any local accounts to choose from because ${e?.message}`));
       if (exit_on_failure) {
-        throw e;
+        process.exit();
       }
       return false;
     }
@@ -84,7 +84,7 @@ export default class AccountUtils {
       }
     } else {
       inquirer.registerPrompt('autocomplete', require('inquirer-autocomplete-prompt'));
-      const can_run_local: boolean = await this.shouldListLocalAccounts();
+      const can_run_local = await this.shouldListLocalAccounts();
       const answers: { account: Account } = await inquirer.prompt([
         {
           type: 'autocomplete',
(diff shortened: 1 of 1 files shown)


src/architect/account/account.utils.ts
```diff
@@ -70,6 +84,7 @@ export default class AccountUtils {
       }
     } else {
       inquirer.registerPrompt('autocomplete', require('inquirer-autocomplete-prompt'));
+      const can_run_local: boolean = await this.shouldListLocalAccounts();
```
tjhiggins, line 87: You don't need to specify the type here because the function return specifies. 
```ts
const can_run_local = await this.shouldListLocalAccounts();
```

src/architect/account/account.utils.ts
```diff
@@ -49,7 +63,7 @@ export default class AccountUtils {
     const token_json = await app.auth.getPersistedTokenJSON();
     if (!token_json || token_json.email === 'unknown') {
       console.log(chalk.yellow('In order to access remote accounts you can login by running `architect login`'));
-      let account: Account;
+      await this.shouldListLocalAccounts(true);
```
tjhiggins, line 66: What does the output look like in this case? Won't it be kind of weird to show a warning log and then hard error?

tjhiggins (changes requested):

src/architect/account/account.utils.ts
devinsag: removed specified type from can_run_local

src/architect/account/account.utils.ts
devinsag: removed hard-error, and replaced with warning and then exit if neither local or remote accounts are available.

src/architect/account/account.utils.ts
```diff
@@ -70,6 +84,7 @@ export default class AccountUtils {
       }
     } else {
       inquirer.registerPrompt('autocomplete', require('inquirer-autocomplete-prompt'));
+      const can_run_local = await this.shouldListLocalAccounts();
```
TylerAldrich, line 87: I'm not 100% sure if in this case we want to log a warning message to users. Example scenario here:
image

I think for the most part, if I don't have docker running, it's not that surprising to not have the local option and this warning message is just kind of superfluous. For a brand new user that maybe doesn't have docker running and would get confused, they probably don't have any remote accounts yet and would hit the hard-fail state above.

I'm leaning towards not logging anything if we're not hard-exiting, curious if anyone has thoughts

PR #663, opened 2022-08-15
482 pass replica name

mtgoncalves1:
Closes issue - https://gitlab.com/architect-io/architect-cli/-/issues/482.  
```
One service:
% architect exec -a my-acc -e example-environment -r 0 -- ls   

One or more services:
% architect exec -a my-acc -e example-environment -r app:0 -- ls   
```


mtgoncalves1:
> @mtgoncalves1 looks like there are conflicts now

Ok. I will fix the conflicts.

src/commands/exec.ts
```diff
@@ -57,6 +57,11 @@ export default class Exec extends BaseCommand {
       default: undefined,
       sensitive: false,
     }),
+    replica: Flags.string({
```
tjhiggins, line 60: Can we have this take a pod name or a number?

Sometimes I just want to get the first replica and I don't care about a specific one.

```
architect exec -r 0
```

@TylerAldrich thoughts?

Might not be the right format, but in a CI system I won't know the name of a pod.

TylerAldrich: I do think we should do something like this - we probably need to at least pass the service name because `architect exec -a .. -e .. -r 0` wouldn't work if you have >1 services with >1 replicas.

If you have services called `foo` and `bar` it would be pretty slick IMO to be able to do `-r foo:0` or `-r bar:1` and exec into the correct service+pod

TylerAldrich: We probably could handle `-r 0` if you only have 1 service or only have 1 service with >1 replica because we could deduce which one you mean, and otherwise complain that `-r 0` is too ambiguous - I'm not sure if that adds a ton of complexity but may be the best possible UX because for a lot of simple cases `-r 0` is good enough and you don't even need to write out the service name

mtgoncalves1: Sounds good. I'll work on that!

mtgoncalves1: I got the `-r foo:0` working but still need to handle `-r 0` for the case where there is only 1 service.

mtgoncalves1:
Hi @tjhiggins, the PR is now ready for review. Thanks!!

src/commands/exec.ts
```diff
@@ -253,13 +260,58 @@ export default class Exec extends BaseCommand {
...
+    let service_replicas = [];
+    let replica_idx;
+    if (new RegExp('^[0-9]+$').test(replica_flag)) {
+      if (resource_names.length > 1) {
+        throw new ArchitectError(`More than one service is found [${resource_names.join(', ')}]. Please specify the replica in the form of <service-name>:<replica-index>.`);
+      }
+
+      service_replicas = replicas;
+      replica_idx = parseInt(replica_flag);
+    } else if (new RegExp('^[a-z]+[a-zA-Z0-9_-]+:[0-9]+$').test(replica_flag)) {
```
TylerAldrich, line 284: This can be simplified a little and not require using regular expressions to check the input
```typescript
replica_parts = replica_flag.split(':');
if (replica_parts.length === 1 && !isNaN(replica_parts[0])) {
    // same code currently there, replica_idx = parseInt(replica_parts[0]) is okay
} else if (replica_parts.length == 2) {
    // same code you've got - already did the split
    const [service_name, service_replica_idx] = replica_parts;
}
```

The other error handling you've got for making sure the service name exists and that the replica_idx is within the bounds of available replicas should handle any weird inputs somebody wants to throw

test/commands/exec.test.ts
```diff
@@ -40,7 +40,21 @@ describe('exec', () => {
...
+      expect(ctx.stdout).to.equal('worked\n');
+    });
+  
+  defaults
+    .stdout()
+    .command(['exec', '-a', account.name, '-e', environment.name, '-r', '0', '--', 'ls', '-la'])
+    .it('exec component and replica of the form <replica-index> when there is only one service', ctx => {
+      expect(ctx.stdout).to.equal('worked\n');
+    });
+
```
TylerAldrich, line 56: We should throw a couple negative tests in here too that make sure the invalid inputs give us errors (like `-r -1` or `-r foo:`)

src/commands/exec.ts
```diff
@@ -57,6 +59,11 @@ export default class Exec extends BaseCommand {
       default: undefined,
       sensitive: false,
     }),
+    replica: Flags.string({
+      description: `Pass replica by <service-name>:<replica-index> or <replica-index> if only 1 service deployed. Only works on remote deploys.`,
```
TylerAldrich, line 63: minor grammar thing but I'd change this to "if only 1 service *is* deployed"

TylerAldrich (changes requested):

src/commands/exec.ts
TylerAldrich: (The main advantage to avoiding a regex on the service name especially is that if we changed the regex for what is allowed in our architect.yml schema, we can very easily forget to change this code - so just accepting any string before the `:` and just checking against our set of service names avoids that potential future bug)

mtgoncalves1: That is a good point. I'll fix that.

TylerAldrich (approved):

PR #664, opened 2022-08-15
updating microservices demo

ryan-cahill:


mueschm:
The last time this project was touched was 7 months ago. Did we test this to make sure its good? Why are we updating it now? Context?

ryan-cahill:
That's related to this ticket: https://gitlab.com/architect-io/architect-cli/-/issues/461. I wanted to see if Mary could run the demo services on their most recent versions, but the issues are still there. Regardless, I tested this both locally and remotely and the services still look good on my non-M1 machine, so this isn't a regression

PR #665, opened 2022-08-15
Handle multiple arguments after -- without requiring quotes

TylerAldrich:
- Enforced `--` is used when exec'ing. Our `--help` usage says that this is required and it makes things less confusing, so it is now enforced
- Keep track of the raw argv array of commands so that we can pass that directly to docker/kubectl. 
  - Without keeping track of this, it's impossible to figure out what parts were quoted in the original command (e.g. `sh -c "echo foo"`  would look the same as `sh -c "echo foo"` which the user probably wanted because we join the cmd into a string, and the shell strips the quotes).

Works on remote via kubectl:
```
~/code/architect/architect-cli (exec-multi-command-args*) » architect exec -- sh -c 'echo 1; ls'
? Select an account tyler-aldrich
? Select a environment example-environment
1
Dockerfile         node_modules       package.json
architect.yml      package-lock.json  src
```

And locally via docker:
```
~/code/architect/architect-cli (exec-multi-command-args*) » architect exec -- sh -c 'echo 1; ls'
? Select an account dev (Local Machine)
? Select a service hellow-world.services.my-app-2
1
Dockerfile         node_modules       package.json
architect.yml      package-lock.json  src
```

### DIFF
--- a/src/base-command.ts
+++ b/src/base-command.ts
@@ -77,10 +77,11 @@ export default abstract class BaseCommand extends Command {
 
     // Support -- input ex. architect exec -- ls -la
     const double_dash_index = argv.indexOf('--');
+    let command: string[] = [];
     if (double_dash_index >= 0) {
-      const command = argv.slice(double_dash_index + 1, argv.length).join(' ');
+      command = argv.slice(double_dash_index + 1, argv.length);
       argv = argv.slice(0, double_dash_index);
-      argv.unshift(command);
+      argv.unshift(command.join(' '));
     }
 
     const args = [];
@@ -103,7 +104,11 @@ export default abstract class BaseCommand extends Command {
       }
     }
 
-    return super.parse(options, [...args, ...flags]);
+    const parsed = await super.parse(options, [...args, ...flags]) as Interfaces.ParserOutput<F, A>;
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    parsed.args.command_array = command;
+    return parsed;
   }
 
   async finally(err: Error | undefined): Promise<any> {
--- a/src/commands/exec.ts
+++ b/src/commands/exec.ts
@@ -3,7 +3,6 @@ import { OutputArgs, OutputFlags } from '@oclif/core/lib/interfaces';
 import chalk from 'chalk';
 import inquirer from 'inquirer';
 import stream from 'stream';
-import stringArgv from 'string-argv';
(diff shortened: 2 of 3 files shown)


src/base-command.ts
```diff
@@ -77,10 +77,11 @@ export default abstract class BaseCommand extends Command {
 
     // Support -- input ex. architect exec -- ls -la
     const double_dash_index = argv.indexOf('--');
+    let command: string[] = [];
     if (double_dash_index >= 0) {
-      const command = argv.slice(double_dash_index + 1, argv.length).join(' ');
+      command = argv.slice(double_dash_index + 1, argv.length);
```
tjhiggins, line 82: would the following work? 

```ts
argv.slice(double_dash_index + 1, argv.length).map((arg) => `'${arg}'`).join(' ');
```

tjhiggins (changes requested):

src/base-command.ts
TylerAldrich: One potential issue with doing that is;

`architect exec -- sh -c "echo \"hi\"; echo 2"` is a valid input, which yields the command  `'sh' '-c' 'echo 'hi'; echo 2'`. 

If you pass that to `stringArgv` you get `[ 'sh', '-c', 'echo ', "hi'; echo 2'" ]` which is wrong, what we want is `[ 'sh', '-c', 'echo', "hi'; echo2" ]`.

I think most arbitrary ways of joining the command into a string can run into similar problems where some input exists that makes it impossible for us to go from array -> string -> array without guessing, unless we chose the poop emoji and then banned anyone from using that in the command 💩

src/base-command.ts
```diff
@@ -77,10 +77,11 @@ export default abstract class BaseCommand extends Command {
...
       argv = argv.slice(0, double_dash_index);
-      argv.unshift(command);
+      argv.unshift(command.join(' '));
```
tjhiggins, line 84: Should we still be unshifting? If we do this then maybe we don't need command to be an arg.

TylerAldrich: That's a good q. I did add enforcement for `-- COMMAND` since previously you could omit `--` for a single command like `ls`, but even our examples have `architect exec ... -- ls`. With the forced-double-dash we could remove COMMAND as an arg and remove this logic.

If this is the only command that supports `--` as well, can we remove this BaseCommand.parse entirely? The logic becomes much simpler and we can throw it into a much smaller custom parse for Exec only

PR #666, opened 2022-08-15
Oclif command examples

devinsag:
please see https://gitlab.com/architect-io/architect-cli/-/issues/481


src/commands/dev.ts
```diff
@@ -23,6 +23,11 @@ export default class Dev extends BaseCommand {
 
   static description = 'Run your stack locally';
 
+  static examples = [
+    'architect dev ./myComponent/architect.yml',
+    'architect dev --port=81 --browser --debug --secret-file=./myComponent/mySecrets.yml ./myComponent/architect.yml',
```
mueschm, line 28: Browser is set to true by default. This should be `--no-browser` to show a change in functionality.

src/commands/platforms/destroy.ts
```diff
@@ -8,7 +8,10 @@ import BaseCommand from '../../base-command';
 export default class PlatformDestroy extends BaseCommand {
   static aliases = ['platforms:deregister', 'platform:destroy', 'platforms:destroy'];
   static description = 'Deregister a platform from Architect';
-
+  static examples = [
+    'architect platform:destroy --account=myAccount architect',
+    'architect platforms:deregister --account=myAccount --type=kubernetes --auto-approve --force architect',
```
mueschm, line 13: There does not seem to be a flag `--type=kubernetes` for this command.

mueschm (changes requested):

src/commands/dev.ts
devinsag: updated to use --no-browser

src/commands/platforms/destroy.ts
devinsag: removed --type from platforms:deregister examples

src/commands/components/index.ts
```diff
@@ -17,7 +17,11 @@ interface Component {
 export default class Components extends BaseCommand {
   static aliases = ['components', 'components:search', 'component:search', 'component:search'];
   static description = 'Search components you have access to';
-
+  static examples = [
+    'architect components',
+    'architect components --account=myAccount',
```
tjhiggins, line 22: Account names need to be lowercase - have this match our regex. Same with component.

```
architect components --account=myaccount
```

src/commands/components/versions.ts
```diff
@@ -12,7 +12,10 @@ interface ComponentVersion {
 export default class ComponentVersions extends BaseCommand {
   static aliases = ['component:versions', 'component:version'];
   static description = 'Search component versions of a particular component';
-
+  static examples = [
+    'architect component:versions myComponentName',
```
tjhiggins, line 16: lets be consistent in naming `mycomponent` or `mycomponentname`

tjhiggins (changes requested):

src/commands/components/index.ts
devinsag: renamed all example camelcase values to lowercase

src/commands/components/versions.ts
devinsag: renamed all example values {value}Name to {value} to be consistent

PR #668, opened 2022-08-15
Prompt user for available port

mtgoncalves1:
Closes issue: https://gitlab.com/architect-io/architect-cli/-/issues/485.  
Screen Shot 2022-08-15 at 4 20 43 PM

### DIFF
--- a/src/commands/dev.ts
+++ b/src/commands/dev.ts
@@ -321,6 +321,27 @@ export default class Dev extends BaseCommand {
     process.exit();
   }
 
+  private async getAvailablePort(port: number): Promise<number> {
+    while (!(await PortUtil.isPortAvailable(port))) {
+      const answers: any = await inquirer.prompt([
+        {
+          type: 'input',
+          name: 'port',
+          message: `Trying to listen on port ${port}, but something is already using it. What port would you like us to run the API gateway on (you can use the '--port' flag to skip this message in the future)?`,
+          validate: (value: any) => {
+            if (new RegExp('^[1-9]+[0-9]*$').test(value)) {
+              return true;
+            }
+            return `Port can only be positive number.`;
+          },
+        },
+      ]);
+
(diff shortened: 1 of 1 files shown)


src/commands/dev.ts
```diff
@@ -353,9 +353,24 @@ export default class Dev extends BaseCommand {
       linked_components
     );
 
-    const port_available = await PortUtil.isPortAvailable(flags.port);
-    if (!port_available) {
-      this.error(`Could not run architect on port ${flags.port}.\nPlease stop an existing process or specify --port to choose a different port.`);
+    let port_available = await PortUtil.isPortAvailable(flags.port);
+    while (!port_available) {
```
mueschm, line 357: You could just turn this into
```
while (!(await PortUtil.isPortAvailable(flags.port)) {
```
then you don't need the variable at all and reduce the mutable state of the application.

src/commands/dev.ts
```diff
@@ -353,9 +353,24 @@ export default class Dev extends BaseCommand {
       linked_components
     );
 
-    const port_available = await PortUtil.isPortAvailable(flags.port);
-    if (!port_available) {
-      this.error(`Could not run architect on port ${flags.port}.\nPlease stop an existing process or specify --port to choose a different port.`);
+    let port_available = await PortUtil.isPortAvailable(flags.port);
```
mueschm, line 356: Can we move this to its own function that just gets the available port?

This would make it easier to read and also would produce something like
```
const port = getAvailablePort(flags.port)
```
Now we just use port for the rest of the function and its immutable. Would also keeps our flags less mutable which is generally a good thing.

mueschm (changes requested):

src/commands/dev.ts
mtgoncalves1: Yes, makes sense.

PR #671, opened 2022-08-16
Fix exec hanging after it processes too much data

TylerAldrich:
This was happening because we piped the websocket output to a `stream.Transform`, which is a `Duplex` stream that is readable and writable. We never actually read from this stream, which meant the read buffer eventually fills up. This causes writes to accumulate in a buffer until the stream is read from -- which never happens.

Turning this into a `stream.Writable` fixes this issue - we never need to read anything from this stream so now we're not storing anything in a useless buffer and can process all the data.

### DIFF
--- a/src/commands/exec.ts
+++ b/src/commands/exec.ts
@@ -100,7 +100,7 @@ export default class Exec extends BaseCommand {
     await new Promise((resolve, reject) => {
       const websocket = createWebSocketStream(ws, { encoding: 'utf-8' });
       const inputTransform = this.getInputTransform();
-      websocket.pipe(this.getOutputTransform());
+      websocket.pipe(this.getOutputStream());
 
       if (process.stdin.isTTY) {
         // This method is only available when stdin is a TTY as it's part of the tty.ReadStream class:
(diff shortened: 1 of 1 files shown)


src/commands/exec.ts
```diff
@@ -189,15 +189,18 @@ export default class Exec extends BaseCommand {
             process.exit(error_code);
           }
         } else {
-          return done(new ArchitectError(`Unknown stream type: ${stream_num}`));
+          // If we get an unrecognized stream number, continue outputting to stdout.
+          // This can happen when writing binary data, and kubectl seems to handle this by just
+          // writing anyways so we do the same.
+          process.stdout.write(buffer);
```
TylerAldrich, line 195: This was a bit of a bonus discovery - I was testing with massive binary files to reproduce this (e.g. `head -c10000000 /dev/urandom > junk.txt`) and `architect exec ... -- cat junk.txt` would fail for some of these files due to `Unknown stream type`. What was happening is we were getting a unicode replacement character `0xef 0xbf 0xbd` followed by `0x01` (expected for stdout) and the rest of the line. Not exactly sure what the deal was with that, but kubectl has no problem cat-ing this entire file, so writing the buffer to stdout even if it's invalid seems to mimic kubectl behavior

PR #675, opened 2022-08-17
Handle terminal resize events in exec

TylerAldrich:
Tested this on mac, windows (powershell + cmd), and in Linux via wsl2 and on each system resizes are working correctly. The initial size of the window is also correct now 🥳.

Fixes issues with programs like vim, less, etc. that require knowledge of the terminal size in order to draw themselves correctly, and without passing a size explicitly it would fall back to a default size and provide a weird experience.

Note that kubernetes on the server side handles backwards compatibility - if somebody is running a (very very) old cluster that doesn't support resize events, this just gets ignored. The kubectl client always sends this data regardless, so we should be safe to do the same.

### DIFF
--- a/src/commands/exec.ts
+++ b/src/commands/exec.ts
@@ -67,10 +67,12 @@ export default class Exec extends BaseCommand {
     parse: async (value: string): Promise<string> => value.toLowerCase(),
   }];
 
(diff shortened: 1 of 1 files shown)


tjhiggins:
LGTM 🧇

PR #676, opened 2022-08-17
Use local SSL for 

mueschm:
This adds support for SSL certificates for local development.

## Changes
1. Script to download certificates from a remote bucket. If unable to download them then we cache them for 30 days.
2. Update to traefik to support SSL.
3. Flag to disable SSL

## Testing
1. Test with ssl `architect dev ./examples/react/`
2. Test without ssl `architect dev ./examples/react/ --no-ssl`
3. Test complex applications like ones made by the the project command that use ingresses from other components.
image

### DIFF
--- a/src/commands/dev.ts
+++ b/src/commands/dev.ts
@@ -227,7 +227,7 @@ export default class Dev extends BaseCommand {
     for (const svc_name of Object.keys(compose.services)) {
       for (const port_pair of compose.services[svc_name].ports || []) {
         const [exposed_port, internal_port] = port_pair && (port_pair as string).split(':');
-        this.log(`${chalk.blue(`http://localhost:${exposed_port}/`)} => ${svc_name}:${internal_port}`);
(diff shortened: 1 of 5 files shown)


tjhiggins (changes requested):

src/commands/dev.ts
```diff
@@ -378,6 +386,53 @@ export default class Dev extends BaseCommand {
...
+        });
+      }).catch(err => {
+        return handleReject(resolve, reject);
+      });
+    });
+  }
+
+  private async downloadSSLCerts() {
+    return Promise.all([
+      this.downloadFile('https://storage.googleapis.com/architect-ci-ssl/fullchain.pem', path.join(this.app.config.getConfigDir(), 'fullchain.pem')),
```
tjhiggins, line 428: Does this work correctly on windows? Sometimes you need to call `path.resolve` to get the correct path for the os. Maybe path.join does the same thing.

mueschm: Just tested on Windows and it works as expected. From my understanding `path.join` is supposed to be platform agnostic.

PR #677, opened 2022-08-18
Dev command leaves containers running when process exits with an error

TylerAldrich:
When we hit this code:

```typescript
try {
    await compose_process;
} catch (ex) {
    if (!is_exiting) {
        throw ex;
    }
}
```

it was possible for the `compose_process` to end (with an error that we throw) - but the process finishing with an error doesn't guarantee that containers get cleaned up. One container may hard-fail, the process ends, and other containers remain running.

To fix that, we now run:
```typescript
    try {
      await this.compose_process;
    } catch (ex) {
      if (!this.is_exiting) {
        // Always call `docker compose stop` here - the process died so there's nothing to kill,
        // need to call `docker compose stop` to clean up any remaining running containers.
        console.error(chalk.red(ex));
        console.log(chalk.red('\nError detected - Stopping running containers...'));
        await this.stop();
      }
    }
```

to call `docker compose stop` after the `compose up` process ends to ensure that containers are stopped. 

I also moved all the code that manages the `docker compose up` process into it's own class to make the logic in this file more easily understandable.
- The `Dev` command class is still handling most of the work
- The `UpProcessManager` now starts/registers the process interrupts/stops the long-running `docker compose up` process and when it's finished running, the `Dev.run` can continue cleanup.

Tested on 
- [x] Mac
- [x] WSL2
- [x] (Windows) Powershell
- [x] (Windows) Cmd

### DIFF
--- a/src/commands/dev.ts
+++ b/src/commands/dev.ts
@@ -1,6 +1,7 @@
 import { Flags, Interfaces } from '@oclif/core';
 import axios from 'axios';
 import chalk from 'chalk';
+import { ExecaChildProcess } from 'execa';
 import fs from 'fs-extra';
 import inquirer from 'inquirer';
 import isCi from 'is-ci';
@@ -218,7 +219,7 @@ export default class Dev extends BaseCommand {
     }, poll_interval);
   }
 
-  async runCompose(compose: DockerComposeTemplate, gateway_port: number, gateway_admin_port: number): Promise<void> {
+  async buildImage(compose: DockerComposeTemplate): Promise<[string, string]> {
     const { flags } = await this.parse(Dev);
 
     const project_name = flags.environment || DockerComposeUtils.DEFAULT_PROJECT;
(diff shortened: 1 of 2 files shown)


src/commands/dev.ts
```diff
@@ -190,6 +327,37 @@ export default class Dev extends BaseCommand {
     return healthy_services;
   }
 
+  setupTraefikServiceMap(compose: DockerComposeTemplate, gateway_port: number): Dictionary<string | undefined> {
```
tjhiggins, line 330: why is this showing up in the diff?

TylerAldrich: I just moved this into its own function similar to `pollTraefik` to make the various parts of `runCompose` easier to understand

tjhiggins: ah thought this was the new funcs I added - ignore

PR #678, opened 2022-08-19
fixing healthcheck liveness probe protocol for port/path

ryan-cahill:
Undoes the change [here](https://github.com/architect-team/architect-cli/pull/676/files#diff-c501db4ece42dfe6ff2e5e68da3e74b84a44bb1e6e14bbb53fc920376a11bc0cR183) as a liveness probe shouldn't be using HTTPS

### DIFF
--- a/src/common/docker-compose/index.ts
+++ b/src/common/docker-compose/index.ts
@@ -61,7 +61,6 @@ export class DockerComposeUtils {
       volumes: {},
     };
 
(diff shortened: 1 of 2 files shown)


tjhiggins:
Good catch

tjhiggins (approved):

tjhiggins:
Lemme know when its not WIP

PR #679, opened 2022-08-22
Docker verify improved

TylerAldrich:
Instead of having `await Docker.verify()` (and similar functions for checking compose/buildx) deep within functions right before we attempted to use them, the checks for various features required has been moved to the `RequiresDocker()` decorator. This solves a few problems:

- Calling `Docker.verify()` right before we exec some docker command means in some cases we end up doing some work before hitting the exception which wastes the users time. With the decorator, we fail as soon as a `run` or `runLocal` function that requires Docker features is called
- The `RequiresDocker` decorator uses a `_DockerHelper` class that figures out what docker features are available at the start of the command. The same results are re-used in the case of a command calling a different command, and we no longer end up in situations where we're running `Docker.verify()` multiple times in some commands like dev or exec
- We are now using` docker info` to figure out if the daemon is running and if compose/buildx exist. Instead of having 4 different checks for various docker features, we have 2. The first is checking if the docker command exists at all, the 2nd is running `docker info` to get all the info we need

I also did a minor refactor to the various docker utilities we had by moving them to a `/common/docker` folder.


src/common/docker/helper.ts
```diff
@@ -0,0 +1,180 @@
...
+    const plugins: Set<string> = new Set();
+    if (plugin_location >= 0) {
+      for (let i = plugin_location + 1; i < docker_info_lines.length; i++) {
+        if (!docker_info_lines[i].startsWith('  ')) {
+          break;
+        }
+        const plugin_name = docker_info_lines[i].trimLeft().split(':')[0];
+        plugins.add(plugin_name);
+      }
+    }
```
TylerAldrich, line 110: Use `docker info --format '{{json .}}'` and `JSON.parse` the output instead to make this way less sketch

TylerAldrich: done, thx for your sage widsom

src/common/docker/cmd.ts
```diff
@@ -0,0 +1,17 @@
+import execa, { Options } from 'execa';
+
+export const docker = async (args: string[], opts = { stdout: true }, execa_opts?: Options): Promise<any> => {
```
tjhiggins, line 3: Should we still have the decorator around the execa methods on the off chance we forget to put in the command?

TylerAldrich: Yeah no harm in dong that and a good idea to have a backup just in case

PR #684, opened 2022-08-23
Enable agent on cli

mueschm:
Add support for the agent to dev and local versions of the CLI. This code modifies the `platform:create` function to allow the agent to be installed as well as still provide the older method of creating a service account whos credentials are used directly.

## Changes
1. Add support for the agent type to the system.
2. Create a service account on the desired cluster
3. Handle the new token required for platforms
4. Refactored some of the logic to be more generic between platforms
5. Small bits of logic have been duplicated to make it easier to deprecate the older method later
6. Updated unit tests + a new one

## Testing
Basic tests `run npm test`. Everything works as intended.
Created an agent platform and installed the applications on a DO cluster.
Created a kuberenetes platform and installed the applications on a DO cluster.


src/commands/platforms/create.ts
```diff
@@ -99,37 +129,34 @@ export default class PlatformCreate extends BaseCommand {
 
     const account = await AccountUtils.getAccount(this.app, flags.account, { account_message: 'Select an account to register the platform with' });
 
-    const platform = await this.createArchitectPlatform(flags);
+    const contexts = await this.setupKubeContext(flags);
```
ryan-cahill, line 132: `contexts` is a pretty vague variable name especially without a named/specified type. Would prefer `kube_contexts` or something

src/commands/platforms/create.ts
```diff
@@ -99,37 +129,34 @@ export default class PlatformCreate extends BaseCommand {
...
 
-    const platform_dto = { name: platform_name, ...platform, flags: flags_map };
+    try {
+      const platform = await this.createArchitectPlatform(flags, contexts.current_context);
```
ryan-cahill, line 135: As this isn't a `Platform` and is `CreatePlatformInput`, naming this `create_platform_input` would be more helpful

src/commands/platforms/create.ts
```diff
@@ -99,37 +129,34 @@ export default class PlatformCreate extends BaseCommand {
...
-      const confirmation = await inquirer.prompt({
-        type: 'confirm',
-        name: 'application_install',
-        message: `Would you like to install the requisite networking applications? This is a required step before using Architect with this platform. More details at the above URL.`,
-      });
-      if (!confirmation.application_install) {
-        this.warn(`Installation cancelled. You will be unable to deploy services to this platform.\n\nIf you decide to proceed with installation, you can do so at the above URL. Or if you would like to deregister this platform from Architect, run: \n\narchitect platform:destroy -a ${account.name} --auto_approve ${platform_name}`);
-        return;
+      CliUx.ux.action.start('Registering platform with Architect');
+      const created_platform = await this.postPlatformToApi(platform_dto, account.id);
```
ryan-cahill, line 140: `created_platform` has a type `any`, but should be type `Platform`

src/commands/platforms/create.ts
```diff
@@ -99,37 +129,34 @@ export default class PlatformCreate extends BaseCommand {
...
+      this.log(`Platform registered: ${this.app.config.app_host}/${account.name}/platforms/new?platform_id=${created_platform.id}`);
+
+      if (flags.type?.toLowerCase() == 'agent') {
+        CliUx.ux.action.start(chalk.blue('Installing the agent'));
+        await AgentPlatformUtils.installAgent(flags, created_platform.token.access_token, AgentPlatformUtils.getServerAgentHost(this.app.config.agent_server_host), this.app.config);
+        await AgentPlatformUtils.waitForAgent(flags);
+        CliUx.ux.action.stop();
+        await this.installAppliations(flags, created_platform, account.name, platform_name);
+      } else {
+        await this.installAppliations(flags, created_platform, account.name, platform_name);
```
ryan-cahill, line 151: Seeing as this line is duplicated, I don't see why the `else` can't be removed. Then this line can just live under the `if` portion

src/commands/platforms/create.ts
```diff
@@ -138,15 +165,23 @@ export default class PlatformCreate extends BaseCommand {
         message: 'What type of platform would you like to register?',
         choices: [
           'kubernetes',
+          ...(this.app.config.environment !== ENVIRONMENT.PRODUCTION ? ['agent (BETA)'] : []),
         ],
       },
     ]);
+    if (!flags.type && platform_type_answers.platform_type.indexOf('agent') === 0) {
```
ryan-cahill, line 172: Rather than searching with `indexOf, why not define the string above the prompt, then just check for equality with that?

src/common/docker-compose/index.ts
```diff
@@ -183,6 +183,7 @@ export class DockerComposeUtils {
           } else {
             liveness_probe.command = ['CMD', ...liveness_probe.command];
           }
+          console.log(liveness_probe.command);
```
ryan-cahill, line 186: :thinking:

src/common/utils/agent-platform.utils.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  private static getLocalServerAgentPort(): string {
+    const container_name_results = execa.sync('docker', ['ps', '-f', 'name=agent-server', '--format', '{{.Names}}']);
+    const results = execa.sync('docker', ['port', container_name_results.stdout, '9081/tcp']);
+    return results.stdout.split(':')[1];
+  }
+
+  public static getServerAgentHost(agent_server_host: string): string {
+    const host = agent_server_host.toLocaleLowerCase().trim();
+    if (host[host.length - 1] === ':') {
```
ryan-cahill, line 26: `host.endsWith(':')`

src/common/utils/agent-platform.utils.ts
```diff
@@ -0,0 +1,223 @@
...
+          '--serviceaccount',
+          `default:${SERVICE_ACCOUNT_NAME}`,
+        ]);
+      }
+    }
+
+    if (!use_existing_sa) {
+      CliUx.ux.action.start('Creating the service account');
+      await AgentPlatformUtils.createKubernetesServiceAccount(untildify(kubeconfig_path), SERVICE_ACCOUNT_NAME);
+      const secret_yml = `
```
ryan-cahill, line 91: Might be nice to add this to something like a generic `PlatformUtils` so that this isn't duplicated and we can update both platforms in the future if we need to. There are some other duplicated things in this file that could be added to that generic util file too

test/commands/platforms/create.test.ts
```diff
@@ -73,10 +82,61 @@ describe('platform:create', function () {
...
 
-    await PlatformCreate.run(['platform-name', '-a', 'test-account-name', '-t', 'kubernetes', '--auto-approve']);
+    await PlatformCreate.run(['platform-name', '-a', 'test-account-name', '-t', 'kubernetes']);
     expect(create_platform_spy.calledOnce).true;
     expect(post_to_api_spy.calledOnce).true;
     expect(kubernetes_configuration_fake.calledOnce).true;
     expect(create_platform_applications_spy.calledOnce).true;
   });
+
+  it('Creates an Agent platform with input', async () => {
```
ryan-cahill, line 92: I know the rest of this file is written this way, but this should use the newer format of writing tests rather than the old format. Personally I think it's way more readable

ex. https://github.com/architect-team/architect-cli/blob/rc/test/commands/dev.test.ts#L676

ryan-cahill (changes requested):

src/common/utils/agent-platform.utils.ts
mueschm: I debated this but because I plan on removing the old logic this seemed fine and kept everything separate without adding a layer of abstraction that was not needed in the future.

ryan-cahill (approved):

src/app-config/config.ts
```diff
@@ -39,6 +40,7 @@ export default class AppConfig {
     this.oauth_host = 'https://auth.architect.io';
     this.oauth_client_id = '079Kw3UOB5d2P6yZlyczP9jMNNq8ixds';
     this.account = '';
+    this.agent_server_host = 'https://agent-client.dev.architect.io';
```
tjhiggins, line 43: this should default to `https://agent-client.architect.io`

src/commands/platforms/create.ts
```diff
@@ -143,15 +171,24 @@ export default class PlatformCreate extends BaseCommand {
         message: 'What type of platform would you like to register?',
         choices: [
           'kubernetes',
+          agent_display_name,
+          //...(this.app.config.environment !== ENVIRONMENT.PRODUCTION ? [agent_display_name] : []),
```
tjhiggins, line 175: remove?

src/common/utils/agent-platform.utils.ts
```diff
@@ -0,0 +1,195 @@
...
+    spec:
+      serviceAccountName: ${SERVICE_ACCOUNT_NAME}
+      containers:
+        - name: agent
+          image: registry.gitlab.com/architect-io/agent/client/client:latest
+          env:
+          - name: KUBERNETES_URL
+            value: "https://kubernetes.default.svc"
+          - name: ARCHITECT_TOKEN
+            value: "${token}"
```
tjhiggins, line 148: shouldn't this also be in a secret?

tjhiggins (changes requested):

src/commands/platforms/create.ts
mueschm: This is temporary since the idea is to hide it in prod for now. The only reason I did this was for when people pulled the branch tomorrow for testing.

TylerAldrich (approved):
LGTM 👍 

Not sure what the status is of some of the outstanding TODO/comments that Ryan/TJ commented on previously but everything else looks good

PR #685, opened 2022-08-24
convert deployCommand auto-approve

devinsag:
Fixing the last two flags I missed in the original pr: https://github.com/architect-team/architect-cli/pull/683 to complete the refactor.

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -23,13 +23,13 @@ export abstract class DeployCommand extends BaseCommand {
       description: `${BaseCommand.DEPRECATED} Please use --auto-approve.`,
       hidden: true,
       sensitive: false,
-      default: undefined,
+      default: false,
     }),
     'auto-approve': booleanString({
       exclusive: ['compose-file', 'compose_file'],
       description: 'Automatically approve the deployment without a review step. Used for debugging and CI flows.',
       sensitive: false,
-      default: undefined,
+      default: false,
     }),
   };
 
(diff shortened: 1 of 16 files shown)


src/commands/deploy.ts
```diff
@@ -18,16 +18,18 @@ export abstract class DeployCommand extends BaseCommand {
 
   static flags = {
     ...BaseCommand.flags,
-    auto_approve: Flags.boolean({
+    auto_approve: booleanString({
       exclusive: ['compose-file', 'compose_file'],
       description: `${BaseCommand.DEPRECATED} Please use --auto-approve.`,
       hidden: true,
       sensitive: false,
+      default: undefined,
```
tjhiggins, line 26: why is this undefined?

tjhiggins (changes requested):

src/commands/deploy.ts
devinsag: booleanString requires a default fallback value to support --flag_name usage, triggering the deprecated messages even if the user had not passed the deprecated flag (failing tests).  Because the deprecated flag has been directly replaced, we set the default value (false) on the active flag and undefined on the deprecated command.  This way,  it does not trigger the deprecated warning unless --auto_approve is specified and present in argv. 

- if neither deprecated or active flag are specified, the active flag will assume the default value (false). 
- if deprecated flag alone is specified, the active flag is assumed the deprecated value and a warning is produced.
- if active flag alone is specified, the active flag will be assigned the passed value.
- if both flags are specified, the active flag will assume the value passed to the active flag (added a check in next commit)

PR #687, opened 2022-08-25
adding decorator to check that git/docker are installed for project init

ryan-cahill:
## Overview
Added checks to project creation function to check that Docker/Compose and Git are all installed before the function/user can proceed

## Changes
* Added `RequiresGit` decorator to check that `git` is installed
* Added `RequiresGit` and `RequiresDocker` decorators to project init function

## Tests
* Uninstalled `git` to check that an error would be thrown
* Reinstalled `git` to ensure that error was no longer thrown


tjhiggins:
Added Tyler since he just did something similar for docker

TylerAldrich (approved):
this looks familiar

PR #692, opened 2022-08-31
Change default project name for `architect dev`, allowing multiple envs w/ same name

TylerAldrich:
Default project name is now the name of the first component in the `architect dev` command. If you run multiple components at once (e.g. `architect dev architect.yml architect-2.yml`), the first architect.yml's component name is used. (Open to suggestions here, but prior it was always named `architect` so this is still better imo).

When using the `-e` flag, that still overrides this default, so `architect dev -e foobar` will run `docker compose ... -p foobar up`.

If you run `architect dev` on the same component multiple times, each subsequent compose is given a number (e.g. `hello-world`, `hello-world-1`, `hello-world-2`, etc). Used this instead of a UID/some random number because each instance creates new containers and we want to re-use those containers when possible. If we always create a new hash/uid/etc, there will be tons of stopped orphaned containers after a bunch of devs which isn't ideal.

Example of running Architect twice working now - 
image
(note - screenshot is slightly outdated, this would be `arc-cloud` and `arc-cloud-1` now)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ $ npm install -g @architect-io/cli
 $ architect COMMAND
 running command...
 $ architect (--version)
-@architect-io/cli/1.24.0-rc.10 linux-x64 node-v16.16.0
+@architect-io/cli/1.24.0-rc.11 linux-x64 node-v16.17.0
 $ architect --help [COMMAND]
 USAGE
   $ architect COMMAND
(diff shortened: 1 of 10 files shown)


mueschm:
@TylerAldrich I made a change to how we detect running projects. Now we ask docker-compose for a list of running projects and query the labels of all active containers. Using the labels, we can determine which ones belong to us, and then we can remove any that are not active by cross-referencing with the active query.

I think this may help make sure our source of truth is always accurate. This also upgrades the way we grab envs in general so should improve the whole system.

TylerAldrich:
> @TylerAldrich I made a change to how we detect running projects. Now we ask docker-compose for a list of running projects and query the labels of all active containers. Using the labels, we can determine which ones belong to us, and then we can remove any that are not active by cross-referencing with the active query.
> 
> I think this may help make sure our source of truth is always accurate. This also upgrades the way we grab envs in general so should improve the whole system.

Ah nice, this does help with the weird behavior if you happen to have lingering ymls in the config dir that didn't get cleaned up properly due to some other bugs so I like this

PR #698, opened 2022-09-01
Prevent `architect register` failures in docker buildx from throwing Sentry errors

TylerAldrich:
Log the error from `docker buildx bake` and exit the process (with whatever error code buildx bake returned with) so that we don't get sentry logs about this.

As we're just running a docker command here, the most likely cause of failure is somebody trying to register a Dockerfile that can't be build properly (for any reason - syntax errors, copying a file that doesn't exist, running a command that fails, etc). This isn't something we can fix, so we shouldn't be throwing an error and should just be informing the user.

### DIFF
--- a/src/commands/register.ts
+++ b/src/commands/register.ts
@@ -240,8 +240,7 @@ export default class ComponentRegister extends BaseCommand {
         });
       } catch (err: any) {
         fs.removeSync(compose_file);
-        console.log(chalk.red(err));
-        process.exit(err.exitCode || 1);
+        this.error(new ArchitectError(err.message, false));
       }
     }
 
--- a/src/commands/register.ts
+++ b/src/commands/register.ts
@@ -240,7 +240,8 @@ export default class ComponentRegister extends BaseCommand {
         });
       } catch (err: any) {
         fs.removeSync(compose_file);
-        this.error(err);
+        console.log(chalk.red(err));
+        process.exit(err.exitCode || 1);
       }
     }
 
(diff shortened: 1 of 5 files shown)


src/commands/register.ts
```diff
@@ -240,7 +240,8 @@ export default class ComponentRegister extends BaseCommand {
         });
       } catch (err: any) {
         fs.removeSync(compose_file);
-        this.error(err);
+        console.log(chalk.red(err));
+        process.exit(err.exitCode || 1);
```
TylerAldrich, line 244: I don't know if it's ever possible for us to not get an exitCode from the execa exception, but technically this err is an `any` so I wanted to be safe

tjhiggins: Should we support some kind of:
`err.track = false`

src/dependency-manager/utils/errors.ts
```diff
@@ -45,7 +45,17 @@ const addLineNumbers = (value: string, errors: ValidationError[]): void => {
...
 };
 
-export class ArchitectError extends Error { }
+export class ArchitectError extends Error {
+  // When track is true, the exception will be captured by endSentryTransaction.
+  // Should be set to false when a raised error is an issue on the users end that doesn't have
+  // anything actionable for us to do.
+  track: boolean;
+
+  constructor(msg?: string, track?: boolean) {
```
mueschm, line 54: Can't we just do
```
constructor(msg?: string, track = true)
```

mueschm (changes requested):

src/dependency-manager/utils/errors.ts
TylerAldrich: ... yep we totally can, somehow didn't yet realize that TS support default parameters 😂

PR #701, opened 2022-09-02
Scale command

ryan-cahill:
## Overview
Adds the `architect scale` command to scale services and set scaling configuration per environment

https://gitlab.com/architect-io/architect-cli/-/issues/442

## Changes
* Added scale command and tests
* Updated component and component version entities

## Tests
* Added tests
* Ran `architect scale` commands with and without flags/args

## Docs
https://github.com/architect-team/architect-documentation/pull/15

## Pictures
image

image

image


src/commands/scale.ts
```diff
@@ -0,0 +1,84 @@
...
+  static args = [{
+    sensitive: false,
+    name: 'component_name',
+    description: 'Name of service to scale, ex. account/component:latest',
+  }];
+
+  async run(): Promise<void> {
+    const { args, flags } = await this.parse(Scale);
+
+    const { component_account_name, component_name, tag, instance_name } = ComponentVersionSlugUtils.parse(args.component_name);
```
mueschm, line 38: We should probably allow the user to select from a list of components here if they did not enter one.

src/commands/scale.ts
```diff
@@ -0,0 +1,84 @@
...
+      sensitive: false,
+    }),
+    replicas: Flags.integer({
+      name: 'replicas',
+      description: 'Number of desired service replicas',
+      sensitive: false,
+    }),
+  };
+
+  static args = [{
```
mueschm, line 29: It feels odd to me this is in an arg. Why is service a flag and this is not a flag? Would like some input from @tjhiggins since I know this keeps some consistency with other commands, but it does feel weird here.

mueschm (changes requested):

mueschm:
What happens if a component exists on the account but that version is not deployed to the env? For instance a user thinks version B is deployed but its version C. What happens?

src/commands/scale.ts
tjhiggins: That is an extremely fair question. I'd be curious for thoughts/proposals. @TylerAldrich @ryan-cahill

TylerAldrich: I wonder if the opposite is more natural here - e.g. `architect scale -e prod --component tylerland:latest --replicas 5000 my_api_service`? Or maybe they should just all be flags...

`architect scale -e prod --replicas 5000 --service my_api_sevice tylerland:latest` makes more sense if we allow them to omit the service and prompt them with a list, though

ryan-cahill: I initially thought I'd use the arg to be more consistent with the deploy command, but I was considering converting them all to flags since I don't think users would want the tag used anyway. It would be both annoying and unnecessary to have to scale by tag

mueschm: We should probably remove the tag entirely since I think it just causes confusion. Like what happens if you apply to version A but then push version B. Both should be scaled to the same size you would assume.

ryan-cahill: You can't currently deploy multiple components of different tags to the same environment anyway, so that's another reason not to use the tag

src/commands/scale.ts
```diff
@@ -47,6 +47,7 @@ export default class Scale extends BaseCommand {
   }];
 
   async run(): Promise<void> {
+    this.log('This feature is in alpha. Please send any comments or questions to support@architect.io.');
```
mueschm, line 50: 2 quick things.
1. Should be yellow
2. Should mention it is subject to change.
```
This feature is in alpha. While the feature should be stable, it may be changed or removed without prior notice. As such we do not recommend using this feature in any automated pipelines.

During this time we greatly appreciate any feedback as we continue to finalize the implementation. You can reach us at support@architect.io.
```
This will make sure
1. It is noticed since a giant yellow text string is usually not missed.
2. They understand we are striving to make it stable but guarantee nothing.
3. Set limitations on how we expect the feature to be used currently.
4. That we value their insight
5. That at some point this will move out of alpha

ryan-cahill: Updated:

image

PR #703, opened 2022-09-09
Optimize by not building the same dockerfile n times.

tjhiggins:
Utilizes docker-compose build.tags.


tjhiggins:
@TylerAldrich good to review, but I still need to add a test case or two.

src/commands/dev/index.ts
```diff
@@ -430,8 +430,6 @@ export default class Dev extends BaseCommand {
     const { flags } = await this.parse(Dev);
     const [project_name, compose_file] = await this.buildImage(compose, default_project_name);
 
-    console.clear();
```
TylerAldrich, line 433: I always thought this was kind of annoying 😛

TylerAldrich (approved):
Changes all LGTM

PR #704, opened 2022-09-12
Fix https to https issue

mueschm:
## Issue
Traefik by default assumes that the downstream system is http. This is not always true since some systems handle their own https connections. 

## Fix
This does two main things.
1. Tell Traefik to ignore self signed certs in the downstream.
2. Tell the load balancer which protocol to use if its defined. This will defer to the protocol field but also attempt to parse the url if available.

## Test
Tested with our local stack and everything seems to be working.

### DIFF
--- a/src/common/docker-compose/index.ts
+++ b/src/common/docker-compose/index.ts
@@ -350,7 +350,7 @@ export class DockerComposeUtils {
           }
           const protocol = component_interface?.protocol || parseUrl(component_interface?.url).protocol;
           if (protocol) {
-            service_to.labels.push(`traefik.http.services.${traefik_service}-service.loadbalancer.server.scheme=${protocol}`)
+            service_to.labels.push(`traefik.http.services.${traefik_service}-service.loadbalancer.server.scheme=${protocol}`);
           }
         }
       }
--- a/src/common/docker-compose/index.ts
+++ b/src/common/docker-compose/index.ts
@@ -1,3 +1,4 @@
+import { parseUrl } from '@sentry/utils';
 import chalk from 'chalk';
 import execa, { Options } from 'execa';
 import fs from 'fs-extra';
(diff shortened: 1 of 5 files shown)


src/common/docker-compose/index.ts
```diff
@@ -1,3 +1,4 @@
+import { parseUrl } from '@sentry/utils';
```
tjhiggins, line 1: This doesn't seem right

src/common/docker-compose/index.ts
```diff
@@ -116,6 +117,8 @@ export class DockerComposeUtils {
           '--providers.docker.exposedByDefault=false',
           `--providers.docker.constraints=Label(\`traefik.port\`,\`${gateway_port}\`)`,
           ...(use_ssl ? [
+            // Ignore local certs being invalid on proxy
+            `--serversTransport.insecureSkipVerify=true`,
```
tjhiggins, line 121: I feel like we should allow for https protocols and use a tcp router (ssl passthrough), but maybe this is the stop gap solution?

mueschm: So I was thinking about this, but by not using the TCP router we actually still get a valid certificate locally for the first hop.

So client <-> traefik <-> service.
While the `traefik <-> service` may not be valid ssl, the `client <-> traefik` is going to be valid ssl since we terminate the ssl with our valid cert and then just passthrough.

tjhiggins: Will the connection still be secure? Sounds like it, but I wanted to confirm to make sure secure cookies for example still work.

tjhiggins (changes requested):

src/common/docker-compose/index.ts
```diff
@@ -345,13 +347,31 @@ export class DockerComposeUtils {
             service_to.labels.push(`traefik.http.routers.${traefik_service}.entrypoints=web`);
             service_to.labels.push(`traefik.http.routers.${traefik_service}.tls=true`);
           }
+
+          const url_protocol = this.getUrlProtocol(component_interface?.url);
```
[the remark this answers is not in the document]
tjhiggins: I think we default to http if it isn't set. `component_interface?.protocol || 'http'`. We should probably be setting it explicitly.

src/common/docker-compose/index.ts
mueschm: Everything appears to work fine. Secure cookies work exactly as expected and the client and server both don't complain.

src/common/docker-compose/index.ts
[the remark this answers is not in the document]
mueschm: I would rather fix the parsing code to always display the protocol then so we do not have to make any sort of assumptions here. Which would you prefer?

PR #706, opened 2022-09-14
483 stop command

mtgoncalves1:
Closes issue https://gitlab.com/architect-io/architect-cli/-/issues/483

## Tests   
Deploy an application locally:   
   % architect dev .   
Stop the local deployment:   
   % architect stop <local-environment-name>   
where `local-environment-name` can be found by running   
   % architect dev:list 

## Docs
https://github.com/architect-team/architect-documentation/pull/13

### DIFF
--- a/src/commands/dev/stop.ts
+++ b/src/commands/dev/stop.ts
@@ -34,7 +34,7 @@ export default class DevStop extends BaseCommand {
 
   async waitForEnviromentToStop(environment: string): Promise<boolean> {
     let attempts = 0;
-    // Essentially add a 2 mintue timeout
+    // Essentially add a 3 mintue timeout
     while (attempts < 180) {
       const environments = await DockerComposeUtils.getLocalEnvironments();
       if (environments.includes(environment)) {
--- a/src/commands/dev/stop.ts
+++ b/src/commands/dev/stop.ts
@@ -1,17 +1,19 @@
+import { CliUx } from '@oclif/core';
 import chalk from 'chalk';
-import BaseCommand from "../base-command";
-import { DockerComposeUtils } from "../common/docker-compose";
 import inquirer from 'inquirer';
-import { RequiresDocker } from '../common/docker/helper';
+import { ArchitectError } from '../..';
+import BaseCommand from "../../base-command";
+import { DockerComposeUtils } from "../../common/docker-compose";
+import { RequiresDocker } from '../../common/docker/helper';
 
-export default class Stop extends BaseCommand {
+export default class DevStop extends BaseCommand {
   async auth_required(): Promise<boolean> {
     return false;
   }
 
   static description = 'Stop a local deployment';
   static examples = [
-    'architect stop <local-environment-name>',
+    'architect dev:stop <local-environment-name>',
   ];
 
   static flags = {
@@ -25,9 +27,26 @@ export default class Stop extends BaseCommand {
     required: false,
   }];
 
+  async waitForEnviromentToStop(environment: string): Promise<boolean> {
+    let attempts = 0;
+    // Essentially add a 2 mintue timeout
+    while (attempts < 180) {
+      const environments = await DockerComposeUtils.getLocalEnvironments();
+      if (environments.includes(environment)) {
+        await new Promise((resolve) => {
+          setTimeout(resolve, 1000);
+        });
+      } else {
+        return true;
+      }
+      attempts++;
+    }
+    return false;
+  }
+
   @RequiresDocker({ compose: true })
   async run(): Promise<void> {
-    const { args } = await this.parse(Stop);
+    const { args } = await this.parse(DevStop);
 
     const env_names = await DockerComposeUtils.getLocalEnvironments();
     if (env_names.length === 0) {
@@ -51,6 +70,13 @@ export default class Stop extends BaseCommand {
 
     const compose_file = DockerComposeUtils.buildComposeFilepath(this.app.config.getConfigDir(), name);
     await DockerComposeUtils.dockerCompose(['-p', name, '-f', compose_file, 'stop']);
-    this.log(chalk.green(`Successfully stopped local deployment '${name}'.`));
+    CliUx.ux.action.start(chalk.blue(`Waiting for ${name} to stop...`));
+    const did_stop = await this.waitForEnviromentToStop(name);
+    CliUx.ux.action.stop();
+    if (did_stop) {
+      this.log(chalk.green(`Successfully stopped local deployment '${name}'.`));
+    } else {
+      this.error(new ArchitectError(`Unable to stop ${name}`));
+    }
   }
 }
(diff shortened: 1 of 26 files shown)


src/commands/stop.ts
```diff
@@ -0,0 +1,56 @@
+import chalk from 'chalk';
+import BaseCommand from "../base-command";
+import { DockerComposeUtils } from "../common/docker-compose";
+import inquirer from 'inquirer';
+import { DockerInspect } from '../common/docker-compose/template';
+
+export default class StopLocalDeployment extends BaseCommand {
```
TylerAldrich, line 7: Based on how other commands are named we probably should just name this class `Stop` (or `StopCommand` if there happens to be some other `Stop` class)

src/commands/stop.ts
```diff
@@ -0,0 +1,56 @@
...
+    name: 'name',
+    description: 'Name of local environment',
+    required: false,
+  }];
+
+  async run(): Promise<void> {
+    const { args } = await this.parse(StopLocalDeployment);
+
+    const local_env_map = await DockerComposeUtils.getLocalEnvironmentContainerMap();
+    const env_names = Object.keys(local_env_map);
```
TylerAldrich, line 32: You can use `DockerComposeUtils.getLocalEnvironments()` if you only need the keys (the `getLocalEnvironments` function is just `return Object.keys(await this.getLocalEnvironmentContainerMap());`)

src/commands/stop.ts
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  static args = [{
+    sensitive: false,
+    name: 'name',
+    description: 'Name of local environment',
+    required: false,
+  }];
+
+  async run(): Promise<void> {
```
TylerAldrich, line 28: This should have the `@RequiresDocker({ compose: true })` decorator on it - I just now realized I (already) forgot about this and didn't include it in the dev:list command (see https://github.com/architect-team/architect-cli/pull/708 - I just now fixed it 😄)

src/common/docker-compose/index.ts
```diff
@@ -672,9 +672,14 @@ export class DockerComposeUtils {
           const state = container_state.State.toLowerCase();
           const health = container_state.Health.toLowerCase();
 
-          const bad_state = state !== 'running';
+          const bad_state = state !== 'running' && container_state.ExitCode !== 0;
           const bad_health = health === 'unhealthy';
 
+          // Stop watching when container exited successfully.
+          if (state === 'exited' && container_state.ExitCode === 0) {
+            return true;
```
TylerAldrich, line 680: We should return `restarted` here instead of `true` - kind of complicated but the result of this is looked at when `dev` is doing it's cleanup and if a container restarted _while_ cleanup is happening, we need to do an additional step, so if we just return `restarted` here that'll do the right thing in weird edge cases

src/common/docker-compose/index.ts
```diff
@@ -672,9 +672,14 @@ export class DockerComposeUtils {
           const state = container_state.State.toLowerCase();
           const health = container_state.Health.toLowerCase();
 
-          const bad_state = state !== 'running';
+          const bad_state = state !== 'running' && container_state.ExitCode !== 0;
```
TylerAldrich, line 675: Ah interesting I forgot we treat all non-running states as "bad" - I think this works but I'll test a couple of the weird dev shutdown cases I was using before to make sure

TylerAldrich (changes requested):
A couple small suggestions - I'm going to pull this branch and repeat some of the "making sure dev stops without leaving containers running" tests I was doing a couple weeks back to make sure this is compatible, I'll let you know if there's anything else I find

src/common/docker-compose/index.ts
mtgoncalves1: This makes sense. Returning `true` will call the `stop` command again.

mtgoncalves1: I'll change it to return `restarted`.

src/common/docker-compose/index.ts
mtgoncalves1: Sounds good. Thank you!!

src/commands/stop.ts
```diff
@@ -0,0 +1,57 @@
...
+    const { args } = await this.parse(Stop);
+
+    const env_names = await DockerComposeUtils.getLocalEnvironments();
+    if (env_names.length === 0) {
+      throw new Error(chalk.red(`No local deployment found.`));
+    }
+
+    if (!args.name) {
+      const answers: { environment: string } = await inquirer.prompt([
```
mueschm, line 38: Neat trick that is used in other places.
```
default: flags.email,
when: !flags.email,
```
This will ask the user only if the flag is not set. If it is set then it won't ask and just set the answer to the flags value. This removes the conditional.

src/commands/stop.ts
```diff
@@ -0,0 +1,57 @@
...
+      ]);
+      args.name = answers.environment;
+    }
+
+    if (!env_names.includes(args.name as string)) {
+      throw new Error(chalk.red(`No local deployment named '${args.name}'. Use command 'architect dev:list' to list local deployments.`));
+    }
+
+    const compose_file = DockerComposeUtils.buildComposeFilepath(this.app.config.getConfigDir(), args.name);
+    await DockerComposeUtils.dockerCompose(['-p', args.name, '-f', compose_file, 'stop']);
```
mueschm, line 54: I don't think you need the `-f` here. I have tested using just `-p` and it has worked fine. Please verify.

mueschm (changes requested):

src/commands/stop.ts
mtgoncalves1: Tyler found this Windows-related issue and it can be fixed with the `-f` flag.

src/commands/stop.ts
mtgoncalves1: That is a good trick.

TylerAldrich (approved):

mueschm:
Actually @mtgoncalves1 can you verify this still works with the `--detached` flag?

mtgoncalves1:
> Actually @mtgoncalves1 can you verify this still works with the `--detached` flag?

@mueschm I tried `docker compose -p arc-react-jmoffc4j -f {path-to-compose-file}/arc-react-jmoffc4j.yml --detached stop` but it failed with an `unknown flag` error.

mueschm:
Bit of confusion. I am talking about the --detached flag on the dev command itself. Want to make sure that we detach and then stop correctly. We should verify everything works by running dev like
```
architect dev --detached  .
```
as well as
```
architect dev . &
```

src/commands/dev/stop.ts
```diff
@@ -0,0 +1,82 @@
...
+  static args = [{
+    sensitive: false,
+    name: 'name',
+    description: 'Name of local environment',
+    required: false,
+  }];
+
+  async waitForEnviromentToStop(environment: string): Promise<boolean> {
+    let attempts = 0;
+    // Essentially add a 2 mintue timeout
```
TylerAldrich, line 32: this is 3 minutes with 1 check per second right?

mueschm: Listen math is hard, but yes it is

PR #711, opened 2022-09-16
Add output type to dev list

mueschm:
## Overview
Dev list allows a user to understand what is running in the background, but if you want to automate behavior based on it, you need to create a special parser to read the data. This allows us to output the same format in json which is much easier to read.

## Changes
Add `--format` to dev list. Default behavior does not change, but you can also choose json.

## Tests
Added 1 new unit test
Tested it out
image

### DIFF
--- a/src/commands/dev/list.ts
+++ b/src/commands/dev/list.ts
@@ -62,7 +62,7 @@ export default class DevList extends BaseCommand {
         };
       }
     }
-    this.log(util.inspect(output, false, 5));
+    this.log(JSON.stringify(output, null, 2));
   }
 
   @RequiresDocker({ compose: true })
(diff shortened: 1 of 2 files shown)


src/commands/dev/list.ts
```diff
@@ -1,33 +1,79 @@
...
+      output[env_name] = {} as Dictionary<any>;
+      const container_names = this.getContainerNames(containers);
+      const statuses = this.getContainerStates(containers);
+      for (let i = 0; i < container_names.length; i++) {
+        output[env_name][container_names[i]] = {
+          status: statuses[i],
+        };
+      }
+    }
+    this.log(util.inspect(output, false, 5));
```
TylerAldrich, line 65: Love this change - one note is it looks like this doesn't output valid JSON though - I think `JSON.stringify()` would work? The thing I noticed was the status key not being quoted

TylerAldrich (changes requested):

TylerAldrich (approved):

PR #713, opened 2022-09-19
Dont always try to setRawMode in runLocal, only do so if tty

TylerAldrich:
Saw Michael ran into this issue on Friday and noticed that the tty testing I was doing a month or so back was almost exclusively in the `runRemote` method and I didn't verify that local worked. We always passed `true` which attempted to call `process.stdin.setRawMode(true)`, which only works if `process.stdin` is a tty ReadStream.

sentry error: https://sentry.io/organizations/architect-io/issues/3598979632/?project=6465948&referrer=slack

### DIFF
--- a/src/commands/exec.ts
+++ b/src/commands/exec.ts
@@ -373,8 +373,9 @@ Alternatively, running "architect --% exec -- ls" will prevent the PowerShell pa
     const service = await DockerComposeUtils.getLocalServiceForEnvironment(compose_file, args.resource);
 
     const compose_args = ['-f', compose_file, '-p', environment_name, 'exec'];
+    const use_tty = flags.tty && process.stdout.isTTY;
     // https://docs.docker.com/compose/reference/exec/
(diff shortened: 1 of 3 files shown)


src/commands/exec.ts
```diff
@@ -383,7 +384,7 @@ Alternatively, running "architect --% exec -- ls" will prevent the PowerShell pa
       compose_args.push(arg);
     }
 
-    await DockerComposeUtils.dockerCompose(compose_args, { stdio: 'inherit' }, true);
+    await DockerComposeUtils.dockerCompose(compose_args, { stdio: 'inherit' }, use_tty);
```
mueschm, line 387: The default value for `flags.tty` is undefined. Doesn't this change the default behavior?

TylerAldrich: If `flags.tty` isn't explicitly set by the user it defaults to `flags.tty = !!process.stdin.isTTY;` in the main `run` method before we get here, so at this point it'll always be a bool

PR #719, opened 2022-09-23
Fix issue where if the terminal is closed, the socket isnt cleaned up but still exists

TylerAldrich:
Note: This is only a problem on Unix environments. Windows sockets get cleaned up automatically when the process that spawned them is closed, so closing the powershell / cmd window with the `architect dev` process closes the socket. 

To reproduce the issue on Unix systems:

- Run `architect dev`, once it starts, close the terminal window
- In a new window, `architect dev:stop` will throw an error because the socket exists but nobody is listening.
- If you stop the docker containers yourself, and run `architect dev` again, it blows up because the socket already exists.
- Only workaround without this change is to manually `rm` the socket

### DIFF
--- a/src/commands/dev/index.ts
+++ b/src/commands/dev/index.ts
@@ -101,11 +101,11 @@ class UpProcessManager {
     let recreated_socket = false;
     this.server.on('error', (e: any) => {
       if (e.code === 'EADDRINUSE' && this.server && !recreated_socket) {
+        recreated_socket = true;
         // Socket already exists (likely from a previous run that didnt get cleaned up properly)
         // Remove it and listen again creating a new socket.
         fs.rmSync(this.socket);
         this.server.listen(this.socket);
-        recreated_socket = true;
       } else {
         throw e;
       }
--- a/src/commands/dev/index.ts
+++ b/src/commands/dev/index.ts
@@ -98,12 +98,16 @@ class UpProcessManager {
       });
     });
 
+    let recreated_socket = false;
     this.server.on('error', (e: any) => {
-      if (e.code === 'EADDRINUSE' && this.server) {
+      if (e.code === 'EADDRINUSE' && this.server && !recreated_socket) {
         // Socket already exists (likely from a previous run that didnt get cleaned up properly)
         // Remove it and listen again creating a new socket.
         fs.rmSync(this.socket);
         this.server.listen(this.socket);
+        recreated_socket = true;
+      } else {
+        throw e;
       }
     });
 
--- a/src/commands/dev/index.ts
+++ b/src/commands/dev/index.ts
@@ -89,7 +89,7 @@ class UpProcessManager {
     const compose_process = DockerComposeUtils.dockerCompose(compose_args,
       { stdout: 'pipe', stdin: 'ignore', detached: !this.is_windows });
 
-    this.server = net.createServer().listen(this.socket);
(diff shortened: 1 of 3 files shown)


src/commands/dev/stop.ts
```diff
@@ -79,11 +84,19 @@ export default class DevStop extends BaseCommand {
       socket.write('stop', () => {
         socket.end();
       });
+
+      socket.on('error', (e: any) => {
+        if (e.code === 'ECONNREFUSED') {
+          socket.end();
+          this.runComposeStop(name);
+        } else {
```
TylerAldrich, line 92: This is the fallback where if you have a socket file but the dev process is closed and the socket is no longer being used, we run compose stop.

src/commands/dev/index.ts
```diff
@@ -98,6 +98,15 @@ class UpProcessManager {
       });
     });
 
+    this.server.on('error', (e: any) => {
+      if (e.code === 'EADDRINUSE' && this.server) {
```
mueschm, line 102: What happens if this errors twice? This seems like it could be an infinite loop.

TylerAldrich: Hmm yep, not sure if there's a circumstance that could make that happen, but theoretically we could get into an infinite delete/recreate/error/repeat loop. I'll update this to hard-fail if we already tried to fix this error and then got it again

src/commands/dev/index.ts
```diff
@@ -98,6 +98,19 @@ class UpProcessManager {
...
     });
 
+    let recreated_socket = false;
+    this.server.on('error', (e: any) => {
+      if (e.code === 'EADDRINUSE' && this.server && !recreated_socket) {
+        // Socket already exists (likely from a previous run that didnt get cleaned up properly)
+        // Remove it and listen again creating a new socket.
+        fs.rmSync(this.socket);
+        this.server.listen(this.socket);
+        recreated_socket = true;
```
mueschm, line 108: This should be the first thing in the function. No idea if `listen` calls an sync function or checks for the pipe beforehand. Also not sure if it changes per version of node which knowing Node would not be surprising.

TylerAldrich: I did now quickly check back to node 12.x and it seems to be a very similar description to the latest version docs https://nodejs.org/docs/latest-v12.x/api/net.html#net_server_listen so I do think it this particular bit of functionality is consistent across node versions 😬 (at least, Listen was still async in previous versions)

I think I understand what you're saying about first thing in the function but just to clarify, what we probably should do is:
```typescript
this.server = net.createSever();
this.server.on('connection', ...);
this.server.on('error', ...);
this.server.listen();
```

so that we don't hit this error before we actually start listening, right? (Listen is async which is why this worked on my machine, but I will definitely change this - let me know if you meant something else 😛 )

mueschm: What I more mean is that `listen` could use `fs.readFileSync` to check that the file exists. If for some reason the cleanup code does not work due to some change then the `recreated_socket` will never hit true so your loop protection would not work. Instead, we should mark we have just been in the function once so that no matter what happens we just exit if we ever hit this function again.

TylerAldrich: Oooooh, I understand what you're saying now, brain is working at half speed today but that makes complete sense lol, I'll move this to the top of the block

PR #720, opened 2022-09-23
544 no dev instances console message

devinsag:
## Overview
Architect dev:list displays empty column headers instead of a message similar to components or environments when none are found.

## Purpose
Update dev:list to display a similar message as components or environments when no running services are found. 

## Changes
Add table length and json length check before printing to console. Print friendly alternative message if no services are running.

### Running dev instances: 
```sh
architect dev:list
architect dev:list --format json
```
### No Running dev instances:
```sh
architect dev:list
architect dev:list --format json
architect dev:list --format JSON
architect dev:list --format table
architect dev:list --format TABLE
```

## Docs
N/A

## Pictures
please see: https://gitlab.com/architect-io/architect-cli/-/issues/544

### DIFF
--- a/src/commands/dev/list.ts
+++ b/src/commands/dev/list.ts
@@ -32,7 +32,7 @@ export default class DevList extends BaseCommand {
       table.push([env_name, container_names, statuses]);
     }
     if (!table.length) {
-      this.log('There have no active dev instances yet. Use `architect dev` to create one.');
+      this.log('There are no active dev instances yet. Use `architect dev` to create one.');
     } else {
       this.log(table.toString());
     }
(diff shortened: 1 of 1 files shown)


src/commands/dev/list.ts
```diff
@@ -57,7 +59,11 @@ export default class DevList extends BaseCommand {
         };
       }
     }
-    this.log(JSON.stringify(output, null, 2));
+    if (!Object.keys(output).length) {
+      this.log('There are no active dev instances yet. Use `architect dev` to create one.');
+    } else {
+      this.log(JSON.stringify(output, null, 2));
+    }
```
TylerAldrich, line 66: I know this is already merged but we may want to change this back - if I specifically want the output format to be json, we shouldn't return non-json even if there's no results (returning `{}` for example would at least not break somebodys `JSON.parse()` equivalent or pipe-to-jq)

PR #721, opened 2022-09-23
always print json format when json format flag is specified

devinsag:
update output for json dev empty to be empty set

---
### Edit

Tested just to be sure the changes will only affect output when using  json format flag specifically.
```
architect-local dev:list --format JSON
{}
architect-local dev:list --format json
{}
architect-local dev:list --format table
There are no active dev instances yet. Use `architect dev` to create one.
architect-local dev:list --format TABLE
There are no active dev instances yet. Use `architect dev` to create one.
architect-local dev:list
There are no active dev instances yet. Use `architect dev` to create one.
```

### DIFF
--- a/src/commands/dev/list.ts
+++ b/src/commands/dev/list.ts
@@ -59,11 +59,7 @@ export default class DevList extends BaseCommand {
         };
       }
     }
-    if (!Object.keys(output).length) {
-      this.log('There are no active dev instances yet. Use `architect dev` to create one.');
-    } else {
-      this.log(JSON.stringify(output, null, 2));
-    }
+    this.log(JSON.stringify(output, null, 2));
   }
 
   @RequiresDocker({ compose: true })


src/commands/dev/list.ts
```diff
@@ -59,11 +59,7 @@ export default class DevList extends BaseCommand {
         };
       }
     }
-    if (!Object.keys(output).length) {
```
tjhiggins, line 62: Talked offline, but keep the friendly error msg unless they use format json

tjhiggins (changes requested):

src/commands/dev/list.ts
TylerAldrich: that's what this is doing I believe - this is within the `outputJSON` function and the other table output still has the error msg

mueschm: This does do that. The output table function still produces the message. Only the output json function had this removed. Unless you mean something else.

devinsag: Tested changes for sanity check. 

```
architect-local dev:list --format JSON
{}
architect-local dev:list --format json
{}
architect-local dev:list --format table
There are no active dev instances yet. Use `architect dev` to create one.
architect-local dev:list --format TABLE
There are no active dev instances yet. Use `architect dev` to create one.
architect-local dev:list
There are no active dev instances yet. Use `architect dev` to create one.
```

devinsag: This is my fault for the horrific naming of this PR. I've updated the title to be more specific and description to include testing results.

tjhiggins: ah my bad - assumed there was one output flow

PR #725, opened 2022-10-03
527 platform secrets

mtgoncalves1:
## Overview
Upload and download platform secrets via CLI.  
Closes https://gitlab.com/architect-io/architect-cli/-/issues/527.  

## Changes
Add platform flag to upload and download secrets

#### Upload platform secrets:
- Create a secret file (Ex: my-secrets.yml)
- Create a platform
- Run the following command
architect-local secrets:set --account=account-name --platform=platform-name my-secrets.yml
- To override existing secret values add the override flag:
architect-local secrets:set --account=account-name --platform=platform-name --override my-secrets.yml

#### Download platform secrets:  
architect-local secrets --account=account-name --platform=platform-name  my-secrets.yml

## Docs
PR - https://github.com/architect-team/architect-documentation/pull/18


src/architect/secret/secret.utils.ts
```diff
@@ -10,25 +11,36 @@ export interface Secret {
...
+    if (environment_name) {
       const environment = await EnvironmentUtils.getEnvironment(app.api, account, environment_name);
       if (inherited) {
         secrets = (await app.api.get(`environments/${environment.id}/secrets/values`, { params: { inherited: true } })).data;
       } else {
         secrets = (await app.api.get(`environments/${environment.id}/secrets/values`)).data;
       }
+    } else if (platform_name) {
+      const platform = await PlatformUtils.getPlatform(app.api, account, platform_name);
+      if (inherited) {
```
mueschm, line 25: Why not just do
```
secrets = (await app.api.get(`platforms/${platform.id}/secrets/values`, { params: { inherited } })).data;
```

src/commands/secrets/download.ts
```diff
@@ -38,7 +41,17 @@ export default class SecretsDownload extends BaseCommand {
       this.error('You do not have permission to download secrets. Please contact your admin.');
     }
 
-    const secrets = await SecretUtils.getSecrets(this.app, account, flags.environment, true);
+    if (flags.platform && flags.environment) {
+      throw new Error('Please provide either the platform flag or the environment flag and not both.');
+    }
+
+    let secrets;
```
mueschm, line 48: This needs to be const. If we are going to tell the user that something does not exist, then we should check it and make sure instead of waiting on an error to be thrown.

mueschm (changes requested):
So something at a high level feels wrong about this command now(Not your fault). It feels like it is ripe for things to be error prone. I almost want to add a `type` flag to make sure users are explicit about what they are uploading or downloading so we can better help them. 

Can you take some examples post them in engineering and lets collect some feedback about this. Otherwise please just fix the other comments.

src/commands/secrets/upload.ts
```diff
@@ -38,13 +45,22 @@ export default class SecretsUpload extends BaseCommand {
 
   async run(): Promise<void> {
     const { flags, args } = await this.parse(SecretsUpload);
-    const account = await AccountUtils.getAccount(this.app, flags.account);
+    if (!flags.account) {
+      this.log(chalk.red('Account is not found. Please see examples below:'));
+      this.log(`  ${SecretsUpload.examples.join('\n  ')}`);
+      return;
```
mueschm, line 51: We want to exit with a non 0 status code here since the operation itself failed. `this.exit(1)` should work or just throwing an `ArchitectError` that is not tracked.

src/commands/secrets/download.ts
```diff
@@ -13,13 +15,15 @@ export default class SecretsDownload extends BaseCommand {
...
-    'architect secrets --account=myaccount --environment=myenvironment ../mysecrets.yml',
+    'architect secrets --account=myaccount ./mysecrets.yml',
+    'architect secrets --account=myaccount --platform=myplatform ./mysecrets.yml',
+    'architect secrets --account=myaccount --environment=myenvironment ./mysecrets.yml',
   ];
   static flags = {
     ...BaseCommand.flags,
     ...AccountUtils.flags,
     ...EnvironmentUtils.flags,
+    ...PlatformUtils.flags,
```
mueschm, line 26: These need to be updated. We no longer want to pull them in like this since if you set the env of the process then those flags can be set implicity. We want to copy them over here instead and remove the `environment` field.

src/commands/secrets/upload.ts
mueschm, line 51: Same as above.

src/commands/secrets/upload.ts
```diff
@@ -38,13 +45,22 @@ export default class SecretsUpload extends BaseCommand {
...
+    }
 
+    const account = await AccountUtils.getAccount(this.app, flags.account);
     const is_admin = await UserUtils.isAdmin(this.app, account.id);
     if (!is_admin) {
       this.error('You do not have permission to upload secrets. Please contact your admin.');
     }
 
+    if (flags.platform && flags.environment) {
+      throw new Error('Please provide either the platform flag or the environment flag and not both.');
```
mueschm, line 61: This should be `ArchitectError` with `track` set to false.

mueschm (changes requested):

src/commands/secrets/upload.ts
tjhiggins: track if false by default now

PR #727, opened 2022-10-05
Add dependencies warnings in register command

mtgoncalves1:
## Overview
Issue https://gitlab.com/architect-io/architect-cli/-/issues/532.  
To make sure the user know what dependencies are needed for a specific component during register, we want to inform them using `warnings` for any missing dependencies that are yet registered.

## Changes
Added a function to warn the user for any dependencies or tags specified in the component spec but not found in the account components.

## Tests
Test for missing dependencies:   
With a new account, register the React app.  

Test for non-existing dependencies tag:  
With a new account, register the NodeJS app.  
In the React architect.yml, change `dependencies` to the following line:
```
dependencies:
  node-rest-api: none
```
Goto React app and register.  

## Pictures
Screen Shot 2022-10-19 at 10 24 57 AM


mueschm:
@ryan-cahill can you review this? I feel like my mental model of component versions is a bit weak right now and want to verify this is solid.

src/commands/register.ts
```diff
@@ -360,4 +364,37 @@ export default class ComponentRegister extends BaseCommand {
...
+    for (const key of Object.keys(component_dependencies)) {
+      if (!existing_account_components[key]) {
+        register_warnings.push(key); 
+      } 
+      if (component_dependencies[key] !== 'latest') {
+        latest_tag_warnings.push(key);
+      }
+    }
+
+    const warning = chalk.hex('#FFA500');
```
ryan-cahill, line 386: Why use this instead of `chalk.yellow()`?

src/commands/register.ts
```diff
@@ -360,4 +364,37 @@ export default class ComponentRegister extends BaseCommand {
   public static getTagFromFlags(flags: any): string {
     return flags.environment ? `${ENV_TAG_PREFIX}${flags.environment}` : flags.tag;
   }
+
+  private async generateDependenciesWarnings(component_dependencies: Dictionary<string>, account_id: string) {
+    const existing_account_components: Dictionary<string> = {};
+    const { rows, _ } = (await this.app.api.get(`/accounts/${account_id}/components`)).data;
```
ryan-cahill, line 370: This endpoint will return the most recently registered component version for up to 100 different components, so I don't think that this covers all of the bases. The version this function is looking for could be two versions ago, or component number 101 registered to the account. I don't think that we really want to overload the logic to find the latest component versions, so we might want to add another endpoint to check existing component versions given a set of component version names. We could call a different endpoint once per dependency to check if it exists, but that's not a great option. @mueschm what are your thoughts here?

src/commands/register.ts
```diff
@@ -360,4 +364,37 @@ export default class ComponentRegister extends BaseCommand {
...
+
+    const warning = chalk.hex('#FFA500');
+    if (register_warnings.length > 0) {
+      this.log(warning(`\nMissing dependencies that need to be registered:`));
+      for (const component_name of register_warnings) {
+        this.log(warning(`  - ${component_name}`));
+      }
+    }
+    if (latest_tag_warnings.length > 0) {
+      this.log(warning(`\nThe following dependencies should have the 'latest' tag:`));
```
ryan-cahill, line 394: If I register a component with the dependencies block

```yml
dependencies:
  architect-ryan/nonsense: test
  architect-ryan/some: 0.0.0
  architect-ryan/thing: latest
```

I get the warning

```
Missing dependencies that need to be registered:
  - architect-ryan/nonsense
  - architect-ryan/some
  - architect-ryan/thing

The following dependencies should have the 'latest' tag:
  - architect-ryan/nonsense
  - architect-ryan/some
```

but I specifically wanted to use the component versions `architect-ryan/nonsense:test` and `architect-ryan/some:0.0.0`, so I wouldn't expect an error here. Why does the `latest` tag need to be registered?

ryan-cahill (changes requested):

src/commands/register.ts
mtgoncalves1: It is a warning, so I thought I should use yellow. Should I use a different color instead?

src/commands/register.ts
mtgoncalves1: Ok. I see the issue now. I'll make the changes.

src/commands/register.ts
ryan-cahill: I agree that it should be yellow, but the shade of yellow should be consistent throughout the rest of the CLI. For example, most if not all places, this is the yellow that's used

https://github.com/architect-team/architect-cli/blob/rc/src/commands/scale.ts#L144

mtgoncalves1: Ahh ok. I'll use that instead.

src/commands/register.ts
mueschm: Your update does not check that the correct tag exists though and just that the component exists. So if a user wants V1 that may not exist but latest would.

I do like the idea of a dependency endpoint. We can pass up the list of component names and versions and it just returns if they exist on the account or not. That would solve the problem, be efficient and provide a possible abstraction that other systems could use if needed.

mtgoncalves1: Sounds good. I'll make the changes.

src/commands/register.ts
```diff
@@ -420,4 +424,38 @@ export default class ComponentRegister extends BaseCommand {
...
+      if (valid.component && valid.component_and_version) {
+        continue;
+      } else if (valid.component && !valid.component_and_version) {
+        tag_warnings.push(component_name);
+      } else {
+        component_warnings.push(component_name);
+      }
+    }
+
+    if (component_warnings.length > 0) {
```
mueschm, line 447: The warnings could provide more information. 
`Some required dependencies for this component have not yet been registered to your account. Please make sure to register the following before you try to deploy.`

src/commands/register.ts
```diff
@@ -420,4 +424,38 @@ export default class ComponentRegister extends BaseCommand {
...
+      this.log(chalk.yellow(`\nThe following dependencies need to be registered:`));
```
mueschm, line 448: Using `\n` should be avoided if possible. Would be better to just call `this.log()` with an empty string to make sure its platform safe.

src/commands/register.ts
```diff
@@ -420,4 +424,38 @@ export default class ComponentRegister extends BaseCommand {
...
+      for (const component_name of component_warnings) {
+        this.log(chalk.yellow(`  - ${component_name}`));
+      }
+    }
+
+    if (tag_warnings.length > 0) {
+      this.log(chalk.yellow(`\nThe following dependencies tag are not found:`));
```
mueschm, line 455: Being a bit more verbose here to make it clear why we split up the list can be helpful.
`The following components exist, but do not have the specified tag`

mueschm (changes requested):
Really small UX tweaks but otherwise code looks good.

src/commands/register.ts
```diff
@@ -420,4 +424,40 @@ export default class ComponentRegister extends BaseCommand {
...
+    for (const [component_name, tag] of Object.entries(component_dependencies)) {
+      dependency_arr.push(`${component_name}:${tag}`);
+    }
+    const dependencies: Dictionary<{ component: boolean, component_and_version: boolean }> = (await this.app.api.get(`accounts/${account_name}/components-tags`, { params: { components: dependency_arr } })).data;
+
+    const component_warnings: string[] = [];
+    const tag_warnings: string[] = [];
+    for (const [component_name, valid] of Object.entries(dependencies)) {
+      if (valid.component && valid.component_and_version) {
+        continue;
```
ryan-cahill, line 439: Why does the `continue` need to exist? Can't the if statements just be

```ts
if (!valid.component && !valid.component_and_version) {
  component_warnings.push(component_name);
} else if (valid.component && !valid.component_and_version) {
  tag_warnings.push(component_name);
}
```

ryan-cahill (changes requested):
Looks great, just one small comment

src/commands/register.ts
mtgoncalves1: Yes, it can. I was overthinking it.

ryan-cahill (approved):

PR #729, opened 2022-10-06
548 http/https validation error

mtgoncalves1:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/548.  
Since Traefik only supports http/https protocols, we want to throw an error if other protocols are provided.

## Changes
Add a validation error to validateGraph if the protocol is not http, https, or undefined.

## Tests
- Add a protocol (Ex. tcp) to a component interface
```
interfaces:
      main: 3000
      protocol: tcp
```
- architect dev .

### DIFF
--- a/src/dependency-manager/manager.ts
+++ b/src/dependency-manager/manager.ts
@@ -21,7 +21,6 @@ import { validateOrRejectSpec } from './spec/utils/spec-validator';
 import { Dictionary, transformDictionary } from './utils/dictionary';
 import { ArchitectError, ValidationError, ValidationErrors } from './utils/errors';
 import { interpolateObjectLoose, interpolateObjectOrReject, replaceInterpolationBrackets } from './utils/interpolation';
-
 export default abstract class DependencyManager {
 
   account?: string;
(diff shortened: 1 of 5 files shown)


src/dependency-manager/manager.ts
```diff
@@ -371,6 +370,20 @@ export default abstract class DependencyManager {
...
 
   validateGraph(graph: DependencyGraph): void {
+    // Check for protocols that are not http or https
+    for (const node of graph.nodes) {
+      if (node.is_external) continue;
+      if (!(node instanceof ComponentNode)) continue;
+
+      const component_interfaces = node.config.interfaces;
+      for (const [component_name, component_interface] of Object.entries(component_interfaces)) {
+        const valid_protocol = [undefined, 'http', 'https'];
```
mueschm, line 380: Move this out since it is a static definition. Could be a private readonly of the class.
Also make plural.

mueschm (changes requested):

src/dependency-manager/manager.ts
mtgoncalves1: Sounds good!

PR #731, opened 2022-10-11
Allow account selection if component account is not found

mtgoncalves1:
## Overview
Closes: https://gitlab.com/architect-io/architect-cli/-/issues/550  
When an invalid account name is specified in architect.yml, we inform the user with an error message and allow them to choose one of their existing accounts. 

## Changes
Search for the component account in user's list of accounts. If it's not found, print out the error message.

## Tests
- Change the component name in an architect.yml to `name: invalid-account-name/hello-world`
- Run `architect-local register .`

### DIFF
--- a/src/architect/account/account.utils.ts
+++ b/src/architect/account/account.utils.ts
@@ -3,6 +3,7 @@ import chalk from 'chalk';
 import inquirer from 'inquirer';
 import AppService from '../../app-config/service';
 import { DockerHelper } from '../../common/docker/helper';
+import Membership from '../user/user.utils';
 import Account from './account.entity';
 
 export default class AccountUtils {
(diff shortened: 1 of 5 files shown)


src/architect/account/account.utils.ts
```diff
@@ -71,6 +72,17 @@ export default class AccountUtils {
       return this.getLocalAccount();
     }
 
+    if (options?.is_component_account_name) {
```
mueschm, line 75: I dont think this belongs here. The general idea is valid. However this adds a lot of complexity to an already complex function.

So instead what we should do is create a function that validates if the user is part of the account and then in the command call that to understand if its valid. So in the called class you would have.

```
const component_account_name = ....
const is_valid_account = AccountUtils.isValidAccount(component_account_name);
if (!is_valid_account) {
// Display message here about account not being right.
}
const account = is_valid_account ? component_account_name : flags.account;
// Pass the above var into the account function
```

src/architect/user/user.utils.ts
```diff
@@ -1,10 +1,19 @@
 import AppService from "../../app-config/service";
+import Account from "../account/account.entity";
+import User from "./user.entity";
+
+export default interface Membership {
```
mueschm, line 5: Nice! Leaving the code better than when you found it.

mueschm (changes requested):
This is also missing a unit test

PR #735, opened 2022-10-12
Init volume

mueschm:
DO NOT MERGE BEFORE K8S OVERRIDE BRANCH
## Overview
This diff adds two big features to the system.
1. Adds a new plugin system that allows us to download and use external binaries.
2. Add support for uploading a volume to our registry to be attached to a cluster during runtime.

## Changes
1. PluginManager: Manages all external binaries that our CLI relies on. Downloads and uses the correct version for the correct platform type.
2. OrasPlugin: Our first plugin that handles Oras. This allows us to send arbitrary payloads up to our system.
3. Volume Spec: Host path can now be used outside the debug section of the spec
4. Volumes: We now upload volumes outside the debug spec to our registry to be used. These get attached to the component spec.

## Testing
1. Run unit tests
2. Manual Testing. Tons of manual testing. Mostly done on my DO cluster.
3. 
To test quickly you can do the following. Go to the react-app example. In the api service move the volume to outside the debug block. Change the host_path to something like `/architect`. Then register and deploy the example app. Once done exec into the api and check the folder to see if it has the correct contents.


src/commands/register.ts
```diff
@@ -109,6 +115,38 @@ export default class ComponentRegister extends BaseCommand {
...
+  private async uploadVolume(component_path: string, component_name: string, service_name: string, volume_name: string, volume: VolumeSpec, account: Account): Promise<VolumeSpec> {
+    const oras_plugin = await PluginManager.getPlugin<OrasPlugin>(this.app.config.getPluginDirectory(), OrasPlugin);
+
+    if (!volume.host_path) {
+      return classToClass(volume);
+    }
+
+    const base_folder = path.join(os.tmpdir(), `/${account.name}`);
+    fs.mkdirpSync(base_folder);
+    const file_name = `volume--${component_name}-${service_name}-${volume_name}`;
```
tjhiggins, line 128: Can we have the name match the format we use for services?

```ts
`${account.name}/${component_name}.services.${service_name}.volumes.${volume_name}`
```
See the existing getImage method.

src/commands/register.ts
```diff
@@ -109,6 +115,38 @@ export default class ComponentRegister extends BaseCommand {
...
+    const file_path = `${account.name}/${file_name}`;
+    const registry_url = new URL(`/${file_path}:latest`, 'http://' + this.app.config.registry_host);
```
tjhiggins, line 130: tag shouldn't be hardcoded to latest?

src/commands/register.ts
```diff
@@ -109,6 +115,38 @@ export default class ComponentRegister extends BaseCommand {
...
+
+    const updated_volume = classToClass(volume);
+    const host_path = path.resolve(component_path, untildify(updated_volume.host_path as string));
```
tjhiggins, line 133: is the `as string` necessary?

src/commands/register.ts
```diff
@@ -109,6 +115,38 @@ export default class ComponentRegister extends BaseCommand {
...
+    updated_volume.host_path = registry_url.href.replace('http://', '');
+
+    const output_file_location = path.join(os.tmpdir(), `${file_path}.tar`);
```
tjhiggins, line 136: Do we cleanup the file?

tjhiggins: This is a good pkg for handling tmp dir/files and it handles cleanup automatically
https://www.npmjs.com/package/tmp

tjhiggins (changes requested):

PR #736, opened 2022-10-12
Better error message to the user when a deploy prompt fails in CI

mtgoncalves1:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/530.  
Show a more descriptive error message to the user when a deploy prompt fails in CI.

## Changes
Add a check for the deploy prompt and throw an error message.

## Tests
- Create a Github repo of any project
- Set Github secrets for `ARCHITECT_EMAIL` and `ARCHITECT_PASSWORD`
- Add a workflow file (Ex: .github/workflows/workflow.yml)
- Push some changes to trigger the CI job


src/common/utils/prompt-utils.ts
```diff
@@ -39,6 +39,9 @@ export default class PromptUtils {
       }
       for (const prompt of prompts) {
         if ((prompt.when && prompt.default === undefined) || prompt.when === undefined) {
+          if (prompt.name === 'deploy' && prompt.type === 'confirm') {
```
ryan-cahill, line 42: There are a few other places within the CLI where we use `confirm` prompts, so ideally this would be a solution that would cover current and future cases/commands. Let me know if you'd like me to have a look too and we can see what we can come up with

ryan-cahill (changes requested):

src/common/utils/prompt-utils.ts
ryan-cahill: @mtgoncalves1 TJ had the idea of printing the error in the `when` for each necessary prompt. For example

```ts
const confirmation = await inquirer.prompt({
      type: 'confirm',
      name: 'deploy',
      message: 'Would you like to apply?',
      when: () => {
        if (isCI) {
          this.error('--auto-approve is required in ci pipelines');
        }
        return true;
      },
    });
```

What are your thoughts on that? If we're going to use custom error messages per prompt, the error should just be part of it rather than in the util file

mtgoncalves1: I like this since it is quite hard to figure out where the error is coming from.

src/commands/clusters/create.ts
```diff
@@ -111,7 +112,16 @@ export default class ClusterCreate extends BaseCommand {
         type: 'input',
         name: 'cluster',
         message: 'What would you like to name your new cluster?',
-        when: !args.cluster,
+        when: () => {
+          if (args.cluster) {
+            return false;
+          }
+
+          if (isCi) {
```
mueschm, line 120: Should this be CI or should we throw on all TTY? Is there more context here @ryan-cahill

test/commands/platforms/create.test.ts
```diff
@@ -128,6 +129,9 @@ describe('platform:create', function () {
         application_install: true,
       }
     })
+    .stub(PromptUtils, 'allowWhen', () => {
```
mueschm, line 132: Why does this return a string here?

mtgoncalves1: My bad. It should return a boolean.

test/test-setup.ts
```diff
@@ -5,9 +5,6 @@ import 'reflect-metadata';
 import sinon from 'sinon';
 import CredentialManager from '../src/app-config/credentials';
 import PortUtil from '../src/common/utils/port';
-import PromptUtils from '../src/common/utils/prompt-utils';
-
-PromptUtils.disable_prompts();
```
ryan-cahill, line 10: Looks like the `disable_prompts` function can be removed as it's no longer used anywhere

src/commands/clusters/create.ts
ryan-cahill: We might as well add a check for `process.env.isTTY` since it's in the [inquirer docs](https://github.com/SBoudrias/Inquirer.js/#use-in-non-interactive-environments)

ryan-cahill (changes requested):

src/commands/clusters/create.ts
mtgoncalves1: `process.env.isTTY` is undefined. I ended up just re-using `prompts_available()`.

ryan-cahill (approved):

tjhiggins:
Pretty sure we cannot merge arc- branches into rc. Can you create a new branch with just your changes? (No changes to readme/package.json)

mtgoncalves1:
New branch so it can be merged to rc: https://github.com/architect-team/architect-cli/compare/rc...530-ci-prompt-err-message

PR #742, opened 2022-10-28
Superset updates

ryan-cahill:
## Overview
Converts the testing superset architect.yml file to a deployable component with the hello-world demo as a dependency. Part of https://gitlab.com/architect-io/architect-ci/-/issues/9 

## Changes
* Modified superset architect.yml file to be deployable both locally and to the cloud
* Moved files that the superset is dependent on to the test folder under /mocks/superset and /mocks/dependency
* Updated some tests to work with the new superset file

## Tests
* Deployed both locally and remotely
* Existing tests pass


test/mocks/superset/stateful-backend/src/index.js
```diff
@@ -0,0 +1,73 @@
+const winston = require('winston');
```
tjhiggins, line 1: Can we reference the examples directory instead of copying?

tjhiggins (changes requested):

PR #745, opened 2022-11-01
Superset updates

ryan-cahill:
## Overview
Closes https://gitlab.com/architect-io/architect-ci/-/issues/9 by running workflows that check that the superset runs both locally and remotely

## Changes
* Created GitHub workflow for local and remote deployments. Checks that
  * the superset runs locally
  * the superset runs remotely
  * certs are generated for the remote deployment
  * network policies are enforced on the remote cluster


.github/workflows/test-deploy.yml
```diff
@@ -0,0 +1,69 @@
...
+      - run: docker volume create volume-key
+      - run: ./bin/dev link examples/hello-world/architect.yml
+      - run: ./bin/dev dev -e superset architect-cli/test/mocks/superset/architect.yml --browser=false -s param_unset=test > ./tmp.txt 2>&1 &
+      - run: sh architect-cli/scripts/server_online.sh
+      - name: Check that the app is accessible locally
+        run: curl --fail -S -I https://hello.localhost.architect.sh
+
+  test_remote:
+    name: Superset remote deployment
+    env:
```
tjhiggins, line 40: Talked offline, but we should set an env here to use in the `architect deploy`. Maybe override hello world text and check the response?

.github/workflows/test-deploy.yml
```diff
@@ -0,0 +1,69 @@
...
+
+on:
+  pull_request:
+    types:
+      - opened
+    branches:
+      - main
+
+jobs:
+  test_local:
```
tjhiggins, line 11: What do you think about running test local on all prs? How long does it take to run?

ryan-cahill: It could take somewhere between 3-5 minutes and I don't see why we couldn't. To be clear, would the trigger then be a push to any branch OR a PR opened to main and/or rc?

tjhiggins: The trigger for test local would be prs to rc/main. The same rules as for running out other tests. Not just when the pr opens.

.github/workflows/test-deploy.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Check that the app is accessible locally
+        run: curl --fail -S -I https://hello.localhost.architect.sh
+      - name: Check that the app returns the expected response
+        run: curl --fail -S https://hello.localhost.architect.sh | grep "Hello Architect!"
+      - name: Run a local task
+        run: ./bin/dev task superset curler-build --local -e superset
+
+  test_remote:
+    name: Superset remote deployment
+    if: github.ref_name == 'main' || github.ref_name == 'rc'
```
tjhiggins, line 35: Lets make this only for main

tjhiggins, line 35: Lets make this only for main

tjhiggins (changes requested):

.github/workflows/test-deploy.yml
tjhiggins: Also is ref_name the branch you'd be merging into or the current branch?

ryan-cahill: That's the branch that was pushed to. We use the same logic in the publish workflow

https://github.com/architect-team/architect-cli/blob/rc/.github/workflows/publish.yml#L58

tjhiggins: We want these tests to run when we open a pr from `rc` into `main`. The idea being that we don't allow anything to go into mainline if the api hasn't been updated. Sorry if that wasn't clear. 

So if we open a pr on `rc` then I'd imagine `github.ref_name` would be `rc`. According to their docs `The short ref name of the branch or tag that triggered the workflow run.`.

I think this is a scenario where we actually want to use `pull_request_target` as the trigger. You'd need to confirm, but that might make things awkward since we want the remote test to only run on prs to main, but the local tests to run on all prs.

tjhiggins: Oh looks like you can target specific events:
https://docs.github.com/en/actions/learn-github-actions/contexts#example-usage-of-the-github-context

ryan-cahill: Wouldn't the issue in using `pull_request_target` be the following from the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)? So using that wouldn't use any changes that we want to merge

> This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does

PR #746, opened 2022-11-01
556 ignore 409 status for env create

devinsag:
## Overview

Please see https://gitlab.com/architect-io/architect-cli/-/issues/556

 To enable cleaner CI configurations and to ensure that jobs don't pass on real errors, we should only warn on 409s when creating an environment. 

## Changes
add validateStatus to environment:create command to manually append 409 as a passing status code. Add warning if http reply status code is 409.

### test/commands/environments/create.test.ts

```typescript
  create_duplicate_environment
    .env({ ARCHITECT_CLUSTER: mock_cluster.name })
    .command(['environment:create', mock_env.name, '-a', mock_account.name])
    .it('should print warning when an environment name is already in use for an account', ctx => {
      expect(ctx.stderr).to.contain(`Unable to create new environment '${mock_env.name}'. Environment name already in use for account '${mock_account.name}'`);
    });
```

## Docs
TODO?

## Pictures (update 11/8/22)
image

### DIFF
--- a/src/commands/environments/create.ts
+++ b/src/commands/environments/create.ts
@@ -86,10 +86,14 @@ export default class EnvironmentCreate extends BaseCommand {
     });
 
     const environment_url = `${this.app.config.app_host}/${account.name}/environments/${environment_name}`;
-    const environment_create_outcome = _environment_already_exists ?
-      chalk.yellow(`Unable to create new environment '${environment_name}'. Environment name already in use for account '${account.name}'`) :
-      chalk.green(`Environment created: ${environment_url}`);
 
-    CliUx.ux.action.stop(environment_create_outcome);
+    CliUx.ux.action.stop();
+
+    if (_environment_already_exists) {
+      this.warn(`Unable to create new environment '${environment_name}'. Environment name already in use for account '${account.name}'`);
+      return;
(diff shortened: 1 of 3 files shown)


src/commands/environments/create.ts
```diff
@@ -76,10 +76,20 @@ export default class EnvironmentCreate extends BaseCommand {
...
+        _environment_already_exists = status === 409;
+        return status === 201 || _environment_already_exists;
+      },
+    });
 
     const environment_url = `${this.app.config.app_host}/${account.name}/environments/${environment_name}`;
-    CliUx.ux.action.stop();
-    this.log(chalk.green(`Environment created: ${environment_url}`));
+    const environment_create_outcome = _environment_already_exists ?
+      chalk.yellow(`Unable to create new environment '${environment_name}'. Environment name already in use for account '${account.name}'`) :
```
tjhiggins, line 90: instead of chalk.yellow you can do `this.warn`

devinsag: refactored to use `this.warn(...)` instead of using `this.log(chalk.yellow(...));`

TylerAldrich (approved):

PR #749, opened 2022-11-03
Remove account name from component name in tests, examples, and spec descriptions

TylerAldrich:
## Overview

The primary code change was updating `LocalDependencyManager` to take in `account` as a parameter in its constructor. This is because now, `this.account` always needs to be available. In practice, we always set `manager.account = {account name}` anyways, this update just prevents us from making future mistakes.

Otherwise, there were a few small tweaks to places that used `component_account_name || this.account` to always ignore `component_account_name` if it exists whenever we are using `ComponentVersionSlugUtils.parse` with user inputted values.

In the backend, we always handle components with `account-name/component-name` format so there's no ambiguity. We should in the long term stop using `ComponentSlugUtils` to build / parse the `account-name/component-name` component slugs used elsewhere. `ComponentSlugUtils` can drop the `component_account_name` completely if it's fully deprecated and `architect.yml`'s aren't accepted with an account name anymore, and we can perhaps build an `AccountComponentSlug` class to handle creating/parsing strings we expect to _always_ contain account_name separate from user component names that don't have account_name.

The rest of the changes are simply updating tests to remove the account name in most places, with the exception being the account scoping test that actually verifies various permutations of names with / without the account name work together as we'd expect and still want for now.

A combination of this PR and https://github.com/architect-team/architect-documentation/pull/24/files will remove the suggestions of adding `account-name/component-name` from our docs/spec - it's already barely mentioned but there's a lot of various examples that still use it.

### DIFF
--- a/src/dependency-manager/spec/utils/slugs.ts
+++ b/src/dependency-manager/spec/utils/slugs.ts
@@ -74,7 +74,7 @@ function parseCurry<S extends string, P extends ParsedSlug>() {
 }
 
 export class ComponentSlugUtils extends SlugUtils {
-  public static Description = `${Slugs.ArchitectSlugDescription}; optionally can be prefixed with a valid Architect account and separated by a slash (e.g. architect/component-name).`;
+  public static Description = Slugs.ArchitectSlugDescription;
 
   static RegexName = `(?:(?<component_account_name>${Slugs.ArchitectSlugRegexBase})${Slugs.NAMESPACE_DELIMITER})?(?<component_name>${Slugs.ArchitectSlugRegexBase})`;
   static RegexInstance = `(?:${Slugs.INSTANCE_DELIMITER}(?<instance_name>${Slugs.ComponentTagRegexBase}))?`;
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -78,7 +78,7 @@ export default class Deploy extends DeployCommand {
   static description = 'Create a deploy job on Architect Cloud';
 
   static examples = [
-    'architect deploy myaccount/mycomponent:latest',
+    'architect deploy mycomponent:latest',
     'architect deploy ./myfolder/architect.yml --secret-file=./mysecrets.yml --environment=myenvironment --account=myaccount --auto-approve',
   ];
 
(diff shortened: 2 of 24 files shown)


src/dependency-manager/spec/utils/slugs.ts
```diff
@@ -74,7 +74,7 @@ function parseCurry<S extends string, P extends ParsedSlug>() {
 }
 
 export class ComponentSlugUtils extends SlugUtils {
-  public static Description = `${Slugs.ArchitectSlugDescription}; optionally can be prefixed with a valid Architect account and separated by a slash (e.g. architect/component-name).`;
+  public static Description = Slugs.ArchitectSlugDescription;
```
tjhiggins, line 77: I'm worried this will still create ambiguity if we don't remove the component_account_name from ParsedComponentSlug/ParsedComponentVersionSlug. 

Maybe we don't do that as part of this pr. There are enough changes here where I'd like to not break anything and fix the error msgs to not include account.

tjhiggins:
@TylerAldrich Wanted to confirm that this won't break any existing components with an account prefix. If so then we can merge, but we should make a ticket to audit existing customers that use a prefix and reach out to them to update.

src/dependency-manager/spec/utils/slugs.ts
TylerAldrich: Yeah so that's something that I mentioned in the description above and is kind of complicated - 

> ComponentSlugUtils can drop the component_account_name completely if it's fully deprecated and architect.yml's aren't accepted with an account name anymore, and we can perhaps build an AccountComponentSlug class to handle creating/parsing strings we expect to always contain account_name separate from user component names that don't have account_name.

(I actually meant the `Parsed` version of the slug here)

The reason we can't remove it now is because we use the `ParsedComponentSlug` and `ParsedComponentVersionSlug` to also extract the account name after we call `ComponentSlugUtils.build(this.account, component_name, tag)` in both the CLI and the API. So, the `ComponentSlugUtils` at the moment needs to match `ParsedComponentSlug` with regard to including the account_name at all times.

I think what we may need to do is create another `AccountParsedComponentSlug` we can use to `.build()` slugs with an account name included for backwards compatibility in the API, but also do a bunch of work in the API to separate account name from the component slug itself. I tried to keep this PR pretty focused on just removing all user facing `account-name/component-name` stuff so we can deal with the rest of the mess more cleanly

TylerAldrich:
> @TylerAldrich Wanted to confirm that this won't break any existing components with an account prefix. If so then we can merge, but we should make a ticket to audit existing customers that use a prefix and reach out to them to update.

Yeah this won't break anything - I tested this pretty extensively with both `account-name/component-name` and just `account-name` for both component names and the names of dependencies, deployed those components with the current master branch of our API and everything works as it did before still

PR #756, opened 2022-11-08
[WIP] ensure that we can run exec for detached local environments

ryan-cahill:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/560 and enables users to run `exec` against a container that was started with `architect dev -d ....`

## Changes
* Fixed some linting issues
* Avoided removing the compose file when the compose process was started in detached mode
* Added line to remove the file (or do nothing) when the process is stopped with `architect stop ....`

## Tests
* Added tests
* Ran local deployments in both normal and detached states, ran exec against them, then stopped them

### DIFF
--- a/src/commands/dev/stop.ts
+++ b/src/commands/dev/stop.ts
@@ -52,7 +52,7 @@ export default class DevStop extends BaseCommand {
   async runComposeStop(project: string): Promise<void> {
     const compose_file = DockerComposeUtils.buildComposeFilepath(this.app.config.getConfigDir(), project);
     await DockerComposeUtils.dockerCompose(['-p', project, '-f', compose_file, 'stop']);
-    fs.removeSync(compose_file);
+    fs.removeSync(compose_file); // remove compose file if dev was started in detached mode
   }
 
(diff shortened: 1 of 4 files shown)


src/commands/dev/stop.ts
```diff
@@ -96,8 +97,8 @@ export default class DevStop extends BaseCommand {
       });
     } else {
       // If there's no socket at all, dev is running in detached mode and we can stop it with docker compose without
-      // working about the health check restarting anything.
-      this.runComposeStop(name);
```
TylerAldrich, line 100: I think I purposely didn't await this and i'm trying to now remember why...

TylerAldrich: Ah - right - I think that when `await`-ing this, we never actually do the whole ``Waiting for ${name} to stop...``  below this and `waitForEnviromentToStop` is what we actually want to block on

ryan-cahill: I see what you mean. I just moved that line to provide the same experience

PR #757, opened 2022-11-10
566 local dev secrets

mtgoncalves1:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/566.  
Allow `architect dev` command to pull in secrets from an environment to be used for local dev.

## Changes
Get secrets from a production environment and merge to other component secrets.

## Tests
- Create a secret in a production environment
- Use that secret in your application `architect.yml`
- Run `$ architect dev . --secrets-env=my-environment -a my-account`

### DIFF
--- a/src/commands/dev/index.ts
+++ b/src/commands/dev/index.ts
@@ -12,6 +12,7 @@ import path from 'path';
 import { ArchitectError, buildSpecFromPath, ComponentSlugUtils, ComponentSpec, ComponentVersionSlugUtils, Dictionary } from '../../';
 import AccountUtils from '../../architect/account/account.utils';
 import { EnvironmentUtils } from '../../architect/environment/environment.utils';
+import SecretUtils from '../../architect/secret/secret.utils';
 import { default as BaseCommand } from '../../base-command';
 import LocalDependencyManager, { ComponentConfigOpts } from '../../common/dependency-manager/local-manager';
 import { DockerComposeUtils } from '../../common/docker-compose';
@@ -20,6 +21,7 @@ import { RequiresDocker } from '../../common/docker/helper';
 import DeployUtils from '../../common/utils/deploy.utils';
 import { booleanString } from '../../common/utils/oclif';
 import PortUtil from '../../common/utils/port';
+import { SecretsDict } from '../../dependency-manager/secrets/type';
 import LocalPaths from '../../paths';
 
 type TraefikHttpService = {
@@ -246,6 +248,7 @@ export default class Dev extends BaseCommand {
 
   static examples = [
     'architect dev ./mycomponent/architect.yml',
+    'architect dev ./mycomponent/architect.yml -a myaccount --secrets-from=myenvironment',
     'architect dev --port=81 --browser=false --debug=true --secret-file=./mycomponent/mysecrets.yml ./mycomponent/architect.yml',
   ];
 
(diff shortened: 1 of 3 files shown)


src/commands/dev/index.ts
```diff
@@ -277,6 +281,9 @@ export default class Dev extends BaseCommand {
       default: [],
       sensitive: false,
     }),
+    'secrets-from': Flags.string({
```
tjhiggins, line 284: We should come to a consensus on the flag name. Guessing you missed the slack conv for this ticket. Tyler had proposed `secret-env`, but that means we would need to provide flags for each entity type.

tjhiggins (changes requested):

src/commands/dev/index.ts
mtgoncalves1: My bad. Would you like me to extend the support for the account and cluster secrets as well?

tjhiggins: Prob worth another conv in engineering. Adding 3 flags feels like a lot. I'd probably be okay with only supporting secrets from environments, but that might not be everyone's expectation.

src/commands/dev/index.ts
```diff
@@ -629,8 +648,16 @@ $ architect dev -e new_env_name_here .`));
     }
 
     const interfaces_map = DeployUtils.getInterfacesMap(flags.interface);
+
+    let env_secrets: SecretsDict = {};
+    if (flags['secrets-env']) {
+      const account = await AccountUtils.getAccount(this.app, flags.account, { ask_local_account: false });
+      env_secrets = await this.getEnvironmentSecrets(account, flags['secrets-env']);
+      flags.account = account.name;
```
mueschm, line 656: What is this used for?

src/common/utils/deploy.utils.ts
```diff
@@ -76,7 +78,7 @@ export default class DeployUtils {
         component_secrets['*'] = {};
       }
       // Shallow merge to ensure CLI arguments replace anything from the secrets file
-      component_secrets['*'] = { ...component_secrets['*'], ...extra_secrets };
+      component_secrets['*'] = { ...component_secrets['*'], ...extra_secrets } as Dictionary<SecretType>;
```
mueschm, line 81: Why do we need the cast here?

test/commands/dev/index.test.ts
```diff
@@ -947,6 +976,59 @@ describe('local dev environment', function () {
       const hello_world_service = runCompose.firstCall.args[0].services[hello_api_ref] as any;
       expect(hello_world_service.environment.compose_escaped_variable).to.equal('variable_split_$$_with_dollar$$signs');
     })
+  
+  test
```
mueschm, line 980: You need a few more tests.
1. What happens if they give a bad env name?
2. Make sure we ask for an account if the account is not given.

mueschm (changes requested):

src/common/utils/deploy.utils.ts
mtgoncalves1: This is because DeployUtils.getExtraSecrets(individual_secrets) returns `Dictionary<string | number | undefined>` and not `Dictionary<SecretType>`. I will change the casting to `const extra_secrets = DeployUtils.getExtraSecrets(individual_secrets) as Dictionary<SecretType>;` instead.

src/commands/dev/index.ts
mtgoncalves1: It's not really used for anything so I will remove it.

PR #759, opened 2022-11-16
561 log errors when liveness probe fails

devinsag:
## Overview
Please see https://gitlab.com/architect-io/architect-cli/-/issues/561

If your liveness probe fails before it hits your service (e.g. you don't have curl installed on the container and it's trying to curl a URL), it silently fails and will just restart your container without any indication as to why. We can possibly use docker status or similar to get the reason the probe failed and can more clearly surface the issue to the user so that they know what is going wrong.

## Changes
Add  Health property to DockerInspect  interface.  Check for failing liveprobe, and conditionally fetch logs from docker inspect. print to screen.

## Docs
N/A

## Pictures
image

### DIFF
--- a/src/common/docker-compose/index.ts
+++ b/src/common/docker-compose/index.ts
@@ -481,6 +481,29 @@ export class DockerComposeUtils {
   }
 
   /**
+  * Prints an error message and updates given a set of container ID values and current container state
+  * Liveness node fails.
+  */
+  public static async printLivenessNodeFailureErrors(container_state: any): Promise<void> {
+    // this is an inneficient but thorough approach to checking all local docker inspect states
+    const local_environments = await this.getLocalEnvironmentContainerMap();
+    for (const local_environment_name of Object.keys(local_environments)) {
+      const container_docker_inspects = local_environments[local_environment_name].filter(inspect => {
+          const logs = inspect.State?.Health?.Log;
+          return container_state?.ExitCode !== 0 || logs[logs.length - 1].ExitCode !== 0;
+        });
+
+      for (const container_docker_inspect of container_docker_inspects) {
+        const log_size = (container_docker_inspect?.State?.Health?.Log || []).length;
+        console.log(chalk.red(`The liveness probe has encountered an error trying to start the service '${container_docker_inspect.Config.Labels['com.docker.compose.service']}'.`));
+        if (log_size > 0) {
+          console.log(chalk.red(`ERROR: ${container_docker_inspect.State.Health.Log[log_size - 1].Output}`));
+        }
+      }
+    }
+  }
+
+  /**
   * Runs `docker inspect` on a container ID and returns the resulting json as a DockerInspect object.
   */
   public static async getContainerInfo(container_id: string): Promise<DockerInspect | undefined> {
@@ -674,6 +697,7 @@ export class DockerComposeUtils {
     }
 
     const service_data_dictionary: Dictionary<{ last_restart_ms: number }> = {};
+    const container_states = JSON.parse((await DockerComposeUtils.dockerCompose(['-f', compose_file, '-p', environment_name, 'ps', '--format', 'json'])).stdout);
 
     // If the last time this loop runs, a container was restarted, we may have to run `docker compose stop`
     // because the restart can happen after the compose process was killed.
@@ -681,11 +705,11 @@ export class DockerComposeUtils {
     while (!should_stop()) {
       try {
         restarted = false;
-        const container_states = JSON.parse((await DockerComposeUtils.dockerCompose(['-f', compose_file, '-p', environment_name, 'ps', '--format', 'json'])).stdout);
         for (const container_state of container_states) {
           const id = container_state.ID;
           const full_service_name = container_state.Service;
 
+          await this.printLivenessNodeFailureErrors(container_state);
           const service_ref = service_ref_map[full_service_name];
           if (!service_ref) {
             continue;
@@ -699,16 +723,6 @@ export class DockerComposeUtils {
 
           const bad_state = state !== 'running' && container_state.ExitCode !== 0;
           const bad_health = health === 'unhealthy';
-
-          if (health === '' && state === 'exited') {
-            const container_docker_inspect: DockerInspect | undefined = await this.getContainerInfo(container_state.ID);
-            if (container_docker_inspect && container_docker_inspect.State.Health.Log?.length > 0) {
-              const log_size = container_docker_inspect.State.Health.Log.length;
-              console.log(chalk.red(`\nThe liveness probe has encountered an error trying to start the service '${full_service_name}'.`));
-              console.log(chalk.red(`The service may try to restart and/or never reach a stable state`));
-              console.log(chalk.red(`ERROR: ${container_docker_inspect.State.Health.Log[log_size - 1].Output}\n`));
-            }
-          }
 
           if (!service_data_dictionary[service_ref]) {
             service_data_dictionary[service_ref] = {
--- a/src/common/docker-compose/index.ts
+++ b/src/common/docker-compose/index.ts
@@ -486,7 +486,7 @@ export class DockerComposeUtils {
   public static async getContainerInfo(container_id: string): Promise<DockerInspect | undefined> {
     if (!container_id) return;
     const inspect_cmd = await docker(['inspect', '--format=\'{{json .}}\'', container_id], { stdout: false });
-    return JSON.parse(inspect_cmd.stdout.substring(1, inspect_cmd.stdout.length - 1)) as DockerInspect;
+    return await JSON.parse(inspect_cmd.stdout.substring(1, inspect_cmd.stdout.length - 1)) as DockerInspect;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/common/docker-compose/index.ts
```diff
@@ -691,6 +700,16 @@ export class DockerComposeUtils {
...
 
+          if (health === '' && state === 'exited') {
+            const container_docker_inspect: DockerInspect | undefined = await this.getContainerInfo(container_state.ID);
+            if (container_docker_inspect && container_docker_inspect.State.Health.Log?.length > 0) {
+              const log_size = container_docker_inspect.State.Health.Log.length;
+              console.log(chalk.red(`\nThe liveness probe has encountered an error trying to start the service '${full_service_name}'.`));
+              console.log(chalk.red(`The service may try to restart and/or never reach a stable state`));
+              console.log(chalk.red(`ERROR: ${container_docker_inspect.State.Health.Log[log_size - 1].Output}\n`));
+            }
+          }
```
TylerAldrich, line 711: One problem doing this is that we will only ever see the error when you initially launch `architect dev` and it's unclear if this is going to work for things that take longer to start up. The `state` just happens to be "exited" initially if you've run `architect dev` before, but on a clean first run with no prior containers existing, you actually don't get an error at all. 

See some logs here, that I grabbed after clearing my existing containers and running `architect dev .` with the go project example:
```
Once the containers are running they will be accessible via the following urls:
https://app.localhost.architect.sh:443/ => go-demo--app

http://localhost:443/ => gateway:443
http://localhost:8080/ => gateway:8080
http://localhost:50000/ => go-demo--app:8000
http://localhost:50001/ => go-demo--app-db:5432

Starting containers...

go-demo--app-db-1  |  The files belonging to this database system will be owned by user "postgres".
go-demo--app-db-1  |  This user must also own the server process.
go-demo--app-db-1  |
go-demo--app-db-1  |  The database cluster will be initialized with locale "en_US.utf8".
go-demo--app-db-1  |  The default database encoding has accordingly been set to "UTF8".
go-demo--app-db-1  |  The default text search configuration will be set to "english".
go-demo--app-db-1  |
go-demo--app-db-1  |  Data page checksums are disabled.
go-demo--app-db-1  |
go-demo--app-db-1  |  fixing permissions on existing directory /var/lib/postgresql/data ... ok
go-demo--app-db-1  |  creating subdirectories ... ok
go-demo--app-db-1  |  selecting dynamic shared memory implementation ... posix
go-demo--app-db-1  |  selecting default max_connections ... 100
go-demo--app-db-1  |  selecting default shared_buffers ... 128MB
go-demo--app-db-1  |  selecting default time zone ... Etc/UTC
go-demo--app-db-1  |  creating configuration files ... ok
gateway-1          |  time="2022-11-16T17:40:01Z" level=info msg="Configuration loaded from flags."
go-demo--app-db-1  |  running bootstrap script ... ok
go-demo--app-1     |  [00] Starting service
go-demo--app-1     |  [00]
go-demo--app-1     |  [00] 2022/11/16 17:40:01 /go/pkg/mod/gorm.io/gorm@v1.24.1/gorm.go:206
go-demo--app-1     |  [00] [error] failed to initialize database, got error failed to connect to `host=go-demo--app-db user=architect database=app-db`: dial error (dial tcp 192.168.16.2:5432: connect: connection refused)
go-demo--app-1     |  [00] 2022/11/16 17:40:01 Unable to connect to db
go-demo--app-1     |  [00] (error exit: exit status 1)
go-demo--app-db-1  |  performing post-bootstrap initialization ... ok
go-demo--app-db-1  |  syncing data to disk ... initdb: warning: enabling "trust" authentication for local connections
go-demo--app-db-1  |  You can change this by editing pg_hba.conf or using the option -A, or
go-demo--app-db-1  |  --auth-local and --auth-host, the next time you run initdb.
go-demo--app-db-1  |  ok
go-demo--app-db-1  |
go-demo--app-db-1  |
go-demo--app-db-1  |  Success. You can now start the database server using:
go-demo--app-db-1  |
go-demo--app-db-1  |      pg_ctl -D /var/lib/postgresql/data -l logfile start
go-demo--app-db-1  |
go-demo--app-db-1  |  waiting for server to start....2022-11-16 17:40:01.960 UTC [48] LOG:  starting PostgreSQL 12.12 (Debian 12.12-1.pgdg110+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit
go-demo--app-db-1  |  2022-11-16 17:40:01.961 UTC [48] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
go-demo--app-db-1  |  2022-11-16 17:40:01.974 UTC [49] LOG:  database system was shut down at 2022-11-16 17:40:01 UTC
go-demo--app-db-1  |  2022-11-16 17:40:01.979 UTC [48] LOG:  database system is ready to accept connections
go-demo--app-db-1  |   done
go-demo--app-db-1  |  server started
go-demo--app-db-1  |  CREATE DATABASE
go-demo--app-db-1  |
go-demo--app-db-1  |
go-demo--app-db-1  |  /usr/local/bin/docker-entrypoint.sh: ignoring /docker-entrypoint-initdb.d/*
go-demo--app-db-1  |
go-demo--app-db-1  |  waiting for server to shut down...2022-11-16 17:40:02.222 UTC [48] LOG:  received fast shutdown request
go-demo--app-db-1  |  .2022-11-16 17:40:02.224 UTC [48] LOG:  aborting any active transactions
go-demo--app-db-1  |  2022-11-16 17:40:02.225 UTC [48] LOG:  background worker "logical replication launcher" (PID 55) exited with exit code 1
go-demo--app-db-1  |  2022-11-16 17:40:02.226 UTC [50] LOG:  shutting down
go-demo--app-db-1  |  2022-11-16 17:40:02.237 UTC [48] LOG:  database system is shut down
go-demo--app-db-1  |   done
go-demo--app-db-1  |  server stopped
go-demo--app-db-1  |
go-demo--app-db-1  |  PostgreSQL init process complete; ready for start up.
go-demo--app-db-1  |
go-demo--app-db-1  |  2022-11-16 17:40:02.334 UTC [1] LOG:  starting PostgreSQL 12.12 (Debian 12.12-1.pgdg110+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit
go-demo--app-db-1  |  2022-11-16 17:40:02.334 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
go-demo--app-db-1  |  2022-11-16 17:40:02.334 UTC [1] LOG:  listening on IPv6 address "::", port 5432
go-demo--app-db-1  |  2022-11-16 17:40:02.337 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
go-demo--app-db-1  |  2022-11-16 17:40:02.349 UTC [76] LOG:  database system was shut down at 2022-11-16 17:40:02 UTC
go-demo--app-db-1  |  2022-11-16 17:40:02.354 UTC [1] LOG:  database system is ready to accept connections
ERROR: go-demo.services.app has encountered an error and is being restarted.
go-demo--app-1     |  Interrupted (terminated). Cleaning up children...
0b0a3b3c10364dc2f80257c2639e9e6072e744a4b56d85cf9ed06e432dc0197f
go-demo--app-1     |  [00] Starting server on port: 8000
^CGracefully stopping..... Please Wait.....
```

A bit verbose due to the db taking a bit to get settled, but we don't see this error message at all and the go service just silently restarts.


Here's what I think we should do to remedy this:

1. We need the container names or ids to run `docker inspect`
2. Docker inspect has the State, Status, Health, ExitCode, etc. information that `docker compose ps` is giving us, AND it provides this Health.Log property - so `docker inspect` is kind of a superset of the `compose ps` information we were using prior

Because of this, my proposal is that we:
- Run the function that sets `container_states` outside of the `while` loop to get a list of container IDs
- Inside the while loop, run `docker inspect` (which allows multiple IDs, e.g. `docker inspect 0b0a3b3c1036 7ad42dbba6be`) and use that information to get ID, full_service-name, state, health, etc.
- Every iteration of this loop we can then check to see if the latest Health.Log entry has ExitCode != 0, and if so, we log this error message.

For the go example, we have this data for example:
```
  "Log": [
                    {
                        "Start": "2022-11-16T17:40:16.435031493Z",
                        "End": "2022-11-16T17:40:16.497305733Z",
                        "ExitCode": 1,
                        "Output": "/bin/sh: curl: not found\n"
                    },
```

so we'd see the latest entry has ExitCode 1, and show this warning.

This _should_ then display the error to the user every time we iterate through this loop (which is every 5s). If the errors go away, it should then stop printing out the previous health check errors.

TylerAldrich (changes requested):
I think for testing this, we're almost completely just testing the output of `docker inspect` in various scenarios, so we should test with a few examples manually rather than writing unit tests that can mock `docker inspect` which won't buy us much.

Some things we want to check:

- If a user has no liveness probes defined, nothing changes for them - no new logged warnings/errors/etc (because obviously nothing can fail in this case)
- If a user has a liveness probe that continuously fails, we should be displaying the most recent failure every time we see it until it goes away
- If a user has a liveness probe that fails the first couple times and then succeeds, they should stop seeing errors. An example of this is a service that takes some amount of time to start up, but the liveness probe may start probing early. The first couple failures may be expected, but once the service warms up and is live, the liveness probe logging should go away.

PR #761, opened 2022-11-21
validator for looser validation on account names

ryan-cahill:
## Overview
Adds a validator for account names, part of https://gitlab.com/architect-io/architect-hub/-/issues/579

## Changes
* Added a validator for case-insensitive slugs to be used as a looser validator for account names or others

## Tests
* Added tests

### DIFF
--- a/src/dependency-manager/spec/utils/slugs.ts
+++ b/src/dependency-manager/spec/utils/slugs.ts
@@ -17,6 +17,10 @@ export class Slugs {
   public static ArchitectSlugRegexBase = `(?!-)(?!.{0,${Slugs.SLUG_CHAR_LIMIT}}--)[a-z0-9-]{1,${Slugs.SLUG_CHAR_LIMIT}}(?<!-)`;
   public static ArchitectSlugValidator = new RegExp(`^${Slugs.ArchitectSlugRegexBase}$`);
 
+  public static ArchitectSlugDescriptionCaseInsensitive = `must contain only alphanumeric and single hyphens or underscores in the middle; max length ${Slugs.SLUG_CHAR_LIMIT}`;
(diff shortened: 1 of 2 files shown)


src/dependency-manager/spec/utils/slugs.ts
```diff
@@ -17,6 +17,10 @@ export class Slugs {
   public static ArchitectSlugRegexBase = `(?!-)(?!.{0,${Slugs.SLUG_CHAR_LIMIT}}--)[a-z0-9-]{1,${Slugs.SLUG_CHAR_LIMIT}}(?<!-)`;
   public static ArchitectSlugValidator = new RegExp(`^${Slugs.ArchitectSlugRegexBase}$`);
 
+  public static ArchitectSlugDescriptionCaseInsensitive = `must contain only alphanumeric and single hyphens or underscores in the middle; max length ${Slugs.SLUG_CHAR_LIMIT}`;
+  public static ArchitectSlugRegexBaseCaseInsensitive = `(?!-)(?!.{0,${Slugs.SLUG_CHAR_LIMIT}}--)[A-Za-z0-9-]{1,${Slugs.SLUG_CHAR_LIMIT}}(?<!-)`;
+  public static ArchitectSlugValidatorCaseInsensitive = new RegExp(`^${Slugs.ArchitectSlugRegexBaseCaseInsensitive}$`);
```
tjhiggins, line 22: When I suggested that we lowercase the username - I was thinking we do that before running validation. Making the validation case insensitive doesn't make the underlying logic case insensitive which feels like of scary. Worried we will end up with different accounts for `Architect` and `architect` respectively.

tjhiggins (changes requested):

src/dependency-manager/spec/utils/slugs.ts
tjhiggins: I guess if this is only used client-side cli/UI it's fine. So maybe it doesn't matter.

ryan-cahill: I wouldn't expect for this to be used in the API. For the frontend, modifying the user's input before we validate it feels kind of dirty. Plus in adding this we also have a standardized description that comes along with it

PR #765, opened 2022-11-29
529 default register path

devinsag:
## Overview

please see: 
https://gitlab.com/architect-io/architect-cli/-/issues/529

>architect dev doesn't require you to specify an architect.yml and will assume you mean ./architect.yml. architect register doesn't do this - if you run architect register it will just do nothing, and you need to run architect register . instead. We should update the register command to support being run with no path similar to dev.

## Changes
Moved path resolution into run method to avoid having to resolve twice. 
Moved deduplication as last step after paths have been resolved.
Set default path as `./architect.yml` (same as architect dev command)

## Tests
```javascript
  mockArchitectAuth()
    .stub(fs, 'move', sinon.stub())
    .stub(ComponentRegister, 'registerComponent', sinon.stub().returns({}))
    .stdout({ print })
    .stderr({ print })
    .command(['register', '-a', 'examples'])
    .catch(e => {
      expect(e.message).contains(path.resolve(untildify('./architect.yml')));
    })
    .it('expect default project path to be ./architect.yml if not provided');
```

## Docs
N/A

## Pictures
N/A

### DIFF
--- a/src/commands/register.ts
+++ b/src/commands/register.ts
@@ -93,7 +93,7 @@ export default class ComponentRegister extends BaseCommand {
     const parsed = await super.parse(options, argv) as Interfaces.ParserOutput<F, A>;
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
-    parsed.args.component = [...new Set(parsed.argv)];
+    parsed.args.component = parsed.argv;
 
     return parsed;
   }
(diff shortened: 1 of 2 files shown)


src/commands/register.ts
```diff
@@ -92,25 +91,28 @@ export default class ComponentRegister extends BaseCommand {
...
   }
 
   @RequiresDocker({ buildx: true })
   async run(): Promise<void> {
     const { args, flags } = await this.parse(ComponentRegister);
-    for (const component of args.component) {
-      const config_path = path.resolve(untildify(component));
 
+    if (!args.component || args.component.length === 0) {
+      args.component = ['./architect.yml'];
```
TylerAldrich, line 106: We need to use `path.join('.', 'architect.yml')` here otherwise this wouldn't work on Windows properly

TylerAldrich (changes requested):

src/commands/register.ts
tjhiggins: `path.resolve` would also work

TylerAldrich: Oh does that also fix paths? this may work as is then because we do call path.resolve later... I can check on my windows machine real quick to verify before Devin changes anything

PR #766, opened 2022-11-29
fixing warning when dependencies are already registered

ryan-cahill:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/574 by making sure that there are actually dependencies to warn about before warning

## Changes
* Fixed array checks for warnings
* Used `\n` char in places where the warning should start after an empty line

## Tests
* Registered superset

### DIFF
--- a/src/commands/register.ts
+++ b/src/commands/register.ts
@@ -442,22 +442,19 @@ export default class ComponentRegister extends BaseCommand {
       }
     }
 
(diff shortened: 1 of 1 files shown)


TylerAldrich (approved):
I just saw this in that deployment log I screenshotted and was wondering what was up, this is very timely, thanks 😄

PR #768, opened 2022-12-02
571-allow-env-destroy-to-fail

devinsag:
Please see: https://gitlab.com/architect-io/architect-cli/-/issues/571

## Overview
This should allow users to avoid using `|| exit 0` or similar in CI jobs when removing environments. `environment:destroy` warns and exits by default when environment is not found to deregister.

## Changes
add new flag for environment: destroy, `--skip-deregistered`
set flag value properties: non-sensitive, default: true
add optional param to EnvironmentUtils.getEnvironment, `skip_deregistered=false`
print warning and return when environment is not found and skip_deregistered is provided.

## Tests
`destroy.test.ts`

```javascript
  mockArchitectAuth()
    .stub(PipelineUtils, 'pollPipeline', async () => null)
    .nock(MOCK_API_HOST, api => api
      .get(`/accounts/${mock_account.name}`)
      .reply(200, mock_account))
    .nock(MOCK_API_HOST, api => api
      .get(`/accounts/${mock_account.id}/environments/${failing_mock_env.name}`)
      .reply(200, failing_mock_env))
    .stdout({ print })
    .stderr({ print })
    .timeout(20000)
    .command(['environments:destroy', '-a', mock_account.name, failing_mock_env.name, '--auto-approve', '--skip-deregistered'])
    .it('should warn and exit with non-error status when skip-deregistered is provided', ctx => {
      expect(ctx.stderr).contains(`Warning: No configured environments found matching ${failing_mock_env.name}.`);
    });

  mockArchitectAuth()
    .stub(PipelineUtils, 'pollPipeline', async () => null)
    .nock(MOCK_API_HOST, api => api
      .get(`/accounts/${mock_account.name}`)
      .reply(200, mock_account))
    .nock(MOCK_API_HOST, api => api
      .get(`/accounts/${mock_account.id}/environments/${failing_mock_env.name}`)
      .reply(200, failing_mock_env))
    .stdout({ print })
    .stderr({ print })
    .timeout(20000)
    .command(['environments:destroy', '-a', mock_account.name, failing_mock_env.name, '--auto-approve', '--skip-deregistered=true'])
    .it('should warn and exit with non-error status when skip-deregistered is set explicitly to true', ctx => {
      expect(ctx.stderr).to.contain(`Warning: No configured environments found matching ${failing_mock_env.name}.`);
    });

  mockArchitectAuth()
    .stub(PipelineUtils, 'pollPipeline', async () => null)
    .nock(MOCK_API_HOST, api => api
      .get(`/accounts/${mock_account.name}`)
      .reply(200, mock_account))
    .nock(MOCK_API_HOST, api => api
      .get(`/accounts/${mock_account.id}/environments/${failing_mock_env.name}`)
      .reply(404))
    .stdout({ print })
    .stderr({ print })
    .timeout(20000)
    .command(['environments:destroy', '-a', mock_account.name, failing_mock_env.name, '--auto-approve', '--skip-deregistered=false'])
    .catch(e => {
      expect(e.message).to.contain('Request failed with status code 404');
    })
    .it('should exit with error status when skip-deregistered is set explicitly to false');
```

### DIFF
--- a/src/architect/environment/environment.utils.ts
+++ b/src/architect/environment/environment.utils.ts
@@ -25,14 +25,21 @@ export class EnvironmentUtils {
     }),
   };
 
-  static async getEnvironment(api: AxiosInstance, account: Account, environment_name?: string): Promise<Environment> {
+  static async getEnvironment(api: AxiosInstance, account: Account, environment_name?: string, skip_deregistered = false): Promise<Environment> {
     if (process.env.ARCHITECT_ENVIRONMENT === environment_name && process.env.ARCHITECT_ENVIRONMENT) {
       console.log(chalk.blue(`Using environment from environment variables: `) + environment_name);
     }
 
     let environment: Environment;
     if (environment_name) {
-      environment = (await api.get(`/accounts/${account.id}/environments/${environment_name}`)).data;
+      let _environment_not_found = false;
+      const response = await api.get(`/accounts/${account.id}/environments/${environment_name}`, {
+        validateStatus: function (status): boolean {
+          _environment_not_found = status === 404;
+          return status === 200 || (_environment_not_found && skip_deregistered);
+        },
+      });
+      environment = await response?.data;
     } else {
       // eslint-disable-next-line unicorn/prefer-module
       inquirer.registerPrompt('autocomplete', require('inquirer-autocomplete-prompt'));
--- a/src/commands/environments/destroy.ts
+++ b/src/commands/environments/destroy.ts
@@ -30,6 +30,11 @@ export default class EnvironmentDestroy extends BaseCommand {
       default: false,
       sensitive: false,
     }),
+    'skip-deregistered': booleanString({
+      description: 'If set to true, does not throw error when an environment cannot be found to deregister',
+      default: false,
+      sensitive: false,
+    }),
     force: booleanString({
       description: 'Force the deletion even if the environment is not empty',
       char: 'f',
@@ -62,7 +67,12 @@ export default class EnvironmentDestroy extends BaseCommand {
     const { args, flags } = await this.parse(EnvironmentDestroy);
 
     const account = await AccountUtils.getAccount(this.app, flags.account);
-    const environment = await EnvironmentUtils.getEnvironment(this.app.api, account, args.environment);
+    const environment = await EnvironmentUtils.getEnvironment(this.app.api, account, args.environment, flags['skip-deregistered']);
+
+    if (!environment.id && flags['skip-deregistered']) {
+      this.warn(`No configured environments found matching ${args?.environment}.`);
+      return;
+    }
 
     let answers = await inquirer.prompt([{
       type: 'input',
--- a/src/commands/environments/destroy.ts
+++ b/src/commands/environments/destroy.ts
@@ -31,8 +31,8 @@ export default class EnvironmentDestroy extends BaseCommand {
       sensitive: false,
     }),
     'skip-deregistered': booleanString({
-      description: 'If set to true, does not throw error when an environment cannot be found to deregister',
-      default: false,
+      description: 'If set to false, throws an error when an environment cannot be found to deregister',
+      default: true,
       sensitive: false,
     }),
     force: booleanString({
--- a/test/commands/environments/destroy.test.ts
+++ b/test/commands/environments/destroy.test.ts
@@ -21,6 +21,60 @@ describe('environment:destroy', () => {
     id: 'test-pipeline-id'
   }
 
+  const failing_mock_env = {
+    id: null,
+    name: 'failing-test-env'
+  }
+
+  mockArchitectAuth()
+    .stub(PipelineUtils, 'pollPipeline', async () => null)
+    .nock(MOCK_API_HOST, api => api
+      .get(`/accounts/${mock_account.name}`)
+      .reply(200, mock_account))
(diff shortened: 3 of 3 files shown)


src/commands/environments/create.ts
```diff
@@ -28,6 +29,11 @@ export default class EnvironmentCreate extends BaseCommand {
       description: 'Environment Description',
       sensitive: false,
     }),
+    strict: booleanString({
```
ryan-cahill, line 32: This was going to be a hidden flag for now, right?

src/commands/environments/destroy.ts
```diff
@@ -30,6 +30,11 @@ export default class EnvironmentDestroy extends BaseCommand {
       default: false,
       sensitive: false,
     }),
+    strict: booleanString({
```
ryan-cahill, line 33: If I remember correctly, this was also going to be hidden for now

src/architect/environment/environment.utils.ts
```diff
@@ -25,14 +25,20 @@ export class EnvironmentUtils {
     }),
   };
 
-  static async getEnvironment(api: AxiosInstance, account: Account, environment_name?: string): Promise<Environment> {
+  static async getEnvironment(api: AxiosInstance, account: Account, environment_name?: string, strict = true): Promise<Environment> {
```
ryan-cahill, line 28: Instead of adding another arg to the function, can you please create an options type to pass in that would look something like the following:

```ts
{
  environment_name?: string;
  strict?: boolean;
}
```

Ideally we can do that to keep the number of function args low.

test/commands/environments/destroy.test.ts
```diff
@@ -3,23 +3,93 @@ import PipelineUtils from '../../../src/architect/pipeline/pipeline.utils';
...
 describe('environment:destroy', () => {
-
   // set to true while working on tests for easier debugging; otherwise oclif/test eats the stdout/stderr
   const print = false;
 
   const mock_account = {
     id: 'test-account-id',
-    name: 'test-account'
-  }
+    name: 'test-account',
```
ryan-cahill, line 11: Could you add tests for a strict `environment:create` command as well please? I see that feature is already merged, but I don't couldn't find any tests for it

ryan-cahill (changes requested):

src/architect/environment/environment.utils.ts
devinsag: added new interface, and updated all calls to this function to use the new signature.

```typescript
export interface GetEnvironmentOptions {
  environment_name?: string;
  strict?: boolean;
}
```

src/commands/environments/create.ts
devinsag: Set flag to hidden

src/commands/environments/destroy.ts
devinsag: set flag to hidden

README.md
```diff
@@ -62,7 +62,7 @@ USAGE
 * [`architect dev:stop [NAME]`](#architect-devstop-name)
 * [`architect doctor`](#architect-doctor)
 * [`architect environments:create [ENVIRONMENT]`](#architect-environmentscreate-environment)
-* [`architect environments:destroy [ENVIRONMENT]`](#architect-environmentsdestroy-environment)
+* [`architect environment:destroy [ENVIRONMENT]`](#architect-environmentsdestroy-environment)
```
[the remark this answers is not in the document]
devinsag: pulled README.md from origin/rc

src/architect/secret/secret.utils.ts
```diff
@@ -32,7 +33,8 @@ export default class SecretUtils {
 
   static async batchUpdateSecrets(app: AppService, secrets: Secret[], account: Account, options?: SecretOptions): Promise<void> {
     if (options?.environment_name) {
-      const environment = await EnvironmentUtils.getEnvironment(app.api, account, options?.environment_name);
+      const getEnvironmentOptions: GetEnvironmentOptions = { environment_name: options.environment_name };
```
ryan-cahill, line 36: Thanks for adding this options object. Can you make the variable namec `getEnvironmentOptions` snake case where it's used please? To be consistent

ryan-cahill (changes requested):

src/architect/secret/secret.utils.ts
devinsag: updated references to use snake case;

test/commands/environments/destroy.test.ts
devinsag: Added environment:create tests as well

src/commands/environments/destroy.ts
```diff
@@ -68,7 +68,7 @@ export default class EnvironmentDestroy extends BaseCommand {
     const { args, flags } = await this.parse(EnvironmentDestroy);
 
     const account = await AccountUtils.getAccount(this.app, flags.account);
-    const get_environment_options: GetEnvironmentOptions = { environment_name: flags.environment, strict: flags.strict };
+    const get_environment_options: GetEnvironmentOptions = { environment_name: args.environment, strict: flags.strict };
```
devinsag, line 71: Thank you for catching this. This change is necessary because `destroy.ts` does not extend the `EnvironmentUtils.flags`, so `environment` would have to be an arg instead of a flag.

ryan-cahill (approved):

PR #777, opened 2022-12-13
Fixing register for services with reserved names

ryan-cahill:
## Overview
Fixes registering a component with a service that includes a reserved name (https://gitlab.com/architect-io/architect-cli/-/issues/579)

## Changes
* Passing `architect_ref` to `getImage` rather than just `ref`

## Tests
* Added test that registers the superset

### DIFF
--- a/src/common/docker-compose/index.ts
+++ b/src/common/docker-compose/index.ts
@@ -331,8 +331,7 @@ export class DockerComposeUtils {
       if (service.build) {
         if (!service.image) {
           // eslint-disable-next-line unicorn/consistent-destructuring
-          const image = options.getImage ? options.getImage(node.config.metadata.ref) : node.ref;
-          service.image = image;
+          service.image = options.getImage ? options.getImage(node.config.metadata.architect_ref) : node.ref;
         }
 
         // Optimization to check if multiple services share the same dockerfile/build config and avoid building unnecessarily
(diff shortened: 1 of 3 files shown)


src/common/docker-compose/index.ts
```diff
@@ -331,8 +331,7 @@ export class DockerComposeUtils {
       if (service.build) {
         if (!service.image) {
           // eslint-disable-next-line unicorn/consistent-destructuring
-          const image = options.getImage ? options.getImage(node.config.metadata.ref) : node.ref;
-          service.image = image;
+          service.image = options.getImage ? options.getImage(node.config.metadata.architect_ref) : node.ref;
```
tjhiggins, line 334: To clarify does the `config.metadata.ref` include the reserved name?

I probably added `architect_ref`, but was immediately confused by this logic.
I'll merge this, but might be worth adding a tech debt ticket since there are 3 refs right now.
`node.ref` // Computed ref - reserved_name
`node.config.metadata.ref` // Feel like this should be the internal ref and remove architect_ref
`node.config.metadata.architect_ref`

ryan-cahill: Registering the service `stateful-api` with the reserved name `stateful-reserved-name`:

* `node.ref` = `stateful-reserved-name`
* `node.config.metadata.ref` = `stateful-reserved-name`
* `node.config.metadata.architect_ref` = `superset.services.stateful-api`

Where by comparison, `stateless-app` that doesn't have a reserved name:

* `node.ref` = `superset--stateless-app`
* `node.config.metadata.ref` = `superset.services.stateless-app`
* `node.config.metadata.architect_ref` = `superset.services.stateless-app`

ryan-cahill: Agreed that that's confusing and I'll make a ticket

PR #787, opened 2022-12-21
Service-level secrets

ryan-cahill:
## Overview
Expands the environment block of services and tasks to let the user define expected secrets there, rather than at the top level. See https://gitlab.com/architect-io/architect-cli/-/issues/570

## Changes
* Allowed users to pass secrets in to and define them at the `service.environment` or `task.environment` level
* Added validation for service and task-level secrets
* Added tests to validate new behavior for both services and tasks

## Tests
* Added a new test file, `test/dependency-manager/resource-environment-interpolation.test.ts`

## Docs
* To be completed with https://gitlab.com/architect-io/product/-/issues/228


ryan-cahill:
@tjhiggins should task environments be updated in the same way as service environments? If so, I have another ticket for that, but can also do that here

src/dependency-manager/manager.ts
```diff
@@ -361,16 +361,91 @@ export default abstract class DependencyManager {
...
+  async getGraph(original_component_specs: ComponentSpec[], all_secrets: SecretsDict = {}, options?: GraphOptions): Promise<DependencyGraph> {
+    options = {
       interpolate: true,
       validate: true,
       ...options,
     };
 
     if (options.validate) {
       SecretsConfig.validate(all_secrets);
+      this.validateEnvironments(original_component_specs, all_secrets);
```
tjhiggins, line 448: Making a note to myself that it'd prob be easier to validate the ComponentConfig than the ComponentSpec. Will come back to this.

tjhiggins:
Closing this pr since we are still undecided on the correct UX. Created a pr off of this one that includes all of the cleanup without func changes:
https://github.com/architect-team/architect-cli/pull/811

PR #789, opened 2022-12-22
add nullsafe chain, and filter out potential null

devinsag:
## Overview
add null-safe chain to make referencing user input via either invalid file parsing, or cli stdio directly more safe.
filter out potentially undefined values coming from stdio or file parsing directly.

please see: https://gitlab.com/architect-io/architect-cli/-/issues/595

## Tests
architect dev using an empty architect.yml configuration file

## Docs
N/A

## Pictures
N/A

### DIFF
--- a/src/sentry.ts
+++ b/src/sentry.ts
@@ -104,7 +104,7 @@ export default class SentryService {
       };
 
       for (const flag of non_default_flags) {
-        sentry_tags[`flag.${flag}`] = command_class.flags?.[flag]?.sensitive ? 'Filtered' : flags?.[flag];
+        sentry_tags[`flag.${flag}`] = command_class.flags[flag].sensitive ? 'Filtered' : flags[flag];
       }
 
       if (sentry_user?.email) {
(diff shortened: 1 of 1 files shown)


src/sentry.ts
```diff
@@ -103,7 +104,7 @@ export default class SentryService {
       };
 
       for (const flag of non_default_flags) {
-        sentry_tags[`flag.${flag}`] = command_class.flags[flag].sensitive ? 'Filtered' : flags[flag];
+        sentry_tags[`flag.${flag}`] = command_class.flags?.[flag]?.sensitive ? 'Filtered' : flags?.[flag];
```
TylerAldrich, line 107: Is this change necessary with the `present_flags_keys` filtering above? It seems like that's guaranteeing that `command_class.flags[flag]` will always exist. Also can `flags` ever be undefined? `flags?[flag]` seems like it shouldn't be necessary, I wouldn't expect parse to return undefined flags

devinsag: Tested a blank `architect.yml` without optional chaining, and it runs OK with the filtering method above using `present_flags_keys`. Updated to remove the chaining

PR #792, opened 2022-12-30
automatically cleaning up cached data that's older than 7 days

ryan-cahill:
(Edited by Tyler)
- Cleans up cached build data that's older than 7 days
- Prunes architect-created images that are dangling (Doesn't always free up space depending on if the same layers are used, but can help)
- Prunes architect-created images for a component if the service name has changed (Same as above, won't always free up space if all layers are the same but does keep the image list clean)


src/commands/dev/index.ts
```diff
@@ -409,6 +412,52 @@ export default class Dev extends BaseCommand {
...
+   */
+  async pruneImages(compose: DockerComposeTemplate): Promise<void> {
+    // Remove build cache layers with our label that are older than 7 days
+    await docker(['builder', 'prune', '--filter', `label=${DOCKER_IMAGE_LABEL}`, '--filter', `unused-for=${7 * 24}h`, '-af'], { stdout: false });
+    // Remove images that are untagged and created by previous `architect dev` runs
+    await docker(['image', 'prune', '--filter', `label=${DOCKER_IMAGE_LABEL}`, `--filter`, 'dangling=true', '-f'], { stdout: false });
+
+    // Find the set of services and name of all components being deployed. Any image that is labelled with the name
+    // of a component being run but uses a Repository name that no longer matches any service names can be
+    // removed. This can happen when changing the name of a service in the architect.yml.
```
ryan-cahill, line 427: Is this a fair assumption? Someone might change their mind and rename their component to something they used five minutes ago. I don't love not putting some kind of time restriction on these operations

src/commands/dev/index.ts
```diff
@@ -409,6 +412,52 @@ export default class Dev extends BaseCommand {
...
+        if (label.startsWith('component=')) {
+          components.push(label.split('=')[1]);
+        }
+      }
+    }
+
+    for (const component_name of components) {
+      const built_image_result = await docker(['image', 'list',
+        '--filter', `label=${DOCKER_IMAGE_LABEL}`,
+        '--filter', `label=component=${component_name}`,
```
ryan-cahill, line 442: Is there a way to do something like `'--filter', `label=component=component_name_1|component=component_name_2|component=component_name_3` to reduce the amount of Docker CLI calls we're waiting on?

src/common/docker-compose/index.ts
```diff
@@ -351,6 +351,17 @@ export class DockerComposeUtils {
...
 
+      if (service.build) {
+        // Add labels to enrich image filtering
+        if (!service.build.labels) {
+          service.build.labels = [];
+        }
+        service.build.labels.push(DOCKER_IMAGE_LABEL);
+
+        const { component_name } = ResourceSlugUtils.parse(node.config.metadata.ref);
+        service.build.labels.push(`component=${component_name}`);
```
ryan-cahill, line 362: Is the label `component` specific enough? People can add their own labels and this could easily conflict

src/commands/dev/index.ts
TylerAldrich: Hmm that's fair, I think in most cases of quick name waffling the docker layer cache will do the heavy lifting and the user wouldn't even know we cleared the previously used image name (e.g. if you just change image name and none of the layers, nothing really gets rebuilt except metadata about the image). Interested in more opinions here (if you have a sec @mueschm) 

One of the downsides to time gated cache clearing is if you're doing something that's causing old images to constantly get created and become danging (like, changing the name of a service + making changes that trigger large rebuilds) you'll rapidly fill your disk with garbage and waiting 7d to clean it would leave users with issues still.

src/commands/dev/index.ts
TylerAldrich: good q I'll look into this, not sure if filters behave as OR/AND and that can maybe prevent the docker call in a loop 👍

src/common/docker-compose/index.ts
TylerAldrich: fair point, can prefix with `architect` (not sure if `.` is allowed but maybe `architect.component`) just to namespace it/free up more general words for people

src/commands/dev/index.ts
TylerAldrich: Did some poking around and it seems like these filters are always ANDed together so there's not a way to combine commands. That being said, this situation only comes up if you `architect dev component_1.yml component_2.yml ...` which is fairly uncommon, most cases involve running a single component so this loop has an N = 1

src/commands/dev/index.ts
```diff
@@ -531,6 +578,8 @@ export default class Dev extends BaseCommand {
     this.log('Building containers...', chalk.green('done'));
     this.log('');
 
+    await this.pruneImages(compose);
```
ryan-cahill, line 581: Why does this run before the user's app starts and not after the user stops the compose process?

TylerAldrich: Technically it doesn't make a difference, I just figured it would be more beneficial to clean up old images after building and before running.

If you run `architect dev` and looked and `docker images` you'd see garbage until you kill `architect dev`, which is kind of weird - doing it in between building/running means even with `architect dev` running in the background, your `docker images` list is clean. Ultimately the things we'd delete doesn't change before/after containers get run since it's all inactive/dangling image based

TylerAldrich: (Also in the case of running `architect dev` detached, we'd want to still do this cleanup presumably and in that case would want to do it before launching containers and detaching)

ryan-cahill: I was less concerned about the list of images than about the UX when images are being deleted. If someone had a bunch of images that needed to be deleted and that took like 20 seconds, it's way more disruptive when they're waiting for their app to start than when they kill the process

TylerAldrich: That's fair, though I don't think it's that common of a scenario - the flip side is after killing their `architect dev` they have to wait 20 seconds to delete images which is also potentially annoying. And in general if somebody only runs `architect dev -d` they'd either never have things cleaned up (unless we deleted before running dev, like this is doing)

tjhiggins:
I tried running architect dev on the django starter project on this branch. It got stuck after building containers:
image

PR #794, opened 2023-01-04
582 602 expected docker compose err

devinsag:
## Overview
Please see:

https://gitlab.com/architect-io/architect-cli/-/issues/582

https://gitlab.com/architect-io/architect-cli/-/issues/602

## Changes

Add a check to see if yaml is empty before parsing. 
* Throws error if empty.

Add try catch around dev buildImage function. 
* Catch identified error message and display message to screen.

## Tests
run dev command against empty architect.yml file. See error message, and no sentry submission
run dev command with a dockerfile containing an exception. See error message, and no sentry submission.

```
test
    .timeout(20000)
    .stub(ComponentBuilder, 'loadFile', () => {
      return '';
    })
    .stub(Dev.prototype, 'failIfEnvironmentExists', sinon.stub().returns(undefined))
    .stub(Dev.prototype, 'runCompose', sinon.stub().returns(undefined))
    .stub(Dev.prototype, 'downloadSSLCerts', sinon.stub().returns(undefined))
    .stub(Dev.prototype, 'readSSLCert', sinon.stub().returns('fake-cert'))
    .stdout({ print })
    .stderr({ print })
    .command(['dev', './examples/hello-world/architect.yml'])
    .catch(err => {
      expect(err.message).to.include('For help getting started take a look at our documentation here: https://docs.architect.io/reference/architect-yml');
    })
    .it('Provide error if architect.yml is empty');
```

### Docker compose error
image

## Empty architect.yml
image

### DIFF
--- a/src/commands/dev/index.ts
+++ b/src/commands/dev/index.ts
@@ -506,9 +506,10 @@ export default class Dev extends BaseCommand {
     try {
       await DockerComposeUtils.dockerCompose(['-f', compose_file, '-p', project_name, 'build', ...build_args], { stdio: 'inherit' });
     } catch (e: any) {
-      if (e.failed && !e.signal && !e.isCancelled && !e.killed) {
-        // eslint-disable-next-line no-process-exit
-        process.exit(e.exitCode || 1);
+      const pattern = new RegExp('Command failed with exit code 17: docker compose -f [-p architect build]?', 'g');
+      const error = e?.message || '';
+      if (pattern.test(error)) {
+        throw new ArchitectError(`Docker compose has encountered an error building the image specified in ${compose_file}`, false);
       }
     }
     return [project_name, compose_file];
--- a/src/commands/dev/index.ts
+++ b/src/commands/dev/index.ts
@@ -503,12 +503,17 @@ export default class Dev extends BaseCommand {
       build_args.push('--build-arg', arg);
     }
 
+    const cmd = DockerComposeUtils.dockerCompose(['-f', compose_file, '-p', project_name, 'build', ...build_args], { stdin: 'inherit', stdout: 'inherit', stderr: 'pipe' });
+
+    let stderr_message = '';
     try {
-      await DockerComposeUtils.dockerCompose(['-f', compose_file, '-p', project_name, 'build', ...build_args], { stdio: 'inherit' });
+      cmd.stderr?.on('data', (bytes) => {
+        stderr_message = (bytes.toString() || '').trim();
+      });
+      await cmd;
     } catch (e: any) {
-      const pattern = new RegExp('Command failed with exit code 17: docker compose -f [-p architect build]?', 'g');
-      const error = e?.message || '';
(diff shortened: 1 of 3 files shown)


src/commands/dev/index.ts
```diff
@@ -503,7 +503,13 @@ export default class Dev extends BaseCommand {
       build_args.push('--build-arg', arg);
     }
 
-    await DockerComposeUtils.dockerCompose(['-f', compose_file, '-p', project_name, 'build', ...build_args], { stdio: 'inherit' });
+    try {
+      await DockerComposeUtils.dockerCompose(['-f', compose_file, '-p', project_name, 'build', ...build_args], { stdio: 'inherit' });
+    } catch (e: any) {
+      const pattern = new RegExp('Command failed with exit code 17: docker compose -f [-p architect build]?', 'g');
```
TylerAldrich, line 509: This pattern isn't sufficient, if I run `architect dev -e foo` the failure is 
```
Command failed with exit code 17: docker compose -f /Users/tyler/.config/architect/local/architect/architect/docker-compose/foo.yml -p foo build
```

the `-p` argument is equivalent to the environment we pass in.

We should be able to get the exit code from `e.exitCode` and check that value instead. Is exit code 17 the only possibility / the only exit code we care about? I imagine any non-zero exit code from `docker compose build` wouldn't be something we can fix on our end, but not 100% sure of that

TylerAldrich (changes requested):

src/commands/dev/index.ts
devinsag: Updated logic to fetch docker compose exit code `e.exitCode` instead of parsing error message.  Updated to throw error when non-zero exit code.

src/commands/dev/index.ts
```diff
@@ -503,7 +503,20 @@ export default class Dev extends BaseCommand {
...
     }
 
-    await DockerComposeUtils.dockerCompose(['-f', compose_file, '-p', project_name, 'build', ...build_args], { stdio: 'inherit' });
+    const cmd = DockerComposeUtils.dockerCompose(['-f', compose_file, '-p', project_name, 'build', ...build_args], { stdin: 'inherit', stdout: 'inherit', stderr: 'pipe' });
+
+    let stderr_message = '';
+    try {
+      cmd.stderr?.on('data', (bytes) => {
+        stderr_message = (bytes.toString() || '').trim();
+      });
```
TylerAldrich, line 512: I'm pretty sure this information is in the exception object returned by execa - I believe `e.stderr`is available so we don't need to set the string this way. You can run `console.log(e)` to quickly see the various keys available - but I think that would be the same information that's being captured here

TylerAldrich (changes requested):

src/commands/dev/index.ts
devinsag: removed  on-data function

TylerAldrich (approved):

PR #798, opened 2023-01-13
418 remove examples

devinsag:
## Overview
Remove examples directory from the CLI repo and move projects there to the architect-community org

please see https://gitlab.com/architect-io/architect-cli/-/issues/418

## Changes
* removed examples folder
* moved examples `architect.yml` files into test/mocks/examples folder.
* renamed example `architect.yml` files to `{project-name}.architect.yml`
* wrote new method to retrieve config for project based on project name

## Tests
`npm run test`

## Docs
N/A

## Pictures
N/A


test/mocks/examples/gcp-pubsub-publisher.architect.yml
```diff
@@ -16,7 +16,7 @@ dependencies:
 services:
```
mueschm, line 16: The only two files that are used here anymore are `hello-world` and `database-seeding`. None of the others need to be kept around.

.github/workflows/test-deploy.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - run: ./bin/dev config:set log_level debug
       - run: ./bin/dev login -e ${{ secrets.ARCHITECT_EMAIL }} -p ${{ secrets.ARCHITECT_PASSWORD }}
       - run: docker volume create volume-key
-      - run: ./bin/dev link examples/hello-world/architect.yml
+      - run: ./bin/dev link test/mocks/examples/hello-world.architect.yml
```
mueschm, line 29: From reading everything, I believe the only reason the `context` folder in examples really needs to exist is because of this right here.

I do like having everything needed to test in one repo, but I almost wonder if we should split this out.

I see 2 potential solutions to clean this all up.
1. We make a new folder in tests for integration testing that contains the original `hello-world` folder.
2. We git clone the repository from somewhere else when needed.

I think number 1 is the cleaner solution

mueschm (changes requested):

test/mocks/examples/gcp-pubsub-publisher.architect.yml
devinsag: removed extra projects that were not being utilized in tests.

.github/workflows/test-deploy.yml
devinsag: Created a new folder `tests/integration`, and moved the utilized projects within.

mueschm: This file looks like it needs to be updated. Shouldn't it point to `tests/integration`? 

Also not every file needed to be moved to integration. Out unit test files should just be in tests/mock and our integration test should be in tests/integration. Also the folder should be renamed from context back to hello-world. It should also contain its architect.yml file. Does that make sense?

devinsag: Did not have a chance to update CI job path until now. React-app and database-seeding are used frequently in `test/commands/register.ts` test. should the mocks for those projects in `test/commands/dev/index.ts` be exported and used for the other tests as well?

mueschm: Anything in tests/commands should never actually need the context to be real though. So we just need the `architect.yml` files to make sure the flow works. The only time we need the context is if we actually call docker which should not happen in a unit test.
Any files that are used for unit tests should be moved under tests/mock without a context folder. Any files that are used for integration tests should be moved to tests/integration and require a src directory since we actually need to run those.

devinsag: Moved the three architect.yml files into the mocks/examples folder.  Copied the hello-world project into the mocks/integration/hello-world folder.

PR #799, opened 2023-01-18
Add timeout to getToken

tjhiggins:
Add expiration window to combat race condition
Upgrade simple-oauth2

### DIFF
--- a/src/app-config/auth.ts
+++ b/src/app-config/auth.ts
@@ -26,6 +26,7 @@ export default class AuthClient {
   checkLogin: () => Promise<User>;
 
   // Provide a window of time before the actual expiration to refresh the token
+  // https://github.com/lelylan/simple-oauth2/tree/1d0b2788ab178ca244c26859c31a34a32d8de979#refresh-an-access-token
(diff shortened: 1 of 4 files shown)


src/app-config/auth.ts
```diff
@@ -130,6 +134,7 @@ export default class AuthClient {
         payload: { 'client_id': this.config.oauth_client_id },
         headers: { 'HOST': url.hostname },
         rejectUnauthorized: !url.hostname.endsWith('.localhost'),
+        timeout: 30000,
```
TylerAldrich, line 137: is this intentionally disconnected from `EXPIRATION_WINDOW_IN_SECONDS` or should this be `EXPIRATION_WINDOW_IN_SECONDS * 100`?

tjhiggins: Has no relation. This is just an arbitrary timeout for an http call.

TylerAldrich: coo 👍

TylerAldrich (approved):

PR #809, opened 2023-01-25
Check kubernetes version in register

mtgoncalves1:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/610   
Check the version of kubernetes cluster to make sure it meets a minimum version before register a cluster.  

## Changes
Add a check for kubernetes version in the createCluster function.

## Tests
Change `clientVersion.gitVersion` in `KubeClusterUtils.getClientVersion()` before the return statement. For examples:  
Fail version   
```
clientVersion.gitVersion = 'v0.22.0';
```
Pass version  
```
clientVersion.gitVersion = 'v2.0.0';
```

## Pictures
`$ architect clusters:create`   
Screenshot 2023-01-26 at 4 21 43 PM


mtgoncalves1:
> This check needs to occur when they create a cluster and not when they register a component. Also your example error message shows that version `1.25.1` failed which is above `1.21` so it should work.

I have updated the check to clusterCreate function instead of the register command. 
The error message was inaccurate because the version is read from the kubernetes `gitVersion`, not the major + minor, and during testing I didn't use the gitVersion.

src/architect/cluster/cluster.utils.ts
```diff
@@ -81,4 +84,17 @@ export default class ClusterUtils {
...
+
+  private static async getClientVersion() {
+    const { stdout } = await execa('kubectl', ['version', '--client', '--output', 'json']);
+    const { clientVersion } = JSON.parse(stdout);
+    return clientVersion;
+  }
+
+  public static async checkClientVersion(): Promise<void> {
+    const clientVersion = await this.getClientVersion();
+    if (Number(clientVersion.major) < MIN_CLUSTER_VERSION.major || (Number(clientVersion.major) === MIN_CLUSTER_VERSION.major && Number(clientVersion.minor) < 22)) {
```
tjhiggins, line 96: Use [semver](https://www.npmjs.com/package/semver) pkg to check version. It's already installed in the cli.

tjhiggins (changes requested):

src/architect/cluster/cluster.utils.ts
```diff
@@ -81,4 +84,17 @@ export default class ClusterUtils {
...
+      throw new Error(`Currently, we only support Kubernetes clusters on version ${MIN_CLUSTER_VERSION.major}.${MIN_CLUSTER_VERSION.minor} or greater. Your cluster is currently on version ${clientVersion.gitVersion} which is below the minimum required version. Please upgrade your cluster before registering it with Architect.`);
```
tjhiggins, line 97: Throw an ArchitectError

tjhiggins (changes requested):

src/architect/cluster/cluster.utils.ts
mtgoncalves1: I have updated to use semver to do version comparison.

ryan-cahill (changes requested):

src/architect/cluster/cluster.utils.ts
```diff
@@ -21,6 +24,8 @@ export interface KubernetesClusterCredentials {
   service_token: string;
 }
 
+export const MIN_CLUSTER_VERSION = { 'major': 1, 'minor': 22, 'gitVersion': '1.22.0' };
```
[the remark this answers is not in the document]
mtgoncalves1: Good point. I have updated it to '1.22.0' instead of an object.

src/commands/clusters/create.ts
```diff
@@ -104,6 +104,8 @@ export default class ClusterCreate extends BaseCommand {
   }
 
   private async createCluster() {
+    await ClusterUtils.checkClientVersion();
```
mueschm, line 107: Did not catch this the first time. This will not work. This will use just use the default context and see if that passes, we need to wait for the user to select which context to use and then check. Also if they specify a different kubeconfig then this also does not work.

Instead, this needs to be moved down to after.
```
const kube_contexts = await this.setupKubeContext(flags);
```

src/commands/clusters/create.ts
```diff
@@ -134,6 +134,7 @@ export default class ClusterCreate extends BaseCommand {
     const account = await AccountUtils.getAccount(this.app, flags.account, { account_message: 'Select an account to register the cluster with' });
 
     const kube_contexts = await this.setupKubeContext(flags);
+    await ClusterUtils.checkClientVersion();
```
mueschm, line 137: This needs to take in `flags.kubeconfig` in order to use the correct file.

mueschm (approved):

src/architect/cluster/cluster.utils.ts
```diff
@@ -81,4 +87,23 @@ export default class ClusterUtils {
...
+    const { clientVersion } = JSON.parse(stdout);
+    return clientVersion.gitVersion;
+  }
+
+  public static async checkClientVersion(kubeconfig: string): Promise<void> {
+    const kubeconfig_path = untildify(kubeconfig);
+    const client_git_version = await this.getClientVersion(kubeconfig_path);
+    const client_semver = semver.coerce(client_git_version);
+    if (!client_semver) {
+      throw new ArchitectError(`Failed to translate Kubernetes cluster version ${client_git_version}.`);
```
ryan-cahill, line 102: @mueschm should this be a hard error? If we get this wrong, the user can't create a cluster when it's possible that they should be able to. Should this be a yellow warning instead?

mueschm: If we can't verify the version we cannot guarantee it works. Better for them to contact support the set up a bad system. 1.21 is still prominent and showcased in a lot of online tutorials so I would prefer to be safe.

ryan-cahill (approved):

PR #819, opened 2023-01-30
Only run linter for git on changed files

mueschm:
During the commit hooks for git, only lint changed files.


package.json
```diff
@@ -104,6 +103,8 @@
     "eslint": "^7.32.0",
     "eslint-config-oclif": "^4",
     "eslint-config-oclif-typescript": "^1.0.3",
+    "husky": "^8.0.3",
+    "is-ci": "^3.0.1",
```
tjhiggins, line 107: We use is-ci at runtime

tjhiggins (changes requested):

PR #820, opened 2023-02-02
Pass up cluster information to allow it to work locally

mueschm:
## Overview
Send up cluster information to allow us to know if it is a local cluster.

## Changes
1. Add a function that can get a server for a context.
2. Send up the server to the API.

## Tests
Tested alongside https://gitlab.com/architect-io/hub-api/-/merge_requests/700

### DIFF
--- a/src/architect/cluster/cluster.utils.ts
+++ b/src/architect/cluster/cluster.utils.ts
@@ -14,11 +14,11 @@ export interface CreateClusterInput {
 export type ClusterCredentials = KubernetesClusterCredentials;
 
 export interface KubernetesClusterCredentials {
-  kind: 'KUBERNETES';
+  kind: 'KUBERNETES' | 'AGENT';
 
   host: string;
-  cluster_ca_cert: string;
-  service_token: string;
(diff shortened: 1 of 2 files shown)


src/architect/cluster/cluster.utils.ts
```diff
@@ -14,11 +14,11 @@ export interface CreateClusterInput {
 export type ClusterCredentials = KubernetesClusterCredentials;
 
 export interface KubernetesClusterCredentials {
-  kind: 'KUBERNETES';
+  kind: 'KUBERNETES' | 'AGENT';
 
   host: string;
-  cluster_ca_cert: string;
-  service_token: string;
+  cluster_ca_cert?: string;
```
ryan-cahill, line 20: Instead of making these optional, should there be a separate `AgentClusterCredentials`?

mueschm: Honestly I was debating removing them all together since they are no longer used, but figured this was okay for now.

ryan-cahill (approved):

PR #821, opened 2023-02-02
Lint staged

mueschm:
## Overview
Switch over to `lint-staged` which has better support for handling out git hooks.

## Tests
Ran local commits to make sure everything works.

### DIFF
--- a/lint-staged.config.js
+++ b/lint-staged.config.js
@@ -1,0 +1,19 @@
+const { ESLint } = require('eslint');
+
+const removeIgnoredFiles = async (files) => {
+  const eslint = new ESLint();
+  const ignoredFiles = await Promise.all(files.map((file) => eslint.isPathIgnored(file)));
+  const filteredFiles = files.filter((_, i) => !ignoredFiles[i]);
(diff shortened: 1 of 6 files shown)


lint-staged.config.js
```diff
@@ -0,0 +1,19 @@
...
+  const filteredFiles = files.filter((_, i) => !ignoredFiles[i]);
+  return filteredFiles.join(' ');
+};
+
+module.exports = {
+  '*.{js,ts}': async (files) => {
+    const filesToLint = await removeIgnoredFiles(files);
+    return [
+      `eslint --fix --max-warnings=0 ${filesToLint}`,
+      'npm run test',
```
tjhiggins, line 15: if you make these separate steps they will run concurrently
```
'*.{js,ts}': 'npm run test',
'*.{js,ts}': 'npm run check-circular',
...
```

mueschm: Because the keys clash it actually does not work, but I found a workaround that does work by using a fake extension like `__fake__` after the real extensions.

tjhiggins (approved):

PR #824, opened 2023-02-08
printing a warning that scaling settings will not be applied for a component run on a free account

ryan-cahill:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/615 by including messaging for the user

## Changes
* Added a warning when a user registers a component with a service that includes scaling settings to a free account

## Tests
* Added two tests to check that warning is shown or not shown where appropriate

## Pictures
image


src/commands/register.ts
```diff
@@ -331,6 +333,12 @@ export default class ComponentRegister extends BaseCommand {
       this.generateDependenciesWarnings(new_spec.dependencies, selected_account.name);
     }
 
+    for (const [service_name, service_config] of Object.entries(component_spec.services || {})) {
+      if (service_config.scaling && account_subscription.subscription_tier === 'Free') {
+        this.warn(chalk.yellow(`The service "${service_name}" contains scaling settings, but scaling is only available for the Team and Growth account tiers. The service will run normally, but scaling will not occur.`));
```
tjhiggins, line 338: Can we make it so the stack trace doesn't show for warning logs? See your screenshot.

tjhiggins (changes requested):

PR #829, opened 2023-02-13
updating ci release message to not skip ci

ryan-cahill:
## Overview
GitHub workflows don't run the CI workflows if the `chore(release)` includes `[skip ci]` and was the last commit of the PR - ex. https://github.com/architect-team/architect-cli/pull/828

## Changes
* Updated default chore commit message to only skip ci on branches that aren't `main`
  * Docs - https://github.com/semantic-release/git#options

### DIFF
--- a/release.config.js
+++ b/release.config.js
@@ -14,7 +14,6 @@ const git = [
       'architect-yml.md',
       'src/dependency-manager/schema/architect.schema.json',
     ],
-    'message': `chore(release): \${nextRelease.version} ${branch === 'main' ? '' : '[skip ci]'}\n\n\${nextRelease.notes}`,
   },
 ];
 const exec = [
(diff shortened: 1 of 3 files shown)


release.config.js
```diff
@@ -14,6 +14,7 @@ const git = [
       'architect-yml.md',
       'src/dependency-manager/schema/architect.schema.json',
     ],
+    'message': `chore(release): \${nextRelease.version} ${branch === 'main' ? '' : '[skip ci]'}\n\n\${nextRelease.notes}`,
```
tjhiggins, line 17: I don't think this will fix - since we want the gha to run before we merge into main

ryan-cahill: Oh yeah I see what you mean. That's weird then. We either need to omit those commits, which feels wrong, or update all of the `chore(release)` messages

ryan-cahill: Doesn't look like we can include a manual step here if `[skip ci]` is also present

https://github.com/architect-team/architect-cli/pull/830

ryan-cahill: Have any other ideas?

PR #833, opened 2023-02-17
Fix build tags

tjhiggins:


tjhiggins:
@mtgoncalves1 could you test buildpack for me on this branch?

Bryan @ Reprise hit a bug with a build optimization we support. I added a test case:
https://github.com/architect-team/architect-cli/pull/833/commits/4c6c0ddb4dc6f4540982c84dbcf68ac136541915

src/common/docker-compose/template.ts
```diff
@@ -7,7 +7,6 @@ interface XBakeConfig {
 
 export interface DockerServiceBuild {
   context?: string;
-  buildpack?: boolean;
```
tjhiggins, line 10: Reason I removed this is because its not part of the docker-compose spec. I wanted what we generate to be valid.

PR #834, opened 2023-02-17
Fixed analytics

davidthor:
## Overview

We had some issues attributing events to individual sessions making it hard to measure performance of our longer running commands.

## Changes

Added tracking event to capture command start time, and updated capture method to use anonymous id.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "opener": "^1.5.2",
     "p-limit": "^3.1.0",
     "patch-package": "^6.4.7",
-    "posthog-node": "^2.4.0",
+    "posthog-node": "^2.5.1",
     "readline": "^1.3.0",
     "reflect-metadata": "0.1.13",
     "semver": "^7.3.7",
--- a/src/app-config/auth.ts
+++ b/src/app-config/auth.ts
@@ -18,12 +18,22 @@ interface AuthResult {
   id_token?: string;
 }
 
+export type AuthClientOptions = {
+  checkLogin: () => Promise<User>;
+  on?: {
+    login?: (user: User) => void;
+  };
+};
+
 export default class AuthClient {
   config: AppConfig;
   credentials: CredentialManager;
   _auth_result?: AuthResult;
   callback_server: CallbackServer;
   checkLogin: () => Promise<User>;
+  on?: {
+    login?: (user: User) => void;
+  };
 
   // Provide a window of time before the actual expiration to refresh the token
   // https://github.com/lelylan/simple-oauth2/tree/1d0b2788ab178ca244c26859c31a34a32d8de979#refresh-an-access-token
@@ -31,11 +41,12 @@ export default class AuthClient {
 
   public static SCOPE = 'openid profile email offline_access';
 
-  constructor(config: AppConfig, checkLogin: () => Promise<User>) {
+  constructor(config: AppConfig, options: AuthClientOptions) {
     this.config = config;
     this.credentials = new CredentialManager(config);
     this.callback_server = new CallbackServer();
-    this.checkLogin = checkLogin;
+    this.checkLogin = options.checkLogin;
+    this.on = options.on;
   }
 
   async init(): Promise<void> {
@@ -151,6 +162,10 @@ export default class AuthClient {
     const email = decoded_token.email;
     await this.setToken(email, access_token.token as AuthResult);
     await this.dockerLogin(email);
+    if (this.on?.login) {
+      const user = await this.checkLogin();
+      await this.on.login(user);
+    }
   }
 
   async logout(): Promise<void> {
--- a/src/app-config/auth.ts
+++ b/src/app-config/auth.ts
@@ -1,6 +1,5 @@
 import { AccessToken, AuthorizationCode } from 'simple-oauth2';
 import { URL } from 'url';
-import User from '../architect/user/user.entity';
 import { docker } from '../common/docker/cmd';
 import LoginRequiredError from '../common/errors/login-required';
 import CallbackServer from './callback-server';
@@ -19,9 +18,8 @@ interface AuthResult {
 }
 
 export type AuthClientOptions = {
-  checkLogin: () => Promise<User>;
   on?: {
-    login?: (user: User) => void;
+    login?: () => Promise<void>;
   };
 };
 
@@ -30,9 +28,8 @@ export default class AuthClient {
   credentials: CredentialManager;
   _auth_result?: AuthResult;
   callback_server: CallbackServer;
-  checkLogin: () => Promise<User>;
   on?: {
-    login?: (user: User) => void;
+    login?: () => Promise<void>;
   };
 
   // Provide a window of time before the actual expiration to refresh the token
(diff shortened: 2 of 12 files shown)


src/app-config/posthog.ts
```diff
@@ -1,25 +1,56 @@
...
+    }
   }
 
-  capture(message: { event: string; properties?: Record<string | number, any>; }): void {
-    // We override reInit so the distinctId isn't used
-    return super.capture({ distinctId: '<unused>', ...message });
+  /**
+   * @override
+   */
+  getPersistedProperty(key: string): any | undefined {
```
tjhiggins, line 32: Hmm I'm super confused. We are extending the PostHog node class.
```
export class PostHog implements PostHogNodeV1 {
  private _sharedClient: PostHogClient
}
```

`getPersistedProperty` and `setPersistedProperty` are functions on the `_sharedClient`. So these would be net new methods not overrides unless I'm missing something.

src/app-config/posthog.ts
```diff
@@ -1,25 +1,56 @@
...
+    if (this._options.persistence === 'file') {
+      return this._properties[key];
+    } else {
+      return super.getPersistedProperty(key as any);
+    }
   }
 
-  // Override reInit to not set distinct_id every call
-  private reInit(distinctId: string): void {
```
tjhiggins, line 20: So confused how unused would ever be called since I overrided this method.

src/app-config/posthog.ts
```diff
@@ -1,25 +1,56 @@
...
+    this._properties[key] = value;
+
+    if (this._options.persistence === 'file') {
+      fs.ensureFileSync(this._options.propertiesFile);
+      fs.writeJSONSync(this._options.propertiesFile, this._properties);
+    }
+  }
+
+  capture(message: { event: string; properties?: Record<string | number, any>; }): void {
+    return super.capture({ distinctId: this.getPersistedProperty('anonymous_id'), ...message });
```
tjhiggins, line 54: Setting the distinctId to the anon will create a new user each time and won't allow for aliasing multiple times

src/app-config/posthog.ts
tjhiggins: Ah I see the new package changed it

package.json
```diff
@@ -47,7 +47,7 @@
     "opener": "^1.5.2",
     "p-limit": "^3.1.0",
     "patch-package": "^6.4.7",
-    "posthog-node": "^2.4.0",
+    "posthog-node": "^2.5.1",
```
tjhiggins, line 50: This pkg requires node 15+which is kind of weird. I guess we have the ci tests for node 14.

src/app-config/posthog.ts
tjhiggins: I think we might just need to override this method and use anon instead of distinct when necessary:
https://github.com/PostHog/posthog-js-lite/blob/cb050098714676b00ddbdae3a5a4e1bdf1feec11/posthog-core/src/index.ts#L145

tjhiggins: This would work for a single user's machine, but will create spam in recurring CI pipelines which is why I had to override in the first place.

davidthor: > This would work for a single user's machine, but will create spam in recurring CI pipelines which is why I had to override in the first place.

Totally forgot about this use case, but should we instead just force disable for CI? This flow works correctly for individual developer usage whereas the other one didn't.

package.json
davidthor: That's odd. The tests seem to have passed on node 14 and npm didn't alert me of the issue. How did you find out about the requirement?

tjhiggins: Was looking at the package-lock.json trying to figure out why my previous patch didn't work. Thought maybe it was using 2.5+ in the package-lock which would explain why the override wasn't working anymore.

src/app-config/posthog.ts
tjhiggins: I'd be down to force disable for CI. Prob not the analytics you're looking for. We have the `is-ci` pkg

davidthor: It's not an ideal answer, but we already collect server side events so I don't think the CI events are worth prioritizing over individual user experiences. I'll look into it now.

tjhiggins: Although my pref would still be to not set distinctId to an anon value. It should be easier now since we have direct access to override the captureStateless and other methods. 

```ts
  protected captureStateless(
    distinctId: string,
    event: string,
    properties?: { [key: string]: any },
    options?: PosthogCaptureOptions
  ): this {
    const distinct_id = this.getPersistedProperty('distinct_id');
    if (distinct_id) {
      const payload = this.buildPayload({ distinct_id: distinctId, event, properties })
      this.enqueue('capture', payload, options)
    } else {
      const payload = this.buildPayload({ anonymous_id: anonymous_id, event, properties })
      this.enqueue('capture', payload, options)
    }

    return this
  }
  ```

Comment (davidthor) [reply src/app-config/posthog.ts:54]: Using the anon-id is the exact way the browser-side versions of posthog work. The docs and marketing site don't have access to the user distinct ID, but the correlation still happens because we ran an "identify" in our web app. I think this is still the advertised way of doing it.

Comment (davidthor) [reply src/app-config/posthog.ts:56]: > I'd be down to force disable for CI. Prob not the analytics you're looking for. We have the is-ci pkg

Added now

Comment (davidthor) [reply src/app-config/posthog.ts:56]: I'm not sure why posthog-node doesn't inherit from this class, but this is what the web and react packages extend so there's easy `getDistinctId()` methods which falls back to the anon ID if there's nothing else there:

https://github.com/PostHog/posthog-js-lite/blob/master/posthog-core/src/index.ts#L606

Comment (tjhiggins) [reply src/app-config/posthog.ts:56]: The docs and marketing sites don't use distinct id at all right? So we shouldn't be using it here when its not the distinct id. What do you think of my proposal for captureStateless?

Comment (tjhiggins) [reply src/app-config/posthog.ts:56]: Yeah so we def want to use the anon id. It seems like an easy change and then we don't need to worry about disabling CI. Although up to you if you still want to disable. We could add a tag for isCI so we could filter out that usage.

Comment (davidthor) [reply src/app-config/posthog.ts:56]: @mueschm looping you in here per @tjhiggins request. He has to take off and said that you can take over reviewing and merging as you see fit.

We spoke offline as well and concluded that the implementation in this PR seems correct. See this docs article which took us a while to find that confirms how the aliasing works:
https://posthog.com/docs/integrate/user-properties#how-are-properties-managed-when-merging-or-aliasing-users

File: src/app-config/auth.ts [hunk 5]
```diff
@@ -18,24 +18,35 @@ interface AuthResult {
   id_token?: string;
 }
 
+export type AuthClientOptions = {
+  checkLogin: () => Promise<User>;
```
tjhiggins, line 22: So with the last minute realization on Friday. Do we need the checkLogin callbacks? Your initial logic which called alias in the checkLogin seemingly would have been sufficient.

davidthor: hmmm, there's definitely some room for re-working, but we can't just outright get rid of it because we use it to acquire the User object here:

https://github.com/architect-team/architect-cli/blob/bugfix/621/src/app-config/auth.ts#L166

Though there's something weirdly circular about the current implementation of the two callbacks now that you point it out.

davidthor: @tjhiggins updated this. Feel free to review and then let's get it merged in so we capture events correctly.

PR #855, opened 2023-03-14
Databases redacted secrets

mueschm:
## Overview
So we sometimes use redacted secrets and run validation. This allows redacted secrets to be valid for `parseUrl`. I am not 100% satisfied with this solution, but this can be reconsidered when we come back to configuration.

## Changes
1. Allow `*`'s as a valid input for parseUrl
2. Added a missing unit test

## Tests
Created `arc-databases-2` to test on the api and it all works.

### DIFF
--- a/src/dependency-manager/utils/parser.ts
+++ b/src/dependency-manager/utils/parser.ts
@@ -199,7 +199,8 @@ export class ArchitectParser {
           } else if (node.callee.value === 'parseUrl') {
             // Handle the edge case where connection_string is set to a secret but no value
             // is set on the secret or the secret is redacted we can just move forward
-            if (!node.arguments[0].value || node.arguments[0].value === '*******') {
+            const redacted_regex = /\**/g;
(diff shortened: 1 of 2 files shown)


src/dependency-manager/utils/parser.ts
```diff
@@ -198,14 +198,15 @@ export class ArchitectParser {
...
+            // is set on the secret or the secret is redacted we can just move forward
+            const redacted_regex = /\**/g;
+            if (!node.arguments[0].value || redacted_regex.test(node.arguments[0].value)) {
               value = '';
             } else {
               try {
                 value = (new URL(node.arguments[0].value) as any)[node.arguments[1].value as string];
               } catch {
-                throw new Error(`Unable to parse url ${value}.`);
+                throw new Error(`Unable to parse url ${node.arguments[0].valu}.`);
```
tjhiggins, line 209: typo

tjhiggins (changes requested):

PR #858, opened 2023-03-16
Add starter flag to init command

mueschm:
## Overview
Add `starter` flag to the init command to specify which starter project to use.
`architect init --starter Go`

## Changes
1. Added starter flag
2. Add basic test case

## Tests
Created projects using each method to make sure nothing broke.
Created project using current command.
`npm run test`

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -1,4 +1,5 @@
 import { Flags, Interfaces } from '@oclif/core';
+import { OutputFlags } from '@oclif/core/lib/interfaces';
 import chalk from 'chalk';
 import fs from 'fs';
 import inquirer from 'inquirer';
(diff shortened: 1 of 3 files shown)


src/commands/init.ts
```diff
@@ -75,12 +80,12 @@ export abstract class InitCommand extends BaseCommand {
 
   @RequiresGit()
   @RequiresDocker({ compose: true })
-  async runProjectCreation(project_name: string): Promise<void> {
+  async runProjectCreation(project_name: string, flags: any): Promise<void> {
```
tjhiggins, line 83:
```ts
flags: OutputFlags<typeof InitCommand.flags>
```

tjhiggins (changes requested):

PR #862, opened 2023-03-21
Use expanded depends_on syntax for service health

tjhiggins:
## Overview
If a service defines a liveness probe then we should wait till it is healthy in the depends_on.

## Pictures
image


TylerAldrich (approved):
Confirmed this works with this branch https://github.com/architect-templates/go/pull/3 and has the original behavior without a db liveness probe 🥳

PR #864, opened 2023-03-22
[DNM] - Update bin/run to hide ExperimentalWarnings

TylerAldrich:
Related to https://gitlab.com/architect-io/architect-cli/-/issues/640

Using node 18.0.0 until 18.1(ish,) Axios has some code that emits a warning, and that happens to run when running `architect init` but also any other architect command - something in oclif is running the code that ends up emitting this warning.

With node 18.0.0, this warning now goes away:
```sh
~/code/architect/architect-cli (arc-tyler) » ./bin/dev init                                                    
? What is the name of your project? (node:12303) ExperimentalWarning: The Fetch API is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
~/code/architect/architect-cli (arc-tyler) » ./bin/run init                                               
? What is the name of your project?
~/code/architect/architect-cli (arc-tyler) » node --version            
v18.0.0
```

Won't affect dev so we can still see these warnings, which seems potentially important

Also confirmed this works on windows via powershell:
```
PS F:\architect\architect-cli> .\bin\run.cmd init
? What is the filename of the Docker Compose file you would like to convert?
PS F:\architect\architect-cli> .\bin\dev.cmd init
? What is the name of your project? (node:5032) ExperimentalWarning: The Fetch API is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
PS F:\architect\architect-cli> ^C
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,26 @@
+# [1.36.0](https://github.com/architect-team/architect-cli/compare/v1.35.0...v1.36.0) (2023-03-21)
+
+
+### Bug Fixes
+
+* **databases:** Databases redacted secrets ([#855](https://github.com/architect-team/architect-cli/issues/855)) ([541e07e](https://github.com/architect-team/architect-cli/commit/541e07eed00abe9427342bf515eeefcdcf20f870))
+* **depends_on:** Use expanded depends_on syntax for service health ([#862](https://github.com/architect-team/architect-cli/issues/862)) ([9bc61ca](https://github.com/architect-team/architect-cli/commit/9bc61ca44e452f76d2cdca7ac1fe830b6f3b0434))
+* **init:** Add starter flag to init command ([#858](https://github.com/architect-team/architect-cli/issues/858)) ([8b9a564](https://github.com/architect-team/architect-cli/commit/8b9a564b248ff1b6ecd18c5b5f0ed2f15298f64e))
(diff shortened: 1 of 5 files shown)


tjhiggins:
> This https://github.com/architect-team/architect-cli/pull/864/files#diff-57b94356302c85712d33811d8bda2a57b5181e3b496466db149873f6ed118265L1 seems a little hacky but I couldn't find another way to actually hide these warnings when a user runs our package without requiring them to set an env variable themselves.
> 
> Curious what @mueschm / @tjhiggins thoughts are, from what I understand the `-S` flag in a shebang is fairly widely supported but technically non-standard and _some_ older systems may have issues with it, so I'm not 100% sure we actually want to fix this in this way

Do we know what is causing the warning to throw in the first place?

TylerAldrich:
> > This https://github.com/architect-team/architect-cli/pull/864/files#diff-57b94356302c85712d33811d8bda2a57b5181e3b496466db149873f6ed118265L1 seems a little hacky but I couldn't find another way to actually hide these warnings when a user runs our package without requiring them to set an env variable themselves.
> > Curious what @mueschm / @tjhiggins thoughts are, from what I understand the `-S` flag in a shebang is fairly widely supported but technically non-standard and _some_ older systems may have issues with it, so I'm not 100% sure we actually want to fix this in this way
> 
> Do we know what is causing the warning to throw in the first place?

Yeah - we make an Axios request to get a list of projects that are selectable from the dropdown. Though now that I think about it, we can probably do that request in a different way and dodge the warning...

tjhiggins:
Looks good - please open a non arc- branch for it

PR #871, opened 2023-03-28
Fix build args integer issue

mtgoncalves1:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/645  
Fix issue with build args value that occurred when converting a docker-compose.yml to architect.yml.

## Changes
If build args is a map, cast each value to string.

## Tests
docker-compose.yml  
```
version: "3"
services:
  app:
    build:
      context: .
      args:
        INT_ARG: 1
        BOOL_ARG: true
```

Run `architect-local init --from-compose=docker-compose.yml test`.   
Should generate the following architect.yml 
```
name: test
services:
  app:
    build:
      args:
        INT_ARG: '1'
        BOOL_ARG: 'true'
      context: .
    reserved_name: app
```

### DIFF
--- a/src/common/docker-compose/converter.ts
+++ b/src/common/docker-compose/converter.ts
@@ -115,6 +115,9 @@ export class ComposeConverter {
       build.args = {};
       if (Array.isArray(compose_build.args)) {
         for (const arg of compose_build.args) {
+          if (!arg) {
+ continue;
+}
           const [key, value] = arg.split('=');
           build.args[key] = value ? value.toString() : null;
         }
--- a/src/common/docker-compose/converter.ts
+++ b/src/common/docker-compose/converter.ts
@@ -116,15 +116,11 @@ export class ComposeConverter {
       if (Array.isArray(compose_build.args)) {
         for (const arg of compose_build.args) {
           const [key, value] = arg.split('=');
-          if (key && value) {
-            build.args[key] = value;
-          } else {
-            warnings.push(`Could not convert environment variable ${arg}`);
-          }
+          build.args[key] = value ? value.toString() : null;
         }
-      } else {
-        for (const [key, value] of Object.entries(compose_build.args as Dictionary<string>)) {
-          build.args[key] = value.toString();
+      } else if (compose_build.args instanceof Object && Object.keys(compose_build.args).length > 0) {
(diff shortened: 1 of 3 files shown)


src/common/docker-compose/converter.ts
```diff
@@ -123,7 +123,9 @@ export class ComposeConverter {
           }
         }
       } else {
-        build.args = compose_build.args;
+        for (const [key, value] of Object.entries(compose_build.args as Dictionary<string>)) {
+          build.args[key] = value.toString();
```
mueschm, line 127: Is the `toString` needed if the Dictionary is already casting everything to a string?

mueschm (approved):
Looks good and makes sense to me

src/common/docker-compose/converter.ts
```diff
@@ -123,7 +123,9 @@ export class ComposeConverter {
           }
         }
       } else {
```
ryan-cahill, line 125: Since this block is using specific logic and not just setting what's passed in to the build args anymore, we should do something like 

```ts
else if (compose_build.args instanceof Object && Object.keys(compose_build.args).length)
```

in order to make sure that the new code will actually be passed a useful dictionary

src/common/docker-compose/converter.ts
ryan-cahill: We shouldn't use `Dictionary<string>` here because what's actually being passed in is a `Dictionary<any>`. The `as` is just telling TS that everything is okay and it'll only see string values, but this isn't actually casting the values. This compose file 

```yml
version: 3
services:
  test-service:
    build:
      args:
        API_VERSION: $API_VERSION
        EMPTY:
```

will cause this to explode because of what I mentioned above. We'll need a little more logic here to handle all of the cases that a user could throw at us, for example:
```ts
build.args[key] = !!value ? value.toString() : value;
```

We should include a test for an empty arg here too

ryan-cahill (changes requested):

src/common/docker-compose/converter.ts
mtgoncalves1: Sounds good. I'll fix it.

ryan-cahill (approved):

PR #872, opened 2023-03-30
check if secrets are loaded correctly for a component

mtgoncalves1:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/632.  
Add a test to check if secrets are loaded correctly for a component when specifying tag and tenant.


test/commands/dev/index.test.ts
```diff
@@ -1170,6 +1170,32 @@ describe('local dev environment', function () {
         expect(runCompose.firstCall.args[0]).to.deep.equal(expected_instance_compose);
       });
 
+    local_dev
+      .stub(Dev.prototype, 'runCompose', sinon.stub().returns(undefined))
+      .stub(Dev.prototype, 'downloadSSLCerts', sinon.stub().returns(undefined))
+      // @ts-ignore
```
mueschm, line 1176: What are we ignoring here and why?

mtgoncalves1: I copied from the other tests. It seems to yield an argument type error without it.
```
error TS2345: Argument of type '(_: any, metadata: any) => ComponentSpec' is not assignable to parameter of type '() => any'
```

PR #873, opened 2023-03-30
Add in stderr to logs

mueschm:
## Overview
Make sure we pipe stderr to stderr

## Changes
Make sure we pipe stderr to stderr

## Tests
Showing errors it did not show before
image

### DIFF
--- a/src/commands/dev/index.ts
+++ b/src/commands/dev/index.ts
@@ -175,7 +175,7 @@ export class UpProcessManager {
     }
 
     const service_colors = new Map<string, chalk.Chalk>();
-    this.compose_process.stdout?.on('data', (data) => {
+    const handleStream = (data: any) => {
       if (this.is_exiting) {
         return;
       }
@@ -198,7 +198,9 @@ export class UpProcessManager {
           console.log(color(service + '| ') + newLine);
         }
       }
-    });
+    };
+    this.compose_process.stdout?.on('data', handleStream);
+    this.compose_process.stderr?.on('data', handleStream);
   }
 
   async run(): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/commands/dev/index.ts
```diff
@@ -175,30 +175,34 @@ export class UpProcessManager {
     }
 
     const service_colors = new Map<string, chalk.Chalk>();
-    this.compose_process.stdout?.on('data', (data) => {
-      if (this.is_exiting) {
-        return;
-      }
+    const createHandleStream = (output_func: (message: string) => void) => {
```
tjhiggins, line 178: I also noticed that we don't pipe stderr here:
```ts
const compose_process = DockerComposeUtils.dockerCompose(compose_args,
      { stdout: 'pipe', stdin: 'ignore', detached: !this.is_windows });
```

tjhiggins: Oh I guess the default is pipe so maybe not a big deal, but confusing to read

src/commands/dev/index.ts
```diff
@@ -175,30 +175,34 @@ export class UpProcessManager {
...
-            service_colors.set(service, chalk.rgb(rand(), rand(), rand()));
-          }
+            if (!service_colors.get(service)) {
+              service_colors.set(service, chalk.rgb(rand(), rand(), rand()));
+            }
 
-          const color = service_colors.get(service) as chalk.Chalk;
-          console.log(color(service + '| ') + newLine);
+            const color = service_colors.get(service) as chalk.Chalk;
+            output_func(color(service + '| ') + newLine);
```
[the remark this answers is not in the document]
mueschm: So docker logs does not do this, so just tried to copy what they did.

src/commands/dev/index.ts
mueschm: Fair I can make that more obvious so no one spends time going down some kind of wrong rabbit hole.

src/commands/dev/index.ts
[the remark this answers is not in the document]
TylerAldrich: Oh interesting, I thought they did but I think what I'm actually thinking of is docker build errors

PR #876, opened 2023-04-03
648 rounding number

mtgoncalves1:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/648.  
Fix rounding issue when reading local environment variable with `dotenv`.

## Changes
If the secret value get rounded when casting to a number, use its string format instead.

## Tests
Export an environment variable    
$ export ARC_world_text=1232456789.1232456789   
Run hello-world example    
$ architect dev   

I tested it this way instead of using the `dotenv` command, `dotenv -e .env architect dev`, because the new code doesn't seem to be picked up.

Test remote   
Use package `"@architect-io/cli": "1.37.0-arc-mary.1"` in the API.    
$ architect-local register  
$ architect-dev register  
Deploy it with the following `world_text` value  
Screenshot 2023-04-04 at 10 00 19 AM
Screenshot 2023-04-04 at 10 07 55 AM

### DIFF
--- a/src/common/utils/deploy.utils.ts
+++ b/src/common/utils/deploy.utils.ts
@@ -1,4 +1,3 @@
-import { isNumberString } from 'class-validator';
 import deepmerge from 'deepmerge';
 import * as dotenv from 'dotenv';
 import * as dotenvExpand from 'dotenv-expand';
@@ -7,7 +6,7 @@ import yaml from 'js-yaml';
 import { ArchitectError, Dictionary } from '../../';
 import { SecretsDict } from '../../dependency-manager/secrets/type';
 import { SecretSpecValue } from '../../dependency-manager/spec/secret-spec';
-import { parseValue } from './secret/helper';
+import { parseEnvironmentVariable } from './secret/helper';
 
 export default class DeployUtils {
   private static getExtraSecrets(secrets: string[] = []): Dictionary<string | number | null> {
@@ -16,8 +15,7 @@ export default class DeployUtils {
     for (const [secret_name, secret_value] of Object.entries(process.env || {})) {
       if (secret_name.startsWith('ARC_')) {
         const key = secret_name.substring(4);
-        const value = parseValue(secret_value);
-        extra_secrets[key] = value;
+        extra_secrets[key] = parseEnvironmentVariable(secret_value);
       }
     }
 
@@ -26,11 +24,7 @@ export default class DeployUtils {
       if (secret_split.length !== 2) {
         throw new ArchitectError(`Bad format for secret ${secret}. Please specify in the format --secret SECRET_NAME=SECRET_VALUE`, false);
       }
-      let value: string | number = secret_split[1];
-      if (isNumberString(value)) {
-        value = Number.parseFloat(value);
-      }
-      extra_secrets[secret_split[0]] = value;
+      extra_secrets[secret_split[0]] = parseEnvironmentVariable(secret_split[1]);
     }
 
     return extra_secrets;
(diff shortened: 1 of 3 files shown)


src/common/utils/deploy.utils.ts
```diff
@@ -7,18 +7,16 @@ import yaml from 'js-yaml';
...
 
     for (const [secret_name, secret_value] of Object.entries(process.env || {})) {
       if (secret_name.startsWith('ARC_')) {
         const key = secret_name.substring(4);
-        let value: string | number | undefined = secret_value;
-        if (value && isNumberString(value)) {
-          value = Number.parseFloat(value);
-        }
+        const value = parseValue(secret_value);
         extra_secrets[key] = value;
```
ryan-cahill, line 20: `parseValue(secret_value);` can just be set here as `extra_secrets[key] = parseValue(secret_value);`

src/common/utils/secret/helper.ts
```diff
@@ -0,0 +1,16 @@
+import { isNumberString } from 'class-validator';
+
+export const parseValue = (secret_value: string | undefined): (string | number | null) => {
```
ryan-cahill, line 3: This function name could be a little more specific. Maybe something like `parseEnvironmentVariable`?

src/common/utils/deploy.utils.ts
```diff
@@ -7,18 +7,16 @@ import yaml from 'js-yaml';
...
+    const extra_secrets: { [s: string]: string | number | null } = {};
 
     for (const [secret_name, secret_value] of Object.entries(process.env || {})) {
       if (secret_name.startsWith('ARC_')) {
         const key = secret_name.substring(4);
-        let value: string | number | undefined = secret_value;
-        if (value && isNumberString(value)) {
-          value = Number.parseFloat(value);
-        }
+        const value = parseValue(secret_value);
```
ryan-cahill, line 19: This function should also be used in the loop below - https://github.com/architect-team/architect-cli/pull/876/files#diff-70a60bd137ea6b541ead0a94f4eba0ffa6a04ee3d84b15b98cdb6d03bf2eb3edR24

This would be necessary if someone uses one or more `-s` flags such as `architect dev -s this=that -s another=1.999999999999999`

src/common/utils/secret/helper.ts
ryan-cahill, line 1: In order to use this in the API from the CLI package, this would need to be exported from `src/index.ts`. It would be good to test with an `arc-` branch to be sure

mueschm:
I guess I am a bit confused as to why not just use the string value all the time then? Like if we have an edge case where we need to treat numbers as strings, it means every downstream system needs to handle that edge case anyway. So even allowing it to be a number here feels odd. 

I know the main reason we allow this stuff is for things like `cpu` on the spec to be interpolated during deploy time from a secret. So the real question is should we just have secrets always be string and have the validation done by each resource on the spec itself?

@tjhiggins what do you think?

ryan-cahill (changes requested):
I know I didn't mark the ticket with the `tests-required` tag at first, but can you add some for the relevant cases please?

src/common/utils/secret/helper.ts
mtgoncalves1: Yes, `parseEnvironmentVariable` definitely sounds better.

ryan-cahill (changes requested):
This looks good, but can you please include a test for variable rounding when an `ARC_` environment variable is set? You can include one in a test by using something like

```ts
test
  .env({ 'ARC_a_required_key': '123456789.123456789' })
...
```

ryan-cahill (approved):

PR #878, opened 2023-04-04
throwing error when build used with image, test updates

ryan-cahill:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/651

## Changes
* Updated `build` and `image` to be exclusive in spec
* Updated JSON schema `ExclusiveOrNeither` to use a good error message
* Updated tests

### DIFF
--- a/src/commands/register.ts
+++ b/src/commands/register.ts
@@ -278,6 +278,7 @@ export default class ComponentRegister extends BaseCommand {
         continue;
       }
       delete service.debug; // we don't need to compare the debug block for remotely-deployed components
+      delete service.build; // build block data isn't relevant to remotely-deployed components
       if (service instanceof ServiceSpec && service.enabled !== undefined && !service.enabled) {
         continue;
       }
--- a/src/dependency-manager/spec/utils/json-schema-annotations.ts
+++ b/src/dependency-manager/spec/utils/json-schema-annotations.ts
@@ -124,6 +124,9 @@ export const ExclusiveOrNeither = (...properties: string[]): DecoratorSchema => 
       type: 'object',
       required: [...properties],
     },
+    errorMessage: {
+      not: `Only one of ${properties.join(', ')} is allowed, or none should exist`,
+    },
   } as DecoratorSchema;
 };
 
--- a/test/dependency-manager/spec/partials/spec-test-harness.ts
+++ b/test/dependency-manager/spec/partials/spec-test-harness.ts
@@ -3,6 +3,10 @@ import fs from 'fs-extra';
 import path from 'path';
 import { parseSourceYml } from '../../../../src';
 
+const xor_spec_arrays: string[][] = [ // can't let conflicting properties be joined in one spec
+  ['build', 'image'],
+];
+
 const lsDirectories = (dir: string) => {
   return fs.readdirSync(path.resolve(dir), { withFileTypes: true })
     .filter(dirent => dirent.isDirectory())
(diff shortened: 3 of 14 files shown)


test/dependency-manager/validation.test.ts
```diff
@@ -56,7 +56,7 @@ services:
       expect(errors[0].start?.column).eq(7);
       expect(errors[0].end?.row).eq(10);
       expect(errors[0].end?.column).eq(12);
-      expect(process.exitCode).eq(1);
+      // expect(process.exitCode).eq(1); // TODO: why is this not set?
```
mueschm, line 59: This needs to be working before we can merge.

test/dependency-manager/spec/partials/spec-test-harness.ts
```diff
@@ -38,17 +42,44 @@ const recursiveMergeTree = (tree: RecursivePartialsTree): object[] => {
...
 
       const merged_partials = tree.partials.map(current_partial => {
-        return child_partials.map(child_partial => {
+        let accums: object[] = [];
+        for (const child_partial of child_partials) {
           if (child_key.startsWith('_')) {
             let accum = current_partial;
-            for (const current_key of Object.keys(accum)) {
-              accum[current_key] = deepmerge(accum[current_key], { [child_key.replace('_', '')]: child_partial });
+            const resource_name = Object.keys(accum)[0];
```
mueschm, line 49: Can you explain what is going on here?

mueschm (changes requested):
Overall liking the changes.

test/dependency-manager/validation.test.ts
ryan-cahill: Good catch, I committed the change locally but for some reason never pushed it

test/dependency-manager/spec/partials/spec-test-harness.ts
ryan-cahill: If we didn't change what we're doing here, a bunch of tests where these partial specs are used would fail validation because we validate that we're not using `build` with `image` now

src/commands/register.ts
```diff
@@ -278,6 +278,7 @@ export default class ComponentRegister extends BaseCommand {
         continue;
       }
       delete service.debug; // we don't need to compare the debug block for remotely-deployed components
+      delete service.build; // build block data isn't relevant to remotely-deployed components
```
mueschm, line 281: I want to keep this so we can understand how someone built their component for debugging purposes. There are times when someone asks for help I go and check their spec.

src/dependency-manager/spec/utils/json-schema-annotations.ts
```diff
@@ -124,6 +124,9 @@ export const ExclusiveOrNeither = (...properties: string[]): DecoratorSchema =>
       type: 'object',
       required: [...properties],
     },
+    errorMessage: {
```
mueschm, line 127: Really like this change!

mueschm (commented):
Looking good happy to merge once the build block conversation is done.

src/commands/register.ts
ryan-cahill: Allowing this to be included would be registering an invalid spec because we insert an `image` property to a service spec when/if an image is built. When this gets merged, using the combination of `build` and `image` should result in an error from the API. Do you have any thoughts on how you'd like to implement a workaround?

mueschm: That is a valid point. I think making a ticket for debug info that gets submitted alongside the spec may be useful, but don't want to block this.

PR #880, opened 2023-04-06
Add support for making an ingress private in spec

tjhiggins:


src/commands/dev/index.ts
```diff
@@ -553,12 +359,70 @@ export default class Dev extends BaseCommand {
...
+    process.on('SIGINT', () => {
+      is_exiting = true;
+    });
+
+    compose_process.on('exit', () => {
+      if (!flags.detached) {
+        fs.removeSync(compose_file);
+      }
+      // eslint-disable-next-line no-process-exit
+      process.exit(0);
```
tjhiggins, line 389: Possible I should be using `this.exit`

src/commands/dev/index.ts
```diff
@@ -367,6 +169,11 @@ export default class Dev extends BaseCommand {
       sensitive: false,
       default: false,
     }),
+    'wait-timeout': Flags.string({
+      description: '[default: 10m] Time to wait for services to be ready/healthy before detaching.',
```
TylerAldrich, line 173: It looks like oclif adds the default text automatically, it shows up twice:

image

TylerAldrich (changes requested):
Code changes all LGTM, did a bunch of testing with dev and interrupting dev with failing/restarting containers and everything seems to be working as expected. Just one tiny --help text issue

src/commands/dev/index.ts
tjhiggins: Good find. Found a couple other places where it was duplicated and fixed.

TylerAldrich (approved):

PR #881, opened 2023-04-07
Convert compose interpolations

mtgoncalves1:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/467.  
Make sure docker compose interpolation gets converted to our syntax correctly.

## Changes
- Convert any compose environment variables with the format of ${*} to Architect format ${{ * }} and create a secret for each of them

## Tests
docker-compose.yml
```
version: '3.2'
services:
  db:
    image: "postgres:${POSTGRES_VERSION}"
    environment:
      POSTGRES_USER: ${DB_USER}
      POSTGRES_PASSWORD: ${DB_PASSWORD}
      POSTGRES_DB: ${DB_NAME}
```

Run the `init` command. It should result in the following architect file.   
$ architect-local init --from-compose=docker-compose.yml test
```
name: test
services:
  db:
    image: postgres:${{ secrets.POSTGRES_VERSION }}
    environment:
      POSTGRES_USER: ${{ secrets.DB_USER }}
      POSTGRES_PASSWORD: ${{ secrets.DB_PASSWORD }}
      POSTGRES_DB: ${{ secrets.DB_NAME }}
    reserved_name: db
secrets:
  POSTGRES_VERSION:
    required: true
  DB_USER:
    required: true
  DB_PASSWORD:
    required: true
  DB_NAME:
    required: true
```


mueschm:
The `server_online.sh` script is not working. More than likely, the number of services changed due to the test update. Should be easy and a good learning experience to update the script and fix everything.

tjhiggins:
> The `server_online.sh` script is not working. More than likely, the number of services changed due to the test update. Should be easy and a good learning experience to update the script and fix everything.

I had the same issue. Which is how I went down the --wait path. This pr should fix the issue:
https://github.com/architect-team/architect-cli/pull/880

mtgoncalves1:
>  I think that secrets specified as just `${VARIABLE}` or `$VARIABLE` shouldn't be required secrets.
Ok. I'll leave them as 
```
secrets:
  VARIABLE:
```

ryan-cahill:
> > I think that secrets specified as just `${VARIABLE}` or `$VARIABLE` shouldn't be required secrets.
> > Ok. I'll leave them as
> 
> ```
> secrets:
>   VARIABLE:
>     default:
> ```

I think that rather than an empty default property, `required: false` might be more appropriate

src/common/docker-compose/converter.ts
```diff
@@ -106,6 +113,55 @@ export class ComposeConverter {
     return { architect_yml, warnings };
   }
 
+  private static parseSecrets(pattern: RegExp, str: string, use_default: boolean, required: boolean): { service_spec: Dictionary<ServiceSpec>, secrets: Dictionary<SecretDefinitionSpec> } {
```
ryan-cahill, line 116: `str` should be a little more specific about what it is

ryan-cahill (changes requested):

ryan-cahill (approved):

PR #892, opened 2023-04-13
Revert "Revert "throwing error when build used with image, test updates""

ryan-cahill:
**Merge only after https://gitlab.com/architect-io/hub-api/-/merge_requests/760 and https://gitlab.com/architect-io/hub-api/-/merge_requests/761**

Reverts architect-team/architect-cli#890


mueschm:
Code is good but we need to deploy the API change to prod and then run the scripts.

PR #893, opened 2023-04-14
Dependencies without tags

TylerAldrich:
Related API PR that updates dependency tag resolution to support the various cases where a tag isn't specified and the component does/doesn't exist: https://gitlab.com/architect-io/hub-api/-/merge_requests/763

When a tag is specified, no functionality is changed and the current rules still apply - that is, whatever is currently deployed in the environment is used unless a component is specified in the `architect deploy` command. E.g., if `dependency:tag1` is deployed in an environment, and `architect deploy component:latest dependency:latest` is run, `dependency:latest` will be deployed and replace the `dependency:tag1` component.  If `architect deploy component:latest` is run, the `dependency` component won't get replaced even if the tag in the `architect.yml` is changed.

When no tag is specified, similar rules apply - the only addition is that `latest` is the assumed tag. So, by default, `dependency:latest` will be deployed into an environment if it's not specified via the command or `architect.yml` and the `dependency` component doesn't currently exist in the environment.

---

The primary change in the CLI is the spec for dependencies now supports an object value as well as a string. So instead of this:
```yml
dependencies:
  foo: latest
```

you can write:
```yml
depenencies:
  foo: {}  # <-- empty object, tag is implied to be "latest"
```

The following is also valid:
```yml
dependencies:
  foo: 
    tag: latest
```

The `DependencySpec` object was implemented as a class with an optional deprecated `tag` argument - a completely empty class was causing a lot of issues so I opted to do this instead and defer the issues with an object that must be empty 😬.


src/dependency-manager/spec/utils/json-schema-annotations.ts
```diff
@@ -23,6 +23,8 @@ export const AnyOf = (...args: any): DecoratorSchema => {
       anyOf.push({
         $ref: `${REF_PREFIX}${arg.name}`,
       });
+    } else if (typeof arg === 'object' && 'type' in arg && 'pattern' in arg) {
+      anyOf.push(arg);
```
TylerAldrich, line 27: This change was needed to specifically support this -
```ts
AnyOf({
  type: 'string',
  pattern: Slugs.ComponentTagValidator.source,
}, DependencySpec)
```

Currently you can only `AnyOf` _any_ string/object/etc, but not a string with a specific pattern, which this now fixes

tjhiggins (changes requested):

src/dependency-manager/spec/component-spec.ts
```diff
@@ -234,19 +247,20 @@ export class ComponentSpec {
...
 
     errorMessage: {
       additionalProperties: ComponentSlugUtils.Description,
     },
 
-    description: 'A key-value set of dependencies and their respective tags. Reference each dependency by component name (e.g. `cloud: latest`)',
+    description: 'A key-value set of dependencies with an empty value. Reference each dependency by component name (e.g. `cloud: {}`)',
   })
-  dependencies?: Dictionary<string>;
+  @Transform(transformDependencySpec(DependencySpec))
```
tjhiggins, line 262: This isn't quite what I meant. The ComponentSpec is supposed to be 1:1 with the architect.yml spec. We have the ComponentConfig to convert the ComponentSpec into something that is easier work with. So I was envisioning the transform would happen when we go from Spec -> Config. 

The ComponentSpec/ComponentConfig concept has its problems, but I don't know if we want to start transforming in multiple places without a larger overhaul.

tjhiggins (changes requested):

src/dependency-manager/spec/component-spec.ts
TylerAldrich: Ahh I see. So one thing I noticed was `ComponentConfig` is very rarely used - in the API and this PR where I had to check the whole "tag_or_object" thing, it was a `ComponentSpec` object that was being used so the `ComponentConfig` doing a transformation wouldn't have actually helped at all (so I assumed you meant the spec 😛 )

TylerAldrich: I can undo this change and instead have the transform happen just for `ComponentConfig`, it just means that we still need to check if `tag_or_object` is a string or an object in the `register` command for example

tjhiggins: node.config is the ComponentConfig which should be used more, but we def haven't realized the initial vision...

PR #895, opened 2023-04-18
Require dev port selected is 80, 443, or between 1024-65535 

TylerAldrich:
TIcket: https://gitlab.com/architect-io/architect-cli/-/issues/644 

This prevents _most_ cases of getting `ERR_UNSAFE_PORT` errors from Chrome/Firefox/etc. if the user happens to choose a port that browsers disallow connections to.

A full list of disallowed ports is browser dependent - [Chrome](https://chromium.googlesource.com/chromium/src.git/+/refs/heads/master/net/base/port_util.cc) and [Firefox](https://www-archive.mozilla.org/projects/netlib/portbanning#portlist) examples here. Ports 1-1023 are [Well-known ports](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports) and many are reserved for specific services, so by forcing users to choose ports >= 1024 and <= 65535 the chance of selecting a well-known port that is blocked is less likely.

### DIFF
--- a/src/commands/dev/index.ts
+++ b/src/commands/dev/index.ts
@@ -58,7 +58,7 @@ export default class Dev extends BaseCommand {
   static examples = [
     'architect dev ./mycomponent/architect.yml',
     'architect dev ./mycomponent/architect.yml -a myaccount --secrets-env=myenvironment',
-    'architect dev --port=81 --browser=false --debug=true --secret-file=./mycomponent/mysecrets.yml ./mycomponent/architect.yml',
+    'architect dev --port=1234 --browser=false --debug=true --secret-file=./mycomponent/mysecrets.yml ./mycomponent/architect.yml',
   ];
 
   static flags = {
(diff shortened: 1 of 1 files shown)


src/commands/dev/index.ts
```diff
@@ -432,10 +434,11 @@ export default class Dev extends BaseCommand {
           name: 'port',
           message: `Trying to listen on port ${port}, but something is already using it. What port would you like us to run the API gateway on (you can use the '--port' flag to skip this message in the future)?`,
           validate: (value) => {
-            if (new RegExp('^[1-9]+\\d*$').test(value)) {
-              return true;
+            const port = Number.parseInt(value);
+            if (Number.isNaN(port) || port < 1024 || port > 65535) {
```
tjhiggins, line 438: I find this a little weird since `--port=443` and `--port=80` should be valid. Also I don't have any of these restrictions in wsl2.

TylerAldrich: That's fair, I didn't add exceptions for 443/80 because those are the defaults and presumably you'd only use `--port` if you wanted something that wasn't the default. Can change it to allow those ports though, the downside is we can't use oclif's min/max in the flag itself.

The port restriction that Mary originally noticed is on the browsers side though, not docker - if you run `architect dev --port=101` for example it should work, but Chrome or Firefox will likely not allow you to connect because it's in their list of ports that they don't allow.

I am not 100% sure if all browsers maintain the same exact list of these ports so I didn't want to necessarily hardcode a list of ports we'd warn users about, and rather have them select something outside of the well-known port range which has plenty of reserved ports

TylerAldrich: example - building and running works

image

but the browser is sad
image

tjhiggins: Fair enough. Not super worried about it, but prob worth at least adding 80/443 to the whitelist

PR #896, opened 2023-04-20
649 liveness probe port path

mtgoncalves1:
## Overview
Warn the user if the component they register or deploy uses liveness_probe `path` and `port`. I set the removal date to be October of 2023 but do let me know when is the best time. Also, we need to merge the docs first. 

## Tests
Add to the hello-world service:
```
    liveness_probe:
      port: 3000
      path: /
```
Then run
- $ architect-local register
Screenshot 2023-04-18 at 9 59 24 PM

- $ architect-dev register   
- $ architect-deploy hello-world  
Screenshot 2023-04-19 at 10 18 32 PM

## Docs
https://github.com/architect-team/documentation/pull/36

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -3,6 +3,7 @@ import chalk from 'chalk';
 import fs from 'fs';
 import inquirer from 'inquirer';
 import AccountUtils from '../architect/account/account.utils';
+import Deployment from '../architect/deployment/deployment.entity';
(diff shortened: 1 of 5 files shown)


TylerAldrich (changes requested):
Code and tests all look good to me - only question is about the date we want to say we'll remove support for `path` and `port` by, just want to make sure that's accurate

TylerAldrich (approved):

PR #902, opened 2023-05-01
Warn for deleted cluster deployment

mtgoncalves1:
## Overview
Closes https://gitlab.com/architect-io/hub-api/-/issues/493.   
Warn the user when cluster deployment is deleted.

## Changes
Add a check for deleted cluster deployment.

## Tests
- Register your own cluster and create an env
- Go to cluster `Apps` and uninstall `Service Mesh`
- Deploy the hello-world component `architect-local deploy hello-world -e env --auto-approve`

## Pictures
Screenshot 2023-05-01 at 11 47 28 AM

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -235,10 +235,11 @@ export default class Deploy extends DeployCommand {
     const environment = await EnvironmentUtils.getEnvironment(this.app.api, account, get_environment_options);
     const has_cluster_deployment = await DeployUtils.hasClusterDeployment(this.app, environment.cluster);
     if (!has_cluster_deployment) {
+      const cluster_apps_url = `${this.app.config.app_host}/${environment.cluster.account.name}/clusters/${environment.cluster.name}/apps`;
       const confirmation_answers = await inquirer.prompt([{
         type: 'confirm',
         name: 'continue_deployment',
-        message: `We detected the application 'API Gateway' is not installed on your cluster '${environment.cluster.name}' which may lead to deployment failure. Would you like to continue your deployment?`,
+        message: `We detected that required applications aren't installed in your cluster which will lead to deployment failure. Please install the applications at ${cluster_apps_url} before attempting to deploy.`,
       }]);
       if (!confirmation_answers.continue_deployment) {
         this.exit(0);
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -233,6 +233,17 @@ export default class Deploy extends DeployCommand {
     const account = await AccountUtils.getAccount(this.app, flags.account);
     const get_environment_options: GetEnvironmentOptions = { environment_name: flags.environment };
     const environment = await EnvironmentUtils.getEnvironment(this.app.api, account, get_environment_options);
+    const has_cluster_deployment = await DeployUtils.hasClusterDeployment(this.app, environment.cluster);
+    if (!has_cluster_deployment) {
+      const confirmation_answers = await inquirer.prompt([{
+        type: 'confirm',
+        name: 'continue_deployment',
+        message: `We detected the application 'API Gateway' is not installed on your cluster '${environment.cluster.name}' which may lead to deployment failure. Would you like to continue your deployment?`,
+      }]);
+      if (!confirmation_answers.continue_deployment) {
+        this.exit(0);
+      }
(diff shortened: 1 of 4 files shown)


src/commands/deploy.ts
```diff
@@ -233,6 +233,17 @@ export default class Deploy extends DeployCommand {
     const account = await AccountUtils.getAccount(this.app, flags.account);
     const get_environment_options: GetEnvironmentOptions = { environment_name: flags.environment };
     const environment = await EnvironmentUtils.getEnvironment(this.app.api, account, get_environment_options);
+    const has_cluster_deployment = await DeployUtils.hasClusterDeployment(this.app, environment.cluster);
+    if (!has_cluster_deployment) {
```
ryan-cahill, line 237: What are your thoughts on printing a yellow warning rather than prompting?

mtgoncalves1: The reason why I did a prompt instead of a yellow warning is because the deployment will still fail with a yellow warning. But with that said, I also I'm not a big fan of the prompt either so we can change that if you want.

ryan-cahill: Yeah I don't think that it should be blocking. If there's already a yellow warning, what are your thoughts on the detail for this ticket being added to the existing one?

mtgoncalves1: What about the following yellow warning?
```
We detected that required applications aren't installed in your cluster and this deployment will not succeed. Please stop the deployment and install the necessary applications at ${cluster_apps_url} before attempting to deploy.
```

mtgoncalves1: What about the following yellow warning?
```
We detected that required applications aren't installed in your cluster and this deployment will not succeed. Please stop the deployment and install the necessary applications at ${cluster_apps_url} before attempting to deploy.
```

ryan-cahill: Maybe instead of "Please stop the deployment", "Please cancel the deployment"? Also I think we should name the specific cluster apps to install since that might not be clear to new users

mtgoncalves1: I have updated it to a warning and currently, it is the only warning in `runRemote`.

src/common/utils/deploy.utils.ts
```diff
@@ -99,4 +102,12 @@ export default class DeployUtils {
     }
     return interfaces_map;
   }
+
+  static async hasClusterDeployment(app: AppService, cluster: Cluster): Promise<boolean> {
+    if (cluster && !cluster.properties.is_shared && !cluster.properties.is_managed) {
+      const cluster_deployment: Deployment = (await app.api.get(`clusters/${cluster.id}/apps`)).data;
+      return !(cluster_deployment && cluster_deployment.action === 'delete');
```
ryan-cahill, line 109: This check looks like it will catch a case of the cluster apps being installed then uninstalled, but not a case of the cluster apps never being installed in the first place

ryan-cahill (changes requested):

src/common/utils/deploy.utils.ts
```diff
@@ -99,4 +102,12 @@ export default class DeployUtils {
     }
     return interfaces_map;
   }
+
+  static async hasClusterDeployment(app: AppService, cluster: Cluster): Promise<boolean> {
+    if (cluster && !cluster.properties.is_shared && !cluster.properties.is_managed) {
+      const cluster_deployment: Deployment = (await app.api.get(`clusters/${cluster.id}/apps`)).data;
+      return Boolean(cluster_deployment) && cluster_deployment.action !== 'delete';
```
ryan-cahill, line 109: Why does this need to be cast to a `Boolean`?

ryan-cahill (changes requested):

src/common/utils/deploy.utils.ts
mtgoncalves1: It doesn't like it when I did just `cluster_deployment`.

ryan-cahill: What's the "it" in this case?

mtgoncalves1: I thought VS code didn't like it but I just checked it and it was fine. I'll push up the change.

ryan-cahill (approved):

PR #904, opened 2023-05-03
Validate interpolation in register

mtgoncalves1:
## Overview
Closes https://gitlab.com/architect-io/architect-cli/-/issues/620.  
Validate interpolation of component during register.

## Changes
Add validation checks for interpolations

## Tests
The following should fail
```
name: hello-world
services:
  api:
    build:
      context: .
    interfaces:
      main: 3000
    environment:
      WORLD_TEXT: ${{ nonsense.world_text }}
```
Test dependencies    
```
name: hello-world
dependencies:
  authentication: latest
services:
  api:
    build:
      context: ./
    environment:
      AUTH_INTERNAL: ${{ dependencies.nonsense.services.auth.interfaces.main.url }}
```

### DIFF
--- a/src/dependency-manager/spec/utils/spec-validator.ts
+++ b/src/dependency-manager/spec/utils/spec-validator.ts
@@ -354,6 +354,50 @@ export const validateOrRejectSpec = (parsed_yml: ParsedYaml, metadata?: Componen
   return component_spec;
 };
 
+const getDependencyName = (interpolation: string): string => {
+  const regex = /dependencies\.([^.[\]]+)/;
+  const match = interpolation.match(regex);
+  return match ? match[1] : '';
+};
+
+const findInterpolationErrors = (errors: ValidationError[], component_spec: ComponentSpec) => {
+  const context_map = buildContextMap(component_spec);
+  const service_interface_config = new Set(['host', 'port', 'protocol', 'username', 'password', 'url', 'path']);
+
+  const interpolation_errors: ValidationError[] = [];
+  for (const error of errors) {
+    let interpolation_str = error.value;
+    const is_dependency_url = interpolation_str.startsWith('dependencies');
+    if (is_dependency_url) {
+      const dependency_name = getDependencyName(interpolation_str);
+      if (!(dependency_name in context_map.dependencies)) {
+        interpolation_errors.push(error);
+      }
+    } else {
+      const parts = interpolation_str.split('.');
+      const last_element = parts[parts.length - 1];
+      const is_service_url = interpolation_str.startsWith('services') && service_interface_config.has(last_element);
+      const is_database_url = interpolation_str.startsWith('databases') && interpolation_str.endsWith('url');
+      if (is_service_url || is_database_url) {
+        interpolation_str = interpolation_str.substring(0, interpolation_str.lastIndexOf('.'));
+      }
+      if (!context_map[interpolation_str]) {
+        interpolation_errors.push(error);
+      }
+    }
+  }
+  return interpolation_errors;
+};
+
+const extractInterpolationErrors = (errors: ValidationError[]) => {
+  // Regular expression for architect variables such as architect.environment and architect.build.tag
+  const architect_regex = /architect\./;
+
+  // Regular expression for environment variables such as environment.ingresses.hello-world.hello.url
+  const architect_environment_regex = /environment\./;
+  return errors.filter(error => error.message.startsWith(RequiredInterpolationRule.PREFIX) && !architect_regex.test(error.value) && !architect_environment_regex.test(error.value));
+};
+
 export const validateInterpolation = (component_spec: ComponentSpec): void => {
   const { errors } = interpolateObject(component_spec, {}, {
     keys: true,
@@ -362,8 +406,11 @@ export const validateInterpolation = (component_spec: ComponentSpec): void => {
   });
 
   const filtered_errors = errors.filter(error => !error.message.startsWith(RequiredInterpolationRule.PREFIX));
+  let interpolation_errors = extractInterpolationErrors(errors);
+  interpolation_errors = findInterpolationErrors(interpolation_errors, component_spec);
+  const all_errors = [...filtered_errors, ...interpolation_errors];
 
-  if (filtered_errors.length > 0) {
-    throw new ValidationErrors(filtered_errors, component_spec.metadata.file);
+  if (all_errors.length > 0) {
+    throw new ValidationErrors(all_errors, component_spec.metadata.file);
   }
 };
--- a/test/dependency-manager/interpolation-validation.test.ts
+++ b/test/dependency-manager/interpolation-validation.test.ts
@@ -188,5 +188,44 @@ describe('interpolation-validation', () => {
       const component_spec = buildSpecFromYml(component_config)
       validateInterpolation(component_spec)
     });
+
+    it('fail when using interpolation where path does not exist', () => {
+      const component_config = `
+        name: hello-world
+        services:
+          api:
+            build:
+              context: .
+            environment:
+              DB_ADDR: \${{ services.database.interfaces.main.url }}
+        `;
+
+      const component_spec = buildSpecFromYml(component_config);
+
+      expect(() => {
+        validateInterpolation(component_spec);
+      }).to.be.throws(ValidationErrors);
+    });
+
+    it('fail when secret does not exist', () => {
+      const component_config = `
+        name: hello-world
+        secrets:
+          world_text:
+            default: World
+        services:
+          api:
+            build:
+              context: .
+            environment:
+              WORLD_TEXT: \${{ secrets.notfound }}
+        `;
+
+      const component_spec = buildSpecFromYml(component_config);
+
+      expect(() => {
+        validateInterpolation(component_spec);
+      }).to.be.throws(ValidationErrors);
+    });
   });
 });
--- a/test/mocks/superset/architect.yml
+++ b/test/mocks/superset/architect.yml
@@ -161,7 +161,7 @@ services:
     interfaces:
       http: 8080
     environment:
-      BACKUP_DB_ADDR: ${{ databases.api-db.connection_string }}
+      BACKUP_DB_ADDR: ${{ databases.api-db2.connection_string }}
       DB_ADDR: ${{ services.api-db.interfaces.postgres.url }}/${{ secrets.db_name }}
       DB_USER: ${{ secrets.db_user }}
       DB_PASS: ${{ secrets.db_pass }}


TylerAldrich (commented):
I've been playing around with this and it seems to help, but it's not quite fully validating interpolation as part of the register.

Using this as my component:
```yaml
name: helow
description: A simple hello-world component that returns "Hello World!" on every request.
homepage: https://github.com/architect-team/architect-cli/tree/master/examples/hello-world

secrets:
  foo:
    required: false

dependencies:
  api2: latest

services:
  api:
    build:
      context: .
    interfaces:
      main:
        port: 8080
        ingress:
          subdomain: hello
    environment:
      FOO: ${{ secrets.foo }}
      BAR: ${{ dependencies.api2.banana.salad }}
      BAZ: ${{ dependencies.api3.interfaces.main.post }}
```

and running register, I get:
```
  24 |     environment:
  25 |       FOO: ${{ secrets.foo }}
  26 |       BAR: ${{ dependencies.api2.banana.salad }}
› 27 |       BAZ: ${{ dependencies.api3.interfaces.main.post }}
     |                ﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋ Invalid interpolation ref: ${{ dependencies.api3.interfaces.main.post }}
  28 |
```

which isn't quite correct - `dependencies.api2` is correct, but `banana.salad` is nonsense.

If instead we changed this:
```ts
const graph = await dependency_manager.getGraph([instanceToInstance(component_spec)], undefined, { interpolate: false, validate: false });
```

to 

```ts
const graph = await dependency_manager.getGraph([instanceToInstance(component_spec)], undefined, { interpolate: true, validate: true });
```

(on `rc`, without these changes) we do get a more accurate error:

```
  23 |           subdomain: hello
  24 |     environment:
  25 |       FOO: ${{ secrets.foo }}
› 26 |       BAR: ${{ dependencies.api2.banana.salad }}
     |                ﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋ Invalid interpolation ref: ${{ dependencies.api2.banana.salad }}
› 27 |       BAZ: ${{ dependencies.api3.interfaces.main.post }}
     |                ﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋ Invalid interpolation ref: ${{ dependencies.api3.interfaces.main.post }} - Did you mean ${{ services.api.interfaces.main.port }}?
  28 |
```

and `api2.banana.salad` is recognized as invalid.

The problem with just doing this, though, is secrets - if I change the `foo` secret to be required, suddenly we get:

```
   5 | secrets:
   6 |   memory:
   7 |     default: .25GB
›  8 |   foo:
     |   ﹋﹋ Required secret 'foo' was not provided
   9 |     required: true
  10 |
  11 | dependencies:
```

which isn't great because `register` doesn't accept secrets (yet, it will for some scenarios with this https://github.com/architect-team/architect-cli/pull/903). 

What I think we actually want to do the validation that's happening within `getGraph`, without the part that cares about whether secrets are actually defined. I don't know the history of `register` but it's kind of weird we have a `validateInterpolation` function but purposely skip the real interpolation validation in `getGraph`... curious if @ryan-cahill maybe knows more about that and there's maybe a better way to implement the more in depth check without making changes to `getGraph` necessarily 🤔

mtgoncalves1:
> I've been playing around with this and it seems to help, but it's not quite fully validating interpolation as part of the register.
> 
> Using this as my component:
> 
> ```yaml
> name: helow
> description: A simple hello-world component that returns "Hello World!" on every request.
> homepage: https://github.com/architect-team/architect-cli/tree/master/examples/hello-world
> 
> secrets:
>   foo:
>     required: false
> 
> dependencies:
>   api2: latest
> 
> services:
>   api:
>     build:
>       context: .
>     interfaces:
>       main:
>         port: 8080
>         ingress:
>           subdomain: hello
>     environment:
>       FOO: ${{ secrets.foo }}
>       BAR: ${{ dependencies.api2.banana.salad }}
>       BAZ: ${{ dependencies.api3.interfaces.main.post }}
> ```
> 
> and running register, I get:
> 
> ```
>   24 |     environment:
>   25 |       FOO: ${{ secrets.foo }}
>   26 |       BAR: ${{ dependencies.api2.banana.salad }}
> › 27 |       BAZ: ${{ dependencies.api3.interfaces.main.post }}
>      |                ﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋ Invalid interpolation ref: ${{ dependencies.api3.interfaces.main.post }}
>   28 |
> ```
> 
> which isn't quite correct - `dependencies.api2` is correct, but `banana.salad` is nonsense.
> 
> If instead we changed this:
> 
> ```ts
> const graph = await dependency_manager.getGraph([instanceToInstance(component_spec)], undefined, { interpolate: false, validate: false });
> ```
> 
> to
> 
> ```ts
> const graph = await dependency_manager.getGraph([instanceToInstance(component_spec)], undefined, { interpolate: true, validate: true });
> ```
> 
> (on `rc`, without these changes) we do get a more accurate error:
> 
> ```
>   23 |           subdomain: hello
>   24 |     environment:
>   25 |       FOO: ${{ secrets.foo }}
> › 26 |       BAR: ${{ dependencies.api2.banana.salad }}
>      |                ﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋ Invalid interpolation ref: ${{ dependencies.api2.banana.salad }}
> › 27 |       BAZ: ${{ dependencies.api3.interfaces.main.post }}
>      |                ﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋ Invalid interpolation ref: ${{ dependencies.api3.interfaces.main.post }} - Did you mean ${{ services.api.interfaces.main.port }}?
>   28 |
> ```
> 
> and `api2.banana.salad` is recognized as invalid.
> 
> The problem with just doing this, though, is secrets - if I change the `foo` secret to be required, suddenly we get:
> 
> ```
>    5 | secrets:
>    6 |   memory:
>    7 |     default: .25GB
> ›  8 |   foo:
>      |   ﹋﹋ Required secret 'foo' was not provided
>    9 |     required: true
>   10 |
>   11 | dependencies:
> ```
> 
> which isn't great because `register` doesn't accept secrets (yet, it will for some scenarios with this #903).
> 
> What I think we actually want to do the validation that's happening within `getGraph`, without the part that cares about whether secrets are actually defined. I don't know the history of `register` but it's kind of weird we have a `validateInterpolation` function but purposely skip the real interpolation validation in `getGraph`... curious if @ryan-cahill maybe knows more about that and there's maybe a better way to implement the more in depth check without making changes to `getGraph` necessarily 🤔

Dependencies are a bit complicated to handle. The reason why I handled it that way was because allow people to register a component before registering its dependencies so then even though the url may seem correct, the program still doesn't know it exists and yields an error.
```
    environment:
      FOO: ${{ secrets.foo }}
      BAZ: ${{ dependencies.api2.services.api.interfaces.main.url }}
```

TylerAldrich:
One bug I found as well is with this yaml:
```yaml
services:
  api:
    build:
      context: .
    interfaces:
      main:
        port: 8080
        ingress:
          subdomain: hello
    environment:
      FOO: ${{ secrets.foo }}
      BAR: ${{ services.api2.interfaces.banana.port }}
      BAZ: ${{ dependencies.api3.interfaces.main.post }}

  api2:
    build:
      context: .
    interfaces:
      main:
        port: 8000
```

I'm referencing a dependency w/ no dependency block - the error though is 

```
Cannot use 'in' operator to search for 'api3' in undefined
TypeError: Cannot use 'in' operator to search for 'api3' in undefined
    at findInterpolationErrors (/Users/tyler/code/architect/architect-cli/src/dependency-manager/spec/utils/spec-validator.ts:373:28)
    at validateInterpolation (/Users/tyler/code/architect/architect-cli/src/dependency-manager/spec/utils/spec-validator.ts:410:26)
    at ComponentRegister.registerComponent (/Users/tyler/code/architect/architect-cli/src/commands/register.ts:177:26)
    at async ComponentRegister.run (/Users/tyler/code/architect/architect-cli/src/commands/register.ts:130:7)
    at async ComponentRegister._run (/Users/tyler/code/architect/architect-cli/node_modules/@oclif/core/lib/command.js:80:22)
    at async Config.runCommand (/Users/tyler/code/architect/architect-cli/node_modules/@oclif/core/lib/config/config.js:272:25)
    at async Object.run (/Users/tyler/code/architect/architect-cli/node_modules/@oclif/core/lib/main.js:76:5)
```

which isn't ideal 😛.

That's a good point about dependencies though - it looks like service references are working properly and it's just dependencies that don't, so it may be totally fine to do things this way

mtgoncalves1:
Thank you for catching that bug! I'll look into it.

mtgoncalves1:
> I worry that this approach is kind of an odd band aid to the situation. It relies very heavily on understanding the spec instead of validating the actual connections.
> 
> If we could create mock nodes for dependencies in the graph with a concept of loose validation I think most of the current logic would just work. This could reduce the amount of regular expressions this system needs and allow us to think of the validation purely in terms of it being a graph rather than a spec.
> 
> For loose validation I am just thinking that nodes can be marked as in an unknown state and therefore we can dynamically use the rest of the spec to validate. The hard part here is in the context map. Though what we could do is just dynamically parse out the spec and have `*` for any value. So the context map may include
> 
> ```
> dependencies.valid.services.*.interfaces.*.host
> ```
> 
> These values would be generated by using the spec ts files. Now the rest of the validation logic I believe would be valid. Without having to worry about regular expression parsing. The other nice part is we know when things are a dictionary so pretty much all of those values for the context map can just be generated by reading the spec at runtime.
> 
> My main concern is based on our current roadmap our graph structure is going to like change a lot and this feels too brittle to me at the moment.

@TylerAldrich I spoke with Michael and this seems to be a good approach. What do you think?

TylerAldrich:
> > I worry that this approach is kind of an odd band aid to the situation. It relies very heavily on understanding the spec instead of validating the actual connections.
> > If we could create mock nodes for dependencies in the graph with a concept of loose validation I think most of the current logic would just work. This could reduce the amount of regular expressions this system needs and allow us to think of the validation purely in terms of it being a graph rather than a spec.
> > For loose validation I am just thinking that nodes can be marked as in an unknown state and therefore we can dynamically use the rest of the spec to validate. The hard part here is in the context map. Though what we could do is just dynamically parse out the spec and have `*` for any value. So the context map may include
> > ```
> > dependencies.valid.services.*.interfaces.*.host
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > These values would be generated by using the spec ts files. Now the rest of the validation logic I believe would be valid. Without having to worry about regular expression parsing. The other nice part is we know when things are a dictionary so pretty much all of those values for the context map can just be generated by reading the spec at runtime.
> > My main concern is based on our current roadmap our graph structure is going to like change a lot and this feels too brittle to me at the moment.
> 
> @TylerAldrich I spoke with Michael and this seems to be a good approach. What do you think?

That sounds reasonable to me, it sounds like it addresses my concerns about the looseness of `validateInterpolation` 👍

mtgoncalves1:
> > > I worry that this approach is kind of an odd band aid to the situation. It relies very heavily on understanding the spec instead of validating the actual connections.
> > > If we could create mock nodes for dependencies in the graph with a concept of loose validation I think most of the current logic would just work. This could reduce the amount of regular expressions this system needs and allow us to think of the validation purely in terms of it being a graph rather than a spec.
> > > For loose validation I am just thinking that nodes can be marked as in an unknown state and therefore we can dynamically use the rest of the spec to validate. The hard part here is in the context map. Though what we could do is just dynamically parse out the spec and have `*` for any value. So the context map may include
> > > ```
> > > dependencies.valid.services.*.interfaces.*.host
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > These values would be generated by using the spec ts files. Now the rest of the validation logic I believe would be valid. Without having to worry about regular expression parsing. The other nice part is we know when things are a dictionary so pretty much all of those values for the context map can just be generated by reading the spec at runtime.
> > > My main concern is based on our current roadmap our graph structure is going to like change a lot and this feels too brittle to me at the moment.
> > 
> > 
> > @TylerAldrich I spoke with Michael and this seems to be a good approach. What do you think?
> 
> That sounds reasonable to me, it sounds like it addresses my concerns about the looseness of `validateInterpolation` 👍

Cool cool. Thank you!

src/dependency-manager/spec/utils/spec-validator.ts
```diff
@@ -354,16 +356,132 @@ export const validateOrRejectSpec = (parsed_yml: ParsedYaml, metadata?: Componen
...
+      return checkInterpolationPath(queue, child[key]);
+    } else {
+      return true;
+    }
+  } else {
+    return keys.includes(key);
+  }
+};
+
+const findInterpolationErrors = (errors: ValidationError[], component_spec: ComponentSpec, graph: Readonly<DependencyGraphMutable>) => {
```
mueschm, line 391: If we have the mock dependencies why can't we just turn the validation on and just let the normal validation logic handle everything?

mtgoncalves1: Like turn on the `validate` in the `getGraph` function? What do you mean by `normal validation logic` here?

mtgoncalves1: Currently, there are 2 issues with setting `interpolate` and `validate` to `true`.
```
1. DEP_ADDR: ${{ dependencies.hello-world.services.hello.interfaces.main.url }}
     |                                      ﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋﹋ Invalid interpolation ref: ${{ dependencies.hello-world.services.hello.interfaces.main.url }} - Did you mean ${{ dependencies.hello-world.services.*.interfaces.*.url }}?
2. param_unset:
     |   ﹋﹋﹋﹋﹋﹋ Required secret 'param_unset' was not provided
```
I'm working on finding solution for these cases.

mueschm: Yeah we need to make it so the validation check allows the * when mock_dependnecies is enabled and then it should work.

mtgoncalves1:
The symbol `ø` is being used instead of `*` (`dependencies.valid.services.ø.interfaces.ø.host`) because `*` is already used as a binary operator in the `parseExpression` logic.

PR #905, opened 2023-05-04
Fix local exec

mtgoncalves1:
## Overview
Allow to run an exec command locally without needing to go through prompts

## Changes
Return account `dev` when requested

## Tests
- Go to springboot starter project
- run `architect dev`
- run `architect exec --account dev springboot.services.app -- mvn compile`

### DIFF
--- a/src/architect/account/account.utils.ts
+++ b/src/architect/account/account.utils.ts
@@ -39,6 +39,10 @@ export default class AccountUtils {
   }
 
   static async getAccount(app: AppService, account_name?: string, options?: { account_message?: string, ask_local_account?: boolean }): Promise<Account> {
+    if (account_name === 'dev') {
+      return this.getLocalAccount();
+    }
+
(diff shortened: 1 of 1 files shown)


src/architect/account/account.utils.ts
```diff
@@ -39,6 +39,10 @@ export default class AccountUtils {
   }
 
   static async getAccount(app: AppService, account_name?: string, options?: { account_message?: string, ask_local_account?: boolean }): Promise<Account> {
+    if (account_name === 'dev') {
```
mueschm, line 42: Instead of doing the comparison here there should be a overloaded function in AccountUtils that takes in an account name and returns if it is a dev account.
Local accounts are kind of their own thing and we try to keep the special logic associated with determining what they are isolated to that utils class.

mueschm (changes requested):

PR #908, opened 2023-05-10
Use a list of restricted ports instead of allowing ports > 1023.

TylerAldrich:
Related to the issue here: https://gitlab.com/architect-io/architect-cli/-/issues/662 and an update of https://github.com/architect-team/architect-cli/pull/895.

Restricting ports to 80, 443, and > 1023 wasn't quite doing what we wanted because there's still restricted ports above 1024 (like 6000, which TJ ran into when randomly choosing a port). 

Instead of limiting ports based on range, instead use Chrome/FF's restricted port list and don't allow those specifically.

### DIFF
--- a/src/commands/dev/index.ts
+++ b/src/commands/dev/index.ts
@@ -24,7 +24,7 @@ import { OverlayServer } from '../../common/overlay/overlay-server';
 import BuildPackUtils from '../../common/utils/buildpack';
 import DeployUtils from '../../common/utils/deploy.utils';
 import { booleanString } from '../../common/utils/oclif';
-import PortUtil from '../../common/utils/port';
+import PortUtil, { RESTRICTED_PORTS } from '../../common/utils/port';
 import { SecretsDict } from '../../dependency-manager/secrets/type';
 
 type TraefikHttpService = {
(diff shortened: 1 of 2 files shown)


src/commands/dev/index.ts
```diff
@@ -493,15 +493,19 @@ export default class Dev extends BaseCommand {
...
         {
           type: 'input',
           name: 'port',
-          message: `Trying to listen on port ${port}, but something is already using it. What port would you like us to run the API gateway on (you can use the '--port' flag to skip this message in the future)?`,
+          message,
           validate: (value) => {
             const port = Number.parseInt(value);
-            if (!this.isValidPort(port)) {
-              return 'Port must be 80, 443, or any port between 1024 and 66535.';
+            if (!port || port <= 0 || port > 65535) {
```
mueschm, line 507: We also need the restricted port check here I believe.

TylerAldrich: We do this question in a loop https://github.com/architect-team/architect-cli/pull/908/files#diff-c803da602bbb1783595bc04de6a99e122c48269d8c8eb1c6df6397bf5360f91bR495 so technically it still works if you type a restricted port (and asks the question again), though better UX to specifically say "Yo this port is restricted" - will update that 👍

PR #923, opened 2023-09-14
allowing an ip whitelist to be submitted as a comma-delimited string of ips/cidrs

ryan-cahill:
## Overview
Allows the `ip-whitelist` to be submitted as a comma-separated list of IPs and CIDRs

## Tests
Added a test for the new format

### DIFF
--- a/test/dependency-manager/interpolation.test.ts
+++ b/test/dependency-manager/interpolation.test.ts
@@ -1455,6 +1455,50 @@ describe('interpolation spec v1', () => {
     });
   });
 
+  it('interpolate interfaces ingress whitelist with a single, comma delimited string', async () => {
+    const component_config = `
+    name: hello-world
+    secrets:
+      ip_whitelist:
+        required: true
+    interfaces:
+      api:
+        url: \${{ services.api.interfaces.main.url }}
+        ingress:
+          ip_whitelist: \${{ secrets.ip_whitelist }}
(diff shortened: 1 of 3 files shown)


mueschm (approved):
So now the API just needs to split the string to use it correct?

ryan-cahill:
> So now the API just needs to split the string to use it correct?

That should be the case. Once I test that on a pre-release CLI version, I'll merge this in and update the API CLI version

test/dependency-manager/interpolation.test.ts
```diff
@@ -1455,6 +1455,50 @@ describe('interpolation spec v1', () => {
...
+        ingress: {
+          ip_whitelist: '1.2.3.4,8.8.8.8,8.8.4.4/24',
+          private: false
+        }
+      },
+      main: {
+        port: 8080
+      }
+    });
+  });
```
davidthor, line 1500: I'm confused about what we did in this PR. Seems like we're only testing to make sure a secret passed in as a string is interpretted as a string. We don't cover any of the cases where we provide an actual array value, do we?

ryan-cahill: That's already supported as shown in the test case above this new one. The point of this PR is to support a comma-delimited string of IPs/CIDRs so that the number of whitelisted addresses is flexible

